PR #66, opened 2019-01-15
Add greyout feature

christophfriedrich:
It is only applied if the `maskBounds` parameter is provided and set to a truthy value.


dbauszus-glx:
Excellent work.

I just did one little post merge change. Replacing params with locale and adding the maskBounds to the default locale as an optional.

https://github.com/GEOLYTIX/xyz/commit/e97f29dba1a158cca5c5f74dd606981a17d85a8d

Allowing to control whether the bounds are masked through the workspace configuration.

PR #68, opened 2019-01-22
Infoj improvements (nicer charts, better groups, add prefix/postfix fields)

christophfriedrich:
- Charts
  - Optional `active` flag for charts (initialises the group with chart view instead of table view)
  - No axes for pie and doughnut charts (axes don't make sense for these types)
  - Nicer chart tooltips that take formatting into consideration
  - Make line charts usable again (button for toggling view wasn't showing)
- Groups
  - Optional `expanded` flag for groups (initialises the group in non-collapsed state)
  - Nicer visualisation (indent a little bit etc.)
- New `prefix` and `postfix` properties for data fields


christophfriedrich:
Glad those conflicts are just SASS's `.map` files. I think you can just discard my changes to it, merge, and then do a fresh run of SASS.

PR #175, opened 2019-10-28
Vinhnew

vinhla92:
Vinh's Material icon changes.


RobAndrewHurst:
All existing material icon classes have been removed and replaced with SVG icons located within the ICONs folders and have been applied to the parent element as a background image. Changes have also been applied to javascript functions where applicable.

PR #515, opened 2022-11-18
Add alert for location listview full

dbauszus-glx:
I have reviewed the location listview module and added some comments.

The list to store records has been renamed as records and is assigned to the listview object.

Params are now assigned to the listview object which is returned.

Assigning the params required the node to be renamed as target in order to keep the params keys the same.

An alert has been added to the dictionary for en only.

The alert will be prompted when no free records are available to store another location in the listview.


RobAndrewHurst:
Is this currently on the 'Latest' instance?

dbauszus-glx:
This has not deployed to the latest instance. I also noticed that location hooks are still written to the url hooks even if selection is not possible. I will need to prevent this in the location/get method.

dbauszus-glx:
I redeployed geolytix.dev/latest for testing with the listview reduced to 3 records.

RobAndrewHurst:
> I redeployed geolytix.dev/latest for testing with the listview reduced to 3 records.

What's the reasoning behind reducing the max listview records to 3?

RobAndrewHurst:
Question about intended behaviour! When having more than 3 locations in the query params (I added 10), I got multiple alerts when loading the instance. Is this intended?

Here is a URL to replicate : https://geolytix.dev/latest?layers=mapbox_base,scratch_mvt&locations=scratch_mvt!13,scratch_mvt!114,scratch_mvt!102,scratch_mvt!51,scratch_mvt!12,scratch_mvt!81,scratch_mvt!89&lat=54.3&lng=-4&z=6.18

dbauszus-glx:
Yes, that is expected. Though it should not be possible to select more locations than the listview records allow.

RobAndrewHurst:
It does prevent additional selections.

RobAndrewHurst (approved):
## Code Review 'Add Alert for location listview full'

Good simple implementation to alert the user that they have reached the maximum number of location selections. 

Good inclusion of additional comments. Flow of module seems good!

Advised possible additions: Other language entries for alert.

PR #517, opened 2022-11-18
Fix for single distinct value for in filter.

dbauszus-glx:
By putting the response from the distinct values query into an array and flattening this array we will always have a flat array response.


RobAndrewHurst (approved):
Code Review 'Fix for single distinct value for in filter.'

Good addition of comments, and use of the `flat()` method!

PR #518, opened 2022-11-21
Isoline entry request fails

dbauszus-glx:
An isoline request will fail if a new location is put into the sea.

A failed isoline requests leaves the location locked (disabled with a mask).

The display of the entry should be set to false if a request fails to resolve with isoline features.

A disabled flag should be set on the entry with the checkbox input being disabled once the location view update event is processed.


RobAndrewHurst (approved):
### Code Review 'Isoline entry request fails'

The code handles a failed request from the Isoline module. 

Possible future changes: Change error message to something more easily understood by the  user.

PR #525, opened 2022-11-25
drawing interaction must provide event e to drawend methods

dbauszus-glx:
The drawing interaction must pass the event on to methods called from within it's execution.


simon-leech (changes requested):
@dbauszus-glx Please can you review this - this isn't currently working. 
There is currently an error occuring 
image

Suggested: 
Should have a ? Before . After properties.

dbauszus-glx:
I fixed the issue with optional chaining in the view script and redeployed the associated instance.

simon-leech (approved):

PR #527, opened 2022-11-28
Encoded in filter values

dbauszus-glx:
In filter values as provided in an array should always be encoded.

```
"filter": {
  "type": "in",
  "dropdown": true,
  "in": [
    "A%26B",
    "City Centre",
    "Town Centre",
    "Urban Centre",
    "Parade"
  ]
}
```

The ui layer filter module should decode the values for the checkbox element.

The multi-dropdown element should also decode the values for display in the dropdown and header.


simon-leech (approved):
This pull request works to decode the filter type in values returned from the database. This is required to ensure values with special characters (&) for example do not break the filter. Works as expected.

PR #530, opened 2022-11-29
increase ctrl width to 350px

dbauszus-glx:
The default width of 333px is to small for some grid views with min-width elements and a scroll bar.


simon-leech (commented):
This PR amends the min width to 350px, which fixes an issue where the information would overspill in certain scenarios.

simon-leech (approved):

PR #532, opened 2022-11-30
add update entry plugin

dbauszus-glx:
The update_entry plugin will add a `type:update_entry` to mapp,ui.location.entries.

A field must be defined as params.field in the entry.

The field value will be used to lookup a matching entry.

The user email from the mapp object will be assigned as newValue to the entry and the location will be updated.

### DIFF
--- a/public/js/plugins/update_entry.js
+++ b/public/js/plugins/update_entry.js
@@ -1,0 +1,30 @@
+export default (function(){
+
+  mapp.ui.locations.entries.update_entry = entry => {
+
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
I have tested this and it works when editing locations, both using togglelocationviewedits:true and without this. This PR stores the email address of the editor to a field defined within the layer.json infoj configuration. 

I have added this new infoj type to the XYZ Config Docs 
https://github.com/GEOLYTIX/xyz/wiki/Configuration/#update-entry

PR #533, opened 2022-11-30
parsefloat on circle draw

dbauszus-glx:
It is possible to set a float value in the input.

The value is provided as a string in the callback sent to the draw circle method.

The value should be parsed as float.


simon-leech (approved):
This PR allows values that are entered by the user as float values to be parsed as float. Whereas previously they were parse as integer.

PR #537, opened 2022-12-06
Short circuit empty cluster requests

dbauszus-glx:
An error is thrown if a cluster request returns an empty array.


RobAndrewHurst (approved):
Does exactly what is intended.

Short circuits the error if no points present in the array.

PR #539, opened 2022-12-06
longclick mouse move

dbauszus-glx:
The longclick event should not happen when using click / mousedown to pan the mapview.


RobAndrewHurst:
Please see https://github.com/GEOLYTIX/xyz/pull/535

PR #540, opened 2022-12-07
update here isoline process

dbauszus-glx:
The outer object from a here isoline response must be decoded and the coords being reversed in order to create a feature from the polyline coords. This wasn't commented and somewhat awkward.

### DIFF
--- a/lib/ui/locations/entries/isoline.mjs
+++ b/lib/ui/locations/entries/isoline.mjs
@@ -171,16 +171,16 @@ function isoline_here(entry) {
         return;
       }
 
-      const _coordinates = mapp.utils.here.decodeIsoline(response.isolines[0].polygons[0].outer)
+      // Decode outer here isoline.
+      const decoded = mapp.utils.here.decodeIsoline(response.isolines[0].polygons[0].outer)
 
-      _coordinates.polyline.forEach(p => p.reverse())
-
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
Reverse is now a flag as second argument for mapp.utils.here.decodeIsoline(polyline,reverse)

RobAndrewHurst:
image

@dbauszus-glx Is this `encode` function in the `here.mjs` module referenced or used in other instances?

dbauszus-glx:
@RobAndrewHurst not to my knowledge. It's not exported and not called inside this module. I will remove this.

dbauszus-glx:
I removed the encoding from here utils module.

RobAndrewHurst (approved):
### update here isoline process Review (13/12/22) 🎅 

- Good simplification and clean up to use the `.reverse()` function when called upon 

```
reverse ? res.map(p => p.reverse()) : res
```

image

Testing the code base against the `scratch` layer on latest resulted in successful creation of the HERE isoline.

p.s thanks for removing that function that is no longer required!

simon-leech:
I'm happy for this PR to be merged. It cleans up the HERE isoline creation process and I've also tested on the scratch layer and created an isoline successfully.

simon-leech (approved):

PR #543, opened 2022-12-12
SVG Templates

dbauszus-glx:
There was a few issues in the previous PR on this branch. https://github.com/GEOLYTIX/xyz/pull/523

The core issue being that an undefined icon.url would cause an OL assertion error.

The svg_templates would return an empty url if no substitutes param was provided with an svg template.


dbauszus-glx:
The svg_templates should probably be a util and not a plugin. I am thinking about the best way to call this since the svg_templates need to be read from the locale config when named plugins execute.

dbauszus-glx:
I added a plugins.mjs module instead of defining plugins as {} in mapp.mjs.

The plugins module exports the svg_templates method which will add the templates object to svgSymbols util.

The template method itself has been added to the svgSymbols.

dbauszus-glx:
mapp.ui.elements.legendIcon() has no concept of geometry types.

The last commit removes the icon from the params sent to the legendIcon method if the cat does not have an implicit icon style defined.

RobAndrewHurst:
@dbauszus-glx This disconnect between the two SVG PR's I am finding extremely frustrating and difficult to understand. Maybe I am not fully understanding, but can a get a better explanation of the `params` used when using a `template` from a list of templates in the `svg_template` collection on a locale and how to define a basic 'target' style for a cluster?
 
I am going to need to be able to understand EVERYTHING and then going to have to explain in great detail for the changes to existing configs to come

dbauszus-glx:
The previous PR got prematurely merged. There is nothing we can do about this now.

This PR moves the logic from a plugin into the utils.

The params for a template are called substitute. I am happy to change this to params but thought that params are very widely used and may not be the best descriptor for substitutes. You will find use cases if you search for `substitute` in the dev workspace.

dbauszus-glx:
I notice that theme styles without icon are not being merged. This should be fixed here https://github.com/GEOLYTIX/xyz/pull/543/commits/f1c6266b65ffc9234f40a2aebc48a0839e63da50

RobAndrewHurst (approved):

RobAndrewHurst (approved):

PR #544, opened 2022-12-12
Tabulator utility methods (zoomToLocation and select)

simon-leech:
I've had a go at creating the RowDblClick function for zoomToLocation: true flag.
At the moment it returns an error, zoomToLocation is not a valid table constructor.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,17 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2021: true,
+    node: true
+  },
+  overrides: [
+  ],
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module'
+  },
+  rules: {
+    eqeqeq: 'off',
+    quotes: ['error', 'single']
+  }
+}
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -1,142 +1,147 @@
 export default async (_this) => {
-
   // The dataview target is defined as string.
-  if (typeof _this.target === 'string') {
-
+  if (typeof _this.target === "string") {
     // assign target element by ID.
-    _this.target = document.getElementById(_this.target)
+    _this.target = document.getElementById(_this.target);
   }
 
   // Dataviews must be rendered into a target element.
   if (!_this.target) {
-
-    console.warn('Dataview creation requires a target key-value')
-    console.log(_this)
+    console.warn("Dataview creation requires a target key-value");
+    console.log(_this);
     return;
   }
 
   // Update method for _this.
   _this.update = async () => {
-
     // Dataviews must not update without a query.
     if (!_this.query) return;
 
     // Compile params object from dataview entry.
-    const params = mapp.utils.queryParams(_this)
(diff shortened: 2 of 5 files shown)


dbauszus-glx:
We cannot add custom feature flags to the Tabulator table config.

The reason why Tabulator warns on custom config is that in the future this may be a viable option within the Tabulator library.

image

dbauszus-glx:
A feature like this should be introduced as a plugin first before this goes into the core.

dbauszus-glx:
I commented out the _this.table.zoomToLocation to call the rowDouble click event handler to assign a method. This should be done by a plugin which assigns a method to the tabulatorUtils to be called as rowDoubleClick event.

I also commented out 'selectable' to assign the click method to select.

I added the 'select' method to the tabulator utilities as an example for the zoomToLocation method which you can assign here with a plugin.

The events are defined on the root of the dataview. I added a lookup for tabulator util methods to the check on the events object entries.

dbauszus-glx:
I commented the rowSelect method in the select utility method.

```js
    if (typeof _this.rowSelect === "function") {
      _this.rowSelect(e, row);
      return;
    }
```

a custom select method should either be directly assigned to the events config or as tabulator util through a plugin.

dbauszus-glx:
Seeing these two methods being almost the same I added that params being sent to the util.

This allows to extend methods like so:

```js
"events": {
  "rowClick": "select",
  "rowDblClick": {
    "util": "select",
    "zoomToLocation": true
  }
},
```

dbauszus-glx (approved):

dbauszus-glx:
The util key/value must be provided to find the util. The entry including the util key/value is provided as second params argument to the utility method identified by the util key.

Params will be provided as string only the events key/value is a string e.g. `rowClick: 'select'`

RobAndrewHurst:
Ah I s

> @RobAndrewHurst The util key/value must be provided to find the util. The entry including the util key/value is provided as second params argument to the utility method identified by the util key.
> 
> Params will be provided as string only the events key/value is a string e.g. `rowClick: 'select'`

Ah I see!!

RobAndrewHurst:
@dbauszus-glx Are you waiting for my review before merging?

dbauszus-glx:
@RobAndrewHurst Yeah, says review still pending.

dbauszus-glx:
@RobAndrewHurst any advise how to lint this PR?

RobAndrewHurst:
@dbauszus-glx I will provide a step by step today for this!

RobAndrewHurst:
@dbauszus-glx @simon-leech 

I have linted the module.

Does it make sense to also require the `"selectable"` property on the dataview if the events are present on the dataview anyways?

I know its not required to make the event trigger, but then it just could be an extra step we don't need.

image

RobAndrewHurst:
> It's just for the highlight. Without the selectable in the table options the row would not highlight on hover.

Yeah understand that!

RobAndrewHurst (approved):
### Tabulator utility methods (zoomToLocation and Select) (13/12/22)

- Good creation of a function in the tabulatorUtils.mjs that can be extended to acomodate other possible functionality!

PR #545, opened 2022-12-13
Multiple assignment in geometry entry button method and callback events.

dbauszus-glx:
The active class is removed and the highlight interaction is assigned in the callback as well as the button click event.

Assigning the highlight will finish the current interaction which will trigger the callback and also assign the highlight interaction, as well as remove the active class.

Removing a class twice is not noticeable. But assigning the highlight interaction twice will double assign events.

When next finished only one set of events will be removed and the highlight interaction events remain active.

As a fix the button event should simply finish the current interaction and let the callback take care of all these things.

### DIFF
--- a/lib/mapview/interactions/modify.mjs
+++ b/lib/mapview/interactions/modify.mjs
@@ -5,9 +5,6 @@ export default function(params){
   // Finish the current interaction.
   this.interaction?.finish()
 
-  // The callback method from the previous call might have assigned the highlight interaction.
-  this.interaction?.finish()
-
   _this = Object.assign({
 
     mapview: this,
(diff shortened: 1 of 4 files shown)


dbauszus-glx:
I moved the delete message note from the custom plugin to the elements context menu. Delete vertex should not be required on a plugin. The modify context menu plugin needs to be reviewed anyways at some point.

RobAndrewHurst:
When testing the multigeom edit, I noticed a usability issue. If you were to try correct the vertex of a geometry, and drag it to the desired location, the save dialog on the vertex comes up now allowing you to edit multiple before hitting save.

In the video posted you will see a vertex being hidden by the save dialog. 


https://user-images.githubusercontent.com/43169158/207916984-13d0dba7-37bc-4115-8962-02f1fff5838d.mov

RobAndrewHurst (approved):
### Code Review (#545)

The highlight interaction of the vertex on a multigeom are interacting correctly.

PR #546, opened 2022-12-14
Missing Pins

dbauszus-glx:
The pin style is generated without a feature.

uitls.style should return an icon style if a style with icon but no feature is received.

### DIFF
--- a/lib/layer/Style.mjs
+++ b/lib/layer/Style.mjs
@@ -138,14 +138,17 @@ export default layer => feature => {
   }
 
   // Check whether the feature referenced in mapview.locations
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
The conditional check for style.selected = null could not possibly work. This is fixed here https://github.com/GEOLYTIX/xyz/pull/546/commits/dbbb8e54440195ca21eeae010108d7e7649becf4

simon-leech (approved):
This PR fixes a previous issue whereby type pin entries were not displayed on the map. 
I have tested on the latest selecting multiple locations and pins are returned, labelled sequentially and coloured differently. 
image

dbauszus-glx (commented):
the default style was not applied for icons without an implicit style

PR #548, opened 2022-12-14
style scale for icons

dbauszus-glx:
fixes an issue where the default style is not applied to an icon without implicit style.


simon-leech (approved):
This PR ensures the hierarchy of scale assignment is as such - 
- If the individual icon has a scale provided, that is used
- If the individual icon has no scale provided, and the default style for the layer has a scale provided, that is used 
- Otherwise the scale is set to 1.

I have tested this on latest by altering the default value to 1.5, and the Asda icon to a scale of 3 and can confirm this is working.

PR #552, opened 2022-12-16
S3 bucket

RobAndrewHurst:
## s3 Upload with MultipartUpload

This pull request includes a new process to upload a file to an s3 bucket using the `@aws-sdk/client-s3` package.
This package will allow the creation of an s3 client that will facilitate the execution of various commands to interact with an s3 bucket. 

The current commands that are in the newly created s3.js provider are the following `PutObjectCommand`, `CreateMultiPartUploadCommand`, `UploadPartCommand`, `CompleteMultipartUploadCommand`

To facilitate the requests being made to these API's we have had to make slight changes to the `lib/utils/xhr.mjs` module.

###  lib/utils/xhr.mjs
- You can specify the content type of the payload of a request by passing in the `contentType` in the params of the module. 

```js
const requestHeader = {
    'Content-Type': params.contentType || 'application/json'
  }
```

- From a promise you can now either resolve the targets response or the full target which will include headers.

```js
resolve(params.resolveTarget ? e.target : e.target.response)
```

### s3Client
The s3 client requires 4 keys : `KEY_AWSACCESSKEYID`, `KEY_AWSSECRETACCESSKEY`, `KEY_AWSREGION`, `KEY_AWSBUCKET` - These keys may be subject to change in terms of their naming.

- AWS Access Keys - Follow these [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html) to create your keys
- `KEY_AWSREGION` - The region where the bucket is stored.
- `KEY_AWSBUCKET` - The bucket name. 

### s3 Plugin Logic

image


dbauszus-glx (changes requested):
The requestHeader can be defined as an object for the XHR module.

This should work without the need to add a check on the contentType param.

```
requestHeader: {
 'Content-Type': 'application/octet-stream',
},
```

Instead of:
```
contentType: 'application/octet-stream'
```

Only secrets should be provided as environment key. Config should be provided as params. It should be possible to provide the bucket and region as params. These should not be part of the environment.

```
"KEY_AWSBUCKET": "test-bucket-glx",
"KEY_AWSREGION": "eu-west-2",
```

URL parameter are always text, never boolean.

Please provide the command as a text param command which will allow you to lookup the matching method.

```js
const commands = {
 upload,
 getuploadID,
 uploadpart,
 completeUpload
}

commands[req.params.command](s3Client, req)
```

Instead of:
```js
  try {

    if (req.params.upload) return upload(s3Client, req)

    if (req.params.getuploadID) return createMultiPartUpload(s3Client, req)

    if (req.params.uploadpart) return uploadpart(s3Client, req)

    if (req.params.completeUpload) return completeMultiPartUpload(s3Client, req)

  } catch (e) {
    console.log(e);
  }
```

Errors should always written as errors not log.  Can you provide a test case for the try catch block?

A key for the logger method should be added to log the S3 interaction if enabled.

The plugin should have the same name as the method e.g. s3bucket_upload.js instead of s3.js.

`layer.s3bucket_upload.btn` should be `layer.s3bucket_upload.input`

Inline styles should be avoided if not necessary. e.g. `font-size: 11px;`.

The upload should work for CSV and JSON.

The input should be limited to only allow selection of json and csv files.

Rename csvFile to blob to avoid confusion.

If/else should not be used if not necessary.

It is preferred to have the smaller method first and shortcircuit. e.g.

```js
// file is small enough for upload as a whole.
if (fileSize <= 1024 * 1024 * 4.5) {

 // upload script
 return;
}

// multipart upload
```

Closing the input without selecting a file after uploading a file causes an error.

Uploading a file with the same name will overwrite the existing file with no questions asked. A check must be run and overwriting an existing file must be confirmed by user.

A download link must be returned from the upload to check that the upload was successful and to test the download capability.

dbauszus-glx:
I removed the plugin from PR and created a test plugin [here](https://geolytix.github.io/mapp/plugins/s3_bucket.js).

A workspace to test this is [here](https://geolytix.github.io/mapp/workspaces/s3_bucket.json).

Required env keys are `KEY_AWSACCESSKEYID` and `KEY_AWSSECRETACCESSKEY`.

dbauszus-glx (approved):

PR #554, opened 2022-12-19
reset longclick cursor on mouseup

dbauszus-glx:
The click event is not called after longclick triggered by the mousedown event in the highlight interaction.

mousemove cannot reset the cursor to prevent excessive processing during mousemove over the same set of highlight candidates.

A mouseup event will always be called after the click event. The mouseup can reset the cursor to auto.


RobAndrewHurst (approved):
### Code Review (#554)

- Good addition to the highlight module to include the `mouseup` event on the module.

Looks all good!

PR #557, opened 2022-12-23
tabulator-bump

dbauszus-glx:
I bumped tabulator import to 5.4.3

I commented the default table render methods. Defaults must be defaults to be overwritten by case if needed.


simon-leech (approved):
This PR bumbs the tabulator to 5.4.3, with the dataviews file changed slightly to comment out the renderHorizontal:virtual assignment.

PR #559, opened 2022-12-23
delay drawend context menu; in filter encoding;

dbauszus-glx:
also lint quotes in draw and contextmenu modules.

### DIFF
--- a/lib/mapview/interactions/draw.mjs
+++ b/lib/mapview/interactions/draw.mjs
@@ -5,9 +5,6 @@ export default function(params){
   // Finish the current interaction.
   this.interaction?.finish()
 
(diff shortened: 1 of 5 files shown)


simon-leech (commented):
The first part of the PR fixes the draw interaction to ensure an event is ended and the save pop-up appears. I tested this on the mapp/uk instance creating a manual circle with a distance popup and can confirm its working.

simon-leech (approved):
The second part of this PR encodes filter values. This fixes a bug where a filter current applied could not then be toggled by the user in the MAPP UI filter panel. 

I pushed to this commit too to provide a filter current on the retail points layer in the latest workspace.

PR #560, opened 2023-01-05
add formatter to tabulatorUtils

dbauszus-glx:
The toLocaleString formatter in the tabulatorUtils will apply the [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) method to the cell value. The locale and options can be defined as formatterParams. `en-GB` is the default locale.

Sample config
```
          {
            "title": "Capacity",
            "field": "capacity",
            "formatter": "toLocalString",
            "formatterParams": {
              "locale": "en-GB",
              "useGrouping": true
            }
          },
          {
            "title": "Actual Members",
            "field": "members",
            "formatter": "toLocalString"
          }
```

### DIFF
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -197,6 +197,16 @@ async function Table(_this) {
       col.headerFilter =
         mapp.ui.utils.tabulator.headerFilter[col.headerFilter](_this);
     }
+
+    // Check for custom formatter in the ui utils.
+    if (
+      typeof col.formatter === 'string' &&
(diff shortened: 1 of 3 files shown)


simon-leech (commented):
I've tested this and can confirm the toLocalString converts numbers to add thousand separators. 
I added an example to the dev.json workspace that can be seen on the scratch layer, and I've written this up in the config docs https://github.com/GEOLYTIX/xyz/wiki/Configuration#table-infoj-dataview

simon-leech (approved):

PR #561, opened 2023-01-09
update headerfilter

dbauszus-glx:
The min and max headerfilter should be applied on [key] input.

The like filter should be added to existing filter and not clear existing filter.


simon-leech:
Secondly, it appears that this PR actually contains things from a separate one - https://github.com/GEOLYTIX/xyz/pull/560 

Please can you remove the toLocaleString method from this PR?

PR #562, opened 2023-01-09
assign style from location instead of layer.

dbauszus-glx:
Assign style from location instead of source layer.


simon-leech (approved):
This PR allows the mvt_clone to assign the style from the location the mvt_clone is called instead of the layer. This means that the mvt_clone return can be coloured based on the location colour - allowing distinguishing between multiple mvt_clones from different locations.

RobAndrewHurst (commented):
Looks good to me!

PR #563, opened 2023-01-09
update header filter in tabulatorUtils

dbauszus-glx:
The min and max headerfilter should be applied on [key] input.

The like filter should be added to existing filter and not clear existing filter.

### DIFF
--- a/lib/ui/utils/tabulatorUtils.mjs
+++ b/lib/ui/utils/tabulatorUtils.mjs
@@ -15,20 +15,27 @@ function like(_this) {
 
     function likeFilter(e) {
 
-      // Get filter for that field if exists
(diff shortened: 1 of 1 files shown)


simon-leech (commented):
@dbauszus-glx This doesn't appear to be working correctly.
I have tested by zooming into this location and typing 'DKR' in the 'Network' field which is a filter that will return nothing. However, once I've done this and use backspace to delete that filter so no filters are applied, the table does not update to be filled again. The table also doesn't update if i use backspace to delete the filter and then click off the filter typing.

image

image

simon-leech (approved):
I can confirm moving the filters below the like filter check resolved the issues I was having. 
I am now able to apply both numeric and like filters on the same table, including using multiple like filters at the same time.

PR #564, opened 2023-01-09
delete feature style icon url in themes

dbauszus-glx:
The icon url must be deleted from the feature style in the theme methods. MVT feature styles are cached. But the style might change. With an existing URL in the feature style the style method will not recreate the URL.


simon-leech (approved):
I can confirm this PR deletes the style icon url. I have tested this using both mvt and cluster layers and can confirm it is working as expected.

PR #565, opened 2023-01-09
add check for OL layer and source to cluster format

dbauszus-glx:
An error would occur previously when the cluster format would try to clear a source of a layer which doesn't exist due to the tables zoom level restriction.


RobAndrewHurst (approved):
### Code review (2023/01/10)

Cluster layer stops rendering layer at a zoom level lower than specified in the config. No errors returned.

simon-leech (approved):
Can confirm no errors are returned and rendering stops at a zoom level lower than configured.

PR #566, opened 2023-01-10
tabulator cross layer selection

dbauszus-glx:
Providing a layer in the event params for the select util will assign this layer to select from on click.

```
              "events": {
                "rowClick": {
                  "util": "select",
                  "layer": "uk_health",
                  "zoomToLocation": true
                }
              },
```

### DIFF
--- a/lib/ui/utils/tabulatorUtils.mjs
+++ b/lib/ui/utils/tabulatorUtils.mjs
@@ -205,20 +205,25 @@ function set(_this) {
 
 }
 
-function select(_this, params) {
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
I can confirm I have tested this PR and it allows the configuration to define a layer to use for selection - which will return the location to the location panel on the defined event. If no layer is defined then the layer on which the dataview exists will be used. 

I have written a short note on this within the XYZ Config Docs too. https://github.com/GEOLYTIX/xyz/wiki/Configuration#table-dataview

image

PR #568, opened 2023-01-11
change format for selectAll highlight popup

dbauszus-glx:
Change format of the location key in the selectAll popup.

image


simon-leech (approved):
This is a minor change that splits the layer key and id for the pop-up in a different way. 

Beforehand this used an ! between the key and id. It now uses square brackets to contain the id - which looks nicer from a UI point of view.
BEFORE
image

AFTER
image

PR #569, opened 2023-01-11
assign interaction object to mapview instead of _this

dbauszus-glx:
Assigning the interaction object as _this outside the module closure didn't allow for interactions on multiple mapviews.

The interaction object must be assigned to the mapview itself.

The interaction is now deleted at the end of the finish method instead of nulled.

The callback method execution must be delayed in order to prevent a circular execution where the callback initiated an interaction which would finish this interaction with the callback again.

### DIFF
--- a/lib/mapview/interactions/draw.mjs
+++ b/lib/mapview/interactions/draw.mjs
@@ -1,12 +1,12 @@
-let _this;
-
 export default function(params){
 
-  // Finish the current interaction.
-  this.interaction?.finish()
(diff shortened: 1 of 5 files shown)


RobAndrewHurst (approved):
### Interactions review.

Test this PR against the 5 different interactions of draw, highlight, modify, snap and zoom.

All behaviour seemed to be as expected.

Also reviewed the changes against multiple custom views and all seemed to work as expected.

PR #571, opened 2023-01-12
render theme meta paragraph with layer style panel legend.

dbauszus-glx:
The string value of the `meta` key in a theme will be prepend to the legend in the style panel of the layer view.


simon-leech:
I've tested this PR - it allows for a meta tag to be added to the layer style panel within a theme. This is useful for custom theme types that cannot have the 'switch all' tag, and also useful to provide more information on the style thematic. 

I've added this to the XYZ Config Docs here https://github.com/GEOLYTIX/xyz/wiki/Configuration#meta-tag

PR #575, opened 2023-01-16
preset here isoline params in workspace

dbauszus-glx:
Elements will not be rendered for params which are defined in the workspace layer edit configuration for HERE isolines.

```
"isoline_here": {
 "transportMode": "pedestrian",
 "optimizeFor": "balanced",
 "dateISO": null
}
```


RobAndrewHurst (approved):
### Isoline Params

Adding in the params to the isoline_here object on a layer will set them, and remove the setting from the options panel. 

If you leave it blank you will receive all the options

image

PR #576, opened 2023-01-17
support cloudinary_url environment variable

dbauszus-glx:
Having the cloudinary folder hardcoded in the environment cloudinary variable isn't practical.

The CLOUDINARY_URL is now supported as environment variable as suggested in the [cloudinary node.js configuration](https://cloudinary.com/documentation/node_integration) notes.

This requires the folder to defined in mapp library calls to the cloudinary provider endpoint.

The folder name is picked up from the cloudinary_folder configuration in type images and documents entries.

```
            {
              "field": "images",
              "type": "images",
              "edit": true,
              "cloudinary_folder": "dev"
            },
            {
              "field": "documents",
              "type": "documents",
              "edit": true,
              "cloudinary_folder": "dev"
            },
```


RobAndrewHurst:
I was able to change the environment variable to the new cloudinary url. This is working.

However I noticed some odd behaviour when adding an image.

Essentially when adding an image, and then refreshing the page you see two previews of the image, and then when removing one the other remains until you refresh the view again.

Bellow is a video showcasing the issue! 

https://user-images.githubusercontent.com/43169158/215056743-080ee509-c35a-48d8-b389-56cfe3af0634.mov

RobAndrewHurst (approved):
### Cloudinary URL Review.

Tested with the new environment variable, `CLOUDINARY_URL`, with its new format, `cloudinary://my_key:my_secret@my_cloud_name`

And all seems to be working as expected!

PR #577, opened 2023-01-18
remove vercel.json

dbauszus-glx:
The vercel.json should not be part of the repository.

Other platform provider maybe used.

The default vercel.json will always have to be altered for different deployments.

There is a risk to deploy with the default vercel.json unintentionaly.

There is explanation to what the vercel.json is.

I added a wiki page for Vercel. https://github.com/GEOLYTIX/xyz/wiki/VERCEL


RobAndrewHurst (approved):
Looks all good to me!

Issue #579, opened 2023-01-18
Node fetch


dbauszus-glx:
Node 18 (LTS) has native support for the fetch API.

The node-fetch package is still on v2.x despite v3 being released.

The node-fetch dependency should be removed in favour of the native fetch API support in current Node LTS.

dbauszus-glx:
Node fetch will removed in this PR.

https://github.com/GEOLYTIX/xyz/pull/581

PR #581, opened 2023-01-19
remove node-fetch

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/issues/579

Node 18 (LTS) has native support for the fetch API.

The node-fetch dependency has been removed in favour of the native fetch API support in current Node LTS (18).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,6 @@ Node.js v12+
 
 [nanoid](https://www.npmjs.com/package/nanoid) - A tiny, secure, URL-friendly, unique string ID generator for JavaScript.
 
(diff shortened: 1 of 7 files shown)


RobAndrewHurst (approved):
### 2023-01-19

Code review removes the node-fetch module and requirement.

This now is built into node 18 LTS.

Did a search for any other possible node fetch requirements and found none.

simon-leech (approved):
This PR removes the node-fetch module as this is now built into node 18. 

I tested this on a version of node <18 and as you would expect this returns a 'referenceError: fetch is not defined' error. After updating node this error is removed - as this requires node 18 to be installed.

PR #582, opened 2023-01-19
Make saml2 dependency optional

dbauszus-glx:
The saml2 library should not be in the package.json.

The package should only be installed if needed.

The latest version of the original []([saml2-js](https://www.npmjs.com/package/saml2-js)) lib should be used, not the fork which was created in response to some dependency vulnerabilities which have since been addressed by the maintainer.

The rewrite for SAML endpoints should only be added when enabled. https://github.com/GEOLYTIX/xyz/wiki/VERCEL#saml

The SAML config notes have been added to the wiki. https://github.com/GEOLYTIX/xyz/wiki/Security#saml-sso

### DIFF
--- a/mod/user/saml.js
+++ b/mod/user/saml.js
@@ -1,41 +1,48 @@
-const saml2 = require("@geolytix/saml2-js");
+let acl, sp, idp;
 
-const { join } = require("path");
-
-const { readFileSync } = require("fs");
-
-const logger = require('../utils/logger')
-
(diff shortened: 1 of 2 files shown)


RobAndrewHurst (approved):
### 2023/02/02

I ran through the process of setting up a new SSO with SAML.

You have to include the SAML package if you want this to work.
Hitting the /saml/login endpoint allowed me to login.

PR #583, opened 2023-01-20
Public folder and resources

dbauszus-glx:
Plugins and workspaces have been removed from the public dir.

I am not a fan of install/deploy everything needed, or not.

Public workspaces and plugins are in the public mapp repository. https://github.com/GEOLYTIX/mapp

The dev, geodata, and blog workspaces were not usable for anybody outside of Geolytix.

The dev workspace has become too bloated. I will create smaller sample workspaces with public datasources in the future to be accessible for all contributors.

I don't want to go through the review process for changes to public resources in this repository.

The xyz/docs path has been removed from the express script since it was a reference to the old documentation.

Public resources should be referenced through the /public path.

This has been added in addition to the root for the express script.

For vercel the public rewrite should be like so:

```
{
 "source": "/public/(.*)",
 "destination": "/$1"
},
```

Referencing some public sub directories isn't great. I want users to be able to put resources in the public dir and be able to folders if required.

```
{
 "source": "/(workspaces|views|css|js|icons)/(.*)",
 "destination": "/$1/$2"
},
```

https://github.com/GEOLYTIX/xyz/wiki/VERCEL#public-resources


RobAndrewHurst (approved):
Good to clean up all these plugins! The process of upgrading existing instances to work with this will be straight forward.

freddie-wallace (approved):
The process for updating instances is straightforward and easy to do.

simon-leech (approved):
Happy to merge this - updating to use the plugins in the new location will be easy

PR #584, opened 2023-01-24
Geojson cluster style and properties

dbauszus-glx:
I created a public workspace to compare a grid resolution cluster with openlayers geojson cluster.

https://github.com/GEOLYTIX/mapp/blob/main/workspaces/geojson_cluster.json

The data is GEOLYTIX Retail Points hosted in a public database. https://bit.io/dbauszus/geolytix

https://geojson-cluster.vercel.app/

The geojson layer is defined as cluster with cluster config block. The only supported key is the distance (in pixel) for the cluster source.

```
"cluster": {
 "distance": 50
},
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ node_modules/*
 *.crt
 
 public/css/.sass-cache/
+public/workspaces/*
 stats.json
 
 caddy_linux_amd64
--- a/lib/layer/Style.mjs
+++ b/lib/layer/Style.mjs
@@ -23,13 +23,7 @@ export default layer => feature => {
   }
 
   // Assign default style as feature.style.
-  feature.style =  Object.assign({}, layer.style.default, layer.style.theme?.style)
-
-  // Assign properties.count from length of cluster layer features array.
-  if (layer.type === 'cluster') {
-
-    feature.properties.count = feature.get('features').length
-  }
+  feature.style =  Object.assign({}, mapp.utils.clone(layer.style.default), layer.style.theme?.style)
 
   // The featureLookup is an array of features
   if (layer.featureLookup) {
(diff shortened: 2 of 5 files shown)


dbauszus-glx:
The cluster aggregation method is now checking whether features have properties to assign.

simon-leech (changes requested):
I've reviewed this PR and have a couple of questions / changes required

1. What does the properties do here? ` 

          "properties": {
            "retailer": "retailer"
          },`
In my testing it appears this is required to get icon styling to correctly display - but how would this work with multiple different thematics pointing at different fields? 
2. It doesn't look like the cluster picks up the cluster styling defined in the layer

I've tested filtering using both the style legend and also the filter panel and both update the clustering.

dbauszus-glx:
The properties are required for theming for example. Also for labels.

Without assigning the retailer property on the geojson feature we wouldn't be able to style for the retailer property.

dbauszus-glx:
You need to define an icon style.

```
        "cluster": {
            "clusterScale": 2,
            "icon": {
                "type": "dot",
                "fillColor": "#009"
            }
        },
```

image

simon-leech (changes requested):
@dbauszus-glx thanks for clarification. Can confirm the properties and cluster styling are working as expected.

1. It appears that providing a tables object instead of just table causes this error in the console. 

image

2. Providing a filter current through the workspace file works correctly.
3. Providing a custom hover definition using a javascript module returns an error in the console.
image

dbauszus-glx:
@simon-leech  Geojson will not recognise the tables object since data is cached and not loaded on zoom change.

dbauszus-glx:
I updated the layer hover to check for properties.count > 1 instead of whether the layer.format is cluster.

The layer filter and coordinates will only be sent if the feature.count is more than 1. It doesn't make sense to have a filter and coordinates for a feature identified by an id.

simon-leech (approved):
I can confirm I've tested this PR. 
I am happy it is working as expected - tested with: 

- filter current from the workspace
- filters on the layer
- style on the layer and style filtering
- cluster styling 
- custom hover interactions

I have added the geojson type to the XYZ Config Docs https://github.com/GEOLYTIX/xyz/wiki/Configuration#layer-types

dbauszus-glx:
The filter method in the categorized switch click event is now debounced.

RobAndrewHurst (approved):
### geoJSON code review

- Performance of geoJSON is a big improvement on current.
- Every thing else seems to be all good!

PR #587, opened 2023-01-30
Make MongoDB install optional

dbauszus-glx:
Mongodb should not be installed by default but only if needed.

The require for MongoDB is now in a try catch block in the provider module.

The sample module has been moved to mapp repository.

https://github.com/GEOLYTIX/mapp/blob/main/modules/mongo_workspace.js


RobAndrewHurst (approved):
Looks all good to me!

PR #589, opened 2023-01-31
Engine check

dbauszus-glx:
The check for the engine should be equals or larger than node 18.

image


RobAndrewHurst (approved):
Looks all good to me!

PR #590, opened 2023-01-31
fix assignment of here isoline defaults

dbauszus-glx:
Fixes an issue where the defaults only were used for the here isoline parameter.


RobAndrewHurst (approved):
Nice one!

Thanks for spotting this!

PR #592, opened 2023-02-01
Prevent mousedown event after touch start

dbauszus-glx:
The long click drilldown shouldn't happen on touch devices.


RobAndrewHurst (changes requested):
I can confirm that the long press does not occur on tablet/mobile devices.

There is a console.log still in the module. Can you remove that please?

RobAndrewHurst (approved):
All good now!

PR #593, opened 2023-02-01
Define default theme as string

dbauszus-glx:
Setting theme as string in the workspace will pick that theme as default from the keys in the themes object.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -27,8 +27,14 @@ export default async layer => {
   // Set the first theme from themes array as layer.style.theme
   if (layer.style && layer.style.themes) {
 
-    layer.style.theme = layer.style.theme
-      || layer.style.themes[Object.keys(layer.style.themes)[0]];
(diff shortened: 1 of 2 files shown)


simon-leech (changes requested):
I've just tested this PR and I can confirm that adding a theme string does allow you to assign the default theme through the workspace onto a template - which will be very useful! 

Only issue I've spotted, the thematic style dropdown doesn't work quite right for this. 
You can see in the screenshot below that the 'Regions' theme is drawn on the map. However, the thematic dropdown label for this theme is incorrectly defined as 'Current Population', which is the existing theme on the template. This label should be 'Regions'. 
If you swap the thematic style then this issue resolves - and the correct labels are assigned to each thematic style. This is just an issue on initial load of the layer. 


image

dbauszus-glx:
This should now be resolved by setting the placeholder in the themes dropdown.

simon-leech (approved):

PR #594, opened 2023-02-02
Set max scale; Adjust attribution.

dbauszus-glx:
I added max-scale to the default view. This prevents the zoom on input focus experienced on iOS mobile devices.

https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone

I also adjusted the attribution.

The XYZ link was dead. Heading for old documentation.

The plate with click-through prevention didn't look great on mobile.

image

The individual attribute links should have their background.

There was also unnecessary padding and the line height was too much.

The css rules are now in the UI css instead of the view header.

image


RobAndrewHurst (approved):
Testing it on an iOS device showed that the search function is working as expected.

PR #598, opened 2023-02-06
assign empty string in option edit

dbauszus-glx:
as opposed to a 1 space string which will not resolve as falsy with a skipFalsyValue flag on the entry.


RobAndrewHurst (approved):
Good spot!

PR #600, opened 2023-02-07
Get cluster feature from features instead of nnearest.

dbauszus-glx:
Previously cluster features were queried with the nnearest method applying a layer filter to select the n nearest features that pass the layer filter at the cluster coordinates.

This PR creates a popup with a list of features in a geojson cluster.

This also enabled to define a cluster.label which will displayed instead of the ID if defined and added as property to the layer properties object.

I updated the [geojson_cluster](https://github.com/GEOLYTIX/mapp/commit/2ba75aca70956cf4b44f795613feddca2957fa73) workspace to test this.

### DIFF
--- a/lib/mapview/interactions/highlight.mjs
+++ b/lib/mapview/interactions/highlight.mjs
@@ -341,6 +341,45 @@ export default function(params){
 
     // Return with a select dialogue for cluster feature.
     if (properties.count > 1) {
+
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
I have tested this PR and can confirm that this change allows a cluster label popup, with the cluster label to be set to a field instead of ID.

PR #601, opened 2023-02-07
enable ni filter entry and filter meta

dbauszus-glx:
This PR enables to define a NI filter in addition to the IN filter.

In addition the filter meta will be added to the filter card underneath the title.

```
        {
            "title": "Retailer",
            "field": "retailer",
            "filter": {
                "meta": "Exclude from layer filter",
                "type": "ni",
                "ni": [
                    "ALDI",
                    "Lidl",
                    "Tesco"
                ]
            }
        },
```


simon-leech (changes requested):
I've been testing this and can confirm it's working well, except one small thing. If you set up a NI Type with distinct:true, dropdown:true and a meta - you will see an error. This error is not seen if you provide the filters options, it's only seen when using distinct:true. 

` "filter": {
                                "type": "ni",
                                "meta":"META",
                                "distinct":true,
                                "dropdown":true
                            }`
image

dbauszus-glx:
Good spot. I rename input to content and make sure that content is always an array prior to the unshift.

I also merged the main into this branch.

simon-leech (approved):
All working now - thanks! Happy to merge!

eo-uk (approved):

PR #602, opened 2023-02-08
Wait for user admin table to be built in promise

dbauszus-glx:
The Tabulator table must be built prior to setting a filter.

Will prevent this warning.

image


RobAndrewHurst (approved):
Perfect! I can confirm that is removes that warning!

PR #603, opened 2023-02-08
check for encoded cat keys in cat theme lookup

dbauszus-glx:
The cat theme method will attempt to look up encoded cat values in the cat keys.


simon-leech (approved):
Tested this PR and can confirm that this change lookups encoded values in the theme lookup. 
I have also updated the XYZ Config Docs to signify that when using encoding, we must use fully encoded values not partial (' & ' should be '%20%26%20' not just ' %26 '). 
https://github.com/GEOLYTIX/xyz/wiki/Configuration#categorized-themes

PR #605, opened 2023-02-10
update set headerfilter

dbauszus-glx:
The set headerfilter must return a dropdown, not a select input element.

Select input elements have severe restrictions in regards to styling. A select element util would have to be written if a select style input element is required. We don't want to construct custom elements but use element utils throughout.

The dropdown and dropdown_multi element methods have been updated to return an element node. The styling has been updated for the UL element to be positioned fixed in order to visible outside the constraints of the header container. The width of the UL is assigned in the dropdown click method, taken from the offsetwidth of the [dropdown]header element.

The placeholder param is used to set the [dropdown]header text.

The options array populates the dropdown.

```js
"headerFilter": "set",
"headerFilterParams": {
  "placeholder": "Filter in",
  "options": [
    "Sainsburys",
    "Lidl",
    "Aldi",
    "Tesco",
    "Waitrose"
  ]
}
```

By default a tabulator 'in' filter is applied to the tabulator table data. The 'layerFilter' flag will apply a layer filter instead.

The default type, for layerFilter is 'in'. The type can be set to 'ni' for layerFilter only.

Distinct field values will be queried for the options array if the 'distinct' flag is set.

```js
"headerFilter": "set",
"headerFilterParams": {
  "distinct": true,
  "layerFilter: true,
  "type": ni
}
```


public/css/_inputs.scss
```diff
@@ -131,8 +131,8 @@ label.checkbox {
 
   >ul {
     display: none;
-    position: absolute;
-    width: 100%;
+    position: fixed;
+    // width: 100%;
```
eo-uk, line 135: recommend /*width: 100%;*/ instead as css doesn't understand // for single line comment

dbauszus-glx: I removed the width. No need to have the comment hang around.

eo-uk (approved):

simon-leech:
Like it - this will be really useful!

simon-leech (approved):

PR #609, opened 2023-02-10
Fix for geojson setsource not cluster

dbauszus-glx:
Fix for geojson layer which isn't a cluster.


RobAndrewHurst (approved):
Straight forward fix!

PR #610, opened 2023-02-14
Responsive Sliding Layers Panel for Mobile

eo-uk:
Added mobile-only buttons to reveal/hide the layers panel.


dbauszus-glx:
We are trying to reduce the core script and views whenever possible. For changes to be accepted for the default view they would need be required all the time by every user, and or reduce the complexity of markup, styling, or script.

A custom view can be created and loaded through the views API.

It is also possible to set a custom view as default in the templates of a workspace.

PR #611, opened 2023-02-14
Add getLocation to location.get() method; Rename getLocation to getFeature

dbauszus-glx:
The highlight interaction getLocation method has been renamed to getFeature since the callback does not pass on a location but a feature.

The location.get() method will check for a custom getLocation method assigned to the mapview.interaction and will pass the location to that method if assigned.

This allows to get a location from the highlight method which was previously not possible. It was only possible to get a feature which may be many features for cluster.

Likewise would the popup from the longclick not work with getLocation.

This can be tested with the compare_locations plugin which assigns a custom getLocation method to the highlight interaction.

https://github.com/GEOLYTIX/mapp/blob/main/plugins/compare_locations.js

### DIFF
--- a/lib/location/get.mjs
+++ b/lib/location/get.mjs
@@ -1,5 +1,11 @@
 export default async function (location, list = location.layer.mapview.locations) {
 
+  if (typeof location.layer.mapview.interaction?.getLocation === 'function') {
+
+    location.layer.mapview.interaction?.getLocation(location)
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


eo-uk (approved):
Only thing I noticed is "location.layer.mapview.interaction?.getLocation instanceof Function" could be more performant without string comparison.

dbauszus-glx:
I changed the ` check` to `instance of Function`.

PR #620, opened 2023-02-17
updated addLayer call, changed function type checks, added docstrings and comments

eo-uk:
Updated addLayer call with an assertation error check.

Changed typeof function type checks to instanceof.

Added docstrings and comments.


dbauszus-glx (approved):
lgtm

PR #621, opened 2023-02-20
skip key, field, or query from infoj_skip array

dbauszus-glx:
The info_skip array allows to define a list of key, field, or query references. If defined on the layer these entries will be skipped.

This allows the composition of templates, or the use of templates with many pre-defined dataview templates but select which ones should be used in the mapp library execution of the infoj method.

e.g. the entry with `query: "uk_postal_area_grocery_summary"` will be skipped.

```json
    "postal_areas": {
      "src": "${TEMPLATES}/templates_v4/uk/postal_area/layer.json",
      "infoj_skip": ["uk_postal_area_grocery_summary"]
    },
```

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -33,7 +33,12 @@ export default (location, infoj) => {
       let fieldEntry = entry.location.infoj.find(_entry => _entry.field === entry.objectMergeFromField)
 
       fieldEntry && mapp.utils.merge(entry, fieldEntry.value)
-    }    
+    }
+
+    // Skip entries from infoj_skip array;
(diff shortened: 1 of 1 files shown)


eo-uk (commented):
Wouldn't this code return true only if the set has entry.key?
.has(entry.key || entry.field || entry.query)

eo-uk (approved):

dbauszus-glx:
Yes, it would be true if set either has the key, field, or query. There might be a situation where an entry does not have a field or query. For this case is the reserved key key which can defined for any key relevant things. This will become more important later when we provide an array lookup for dataviews and other object lists.

PR #622, opened 2023-02-20
Assign template from layer key

dbauszus-glx:
Initial logic had a double check on the layer.template. More importantly this rewarded poor layer keys and or layer template keys.

It shouldn't be necessary to define a layer key and a template of the same name, and then set the template to the same name as the key. This should be obvious.

The workspace method will attempt to find a template from the template key of defined in the layer json. If not template key is defined the method will look for a layer template of the same name as the layer key and assign this template to the layer json.

### DIFF
--- a/mod/workspace/assignDefaults.js
+++ b/mod/workspace/assignDefaults.js
@@ -20,13 +20,13 @@ module.exports = async workspace => {
 
       layer.key = layer_key
 
-      layer = layer.template && Object.assign({},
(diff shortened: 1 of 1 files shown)


eo-uk (approved):

RobAndrewHurst (commented):
Sweet!

simon-leech (approved):
LGTM - Nice!

PR #623, opened 2023-02-20
Warning for failed mvt_clone lookup

dbauszus-glx:
Warn and return from mvt_clone entry method if the reference layer is not found in mapview.layers object.


RobAndrewHurst (approved):
This will be useful!

PR #625, opened 2023-02-21
Encode in filter values

dbauszus-glx:
IN filter values should be encoded prior to being passed to the control element (check box or dropdown).

Filter titles (or label) derived from the value should be decoded.


eo-uk (approved):

cityremade:
Just tested with the initially reported use case, works fine, the layer updates correct.

cityremade (approved):

PR #626, opened 2023-02-22
mvt_clone entry.L must be created prior to calling the style method.

dbauszus-glx:
Otherwise the style method will fail trying to get the layer L opacity.


simon-leech (approved):
I have tested this on two different instances - one with multiple thematics on an mvt_clone and one with a single thematic. I can confirm it works for both instances - all good with me!

RobAndrewHurst (approved):
Fixes the issue!

RobAndrewHurst:
This relates to this PR : Add asertation checks on mvt_clone entry #624

PR #627, opened 2023-02-22
added possible cause and fix message to common exception

eo-uk:
Added a try-catch block to explain the possible cause of the error and suggest a possible fix. This is a common error on filters due to v3 - v4 differences.

### DIFF
--- a/lib/ui/layers/panels/filter.mjs
+++ b/lib/ui/layers/panels/filter.mjs
@@ -52,7 +52,14 @@ export default layer => {
 
       if (entry.filter.card) return;
 
-      entry.filter.field = entry.filter.field || entry.field
+      try {
+        entry.filter.field = entry.filter.field || entry.field;
+      } catch(e) {
+        if (e instanceof TypeError) {
+          console.error('Possible issue: Filter value must be an object. Possible fix: {"type": "integer"}');
+        }
+        throw e;
+      }
 
       entry.filter.remove = () => {
         delete layer.filter.current[entry.filter.field];
--- a/lib/ui/layers/panels/filter.mjs
+++ b/lib/ui/layers/panels/filter.mjs
@@ -52,14 +52,10 @@ export default layer => {
 
       if (entry.filter.card) return;
 
-      try {
-        entry.filter.field = entry.filter.field || entry.field;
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (changes requested):
Can you also include the field name in the error, so we know exactly what it relates to?

eo-uk:
> Can you also include the field name in the error, so we know exactly what it relates to?

The actual error thrown after the message has that info. Eg: "Cannot create property 'field' on string 'numeric'"

RobAndrewHurst:
> > Can you also include the field name in the error, so we know exactly what it relates to?
> 
> The actual error thrown after the message has that info. Eg: "Cannot create property 'field' on string 'numeric'"

Ah yes! Sorry, didn't see the `throw`!

RobAndrewHurst (approved):
Looks all good to me!

simon-leech (changes requested):
This looks good to me. 

However, please could we also check that the filter object contains a type too, and if not provide a message for this?

I noticed that if you provide "filter":{} (without "type":"like" for instance) no error is raised but no filter is created either.

dbauszus-glx (changes requested):
How is that line failing?

```
entry.filter.field = entry.filter.field || entry.field;
```

Can provide an example of the error thrown with the call stack.

eo-uk:
> How is that line failing?
> 
> ```
> entry.filter.field = entry.filter.field || entry.field;
> ```
> 
> Can provide an example of the error thrown with the call stack.

image

dbauszus-glx:
Instead of trying to assign a property to a string it should be checked whether entry.filter is an object.

eo-uk:
> Instead of trying to assign a property to a string it should be checked whether entry.filter is an object.

Was about to do that, LBYL works well here. Will edit it soon. However, I think try-catch can be applied to a broader range of errors without spending too much time on them and is better than leaving a cryptic error alone.

simon-leech (approved):
This looks good to me - thanks for adding the check if type is provided too :)

dbauszus-glx:
Would it be possible to create the object with the type if the check fails?

dbauszus-glx (approved):

PR #628, opened 2023-02-22
Lookup icon in cat theme for style assignment

dbauszus-glx:
This is an issue with geojson layer which may consist of icon only and not have the icon in a style block since this is not required for the default (or highlight) style object.


simon-leech (approved):
Tested on the reported instance with the issue and this is working.

RobAndrewHurst (approved):
Looks fine!

PR #629, opened 2023-02-23
fixed dropdown issue on mobile

eo-uk:
Fixed the issue where dropdown items would not display on mobile. Dropdowns under tabview require fixed position whereas ones under the side panel need absolute position.


RobAndrewHurst (changes requested):
Shouldn't this change include changes to the sass file too, not just to the bundled css

eo-uk:
Right, missed those. Added them now.

RobAndrewHurst:
> 

Nice! I also just pushed the changes!

RobAndrewHurst (approved):
Awesome! Looks good to me!

PR #632, opened 2023-02-23
add csvDownload util for tabulator

dbauszus-glx:
Provide a csvDownload util which will requires an array as argument to be downloaded as a csv file.

The title for the file can be set in the params argument.

If defined as an object. The fields array can be used to determine which column fields should be exported in which order, with which title.

Using the string flag to escape string value fields.

Quotes need to be escaped in the title.

```js
      "download_csv": {
        "title": "Comparison",
        "fields": [
          {
            "title": "Name",
            "field": "name",
            "string": true
          }, {
            "title": "Hard Pressed Living",
            "field": "hard_pressed_living"
          }, {
            "title": "\" 0-9 (S, U, HPL)\"",
            "field": "pop_0_9_u_s_hpl"
          }
]}
```

### DIFF
--- a/lib/ui/elements/toolbar_el.mjs
+++ b/lib/ui/elements/toolbar_el.mjs
@@ -32,7 +32,31 @@ function download_csv(dataview) {
   return mapp.utils.html`
     <button class="flat"
       onclick=${() => {
+
+        // The data array must have a length
+        if (!dataview.data.length) return;
+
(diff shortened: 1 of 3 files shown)


RobAndrewHurst (approved):
Nice! Good to have this as a util now! 

Good comments, easy to follow!

PR #633, opened 2023-02-24
Add tabview disabled class styling; Fix tabulator like layer filter

dbauszus-glx:
The .disabled class on the tabview element should draw a css ::after mask on top prevent interaction with the disabled tabview.

The tabulator like layer filter was broken. The layer filter must be applied **only** if flagged in the headerfilter params. The tabulator like filter should only be applied if no layer filter is applied.

Like should be the default for the layer filter. `match` can be set as optional filter type.


eo-uk:
Have you considered a "pointer-events: none;" rule?

eo-uk (dismissed):

dbauszus-glx:
Yes, I have but not needed because the element masks and does natively prevent clickthrough.

eo-uk:
Makes sense to me.

dbauszus-glx:
Ah like header filter can be enabled as layer filter like so:

```js
"headerFilter": "like",
"headerFilterParams": {
  "layer": true
}
```

A like / match layer filter can be set like so.

```js
"headerFilter": "like",
"headerFilterParams": {
  "layer": true,
  "type": "match"
}
```

RobAndrewHurst (approved):
Approved! Looks all fine!

PR #635, opened 2023-02-24
Trigger mapchange on remove filter. Create filter object from string …

dbauszus-glx:
Will revert changes from https://github.com/GEOLYTIX/xyz/pull/627

An object with the field should be created if the filter type is defined as string type filter key value.

Also adds mapchange event trigger to remove filter method. This was previously only on applyfilter method.


eo-uk (approved):
It works apart from Simon's empty object check request.

PR #636, opened 2023-02-24
Merge dropdown and dropdown_multi

dbauszus-glx:
There was a lot of code repition between the dropdown and dropdown_multi element utils.

The dropdown_multi method will return the dropdown method with the multi flag set in the params.


RobAndrewHurst (changes requested):
The merge is looking good! 

I think it would be better for the multi select to keep focus on the options to select multi options quicker and easier!

As you can see in the video recording you have to make multi more clicks to select multiple options.


https://user-images.githubusercontent.com/43169158/221589168-48efbacd-724e-4881-908c-2af66e283377.mov

RobAndrewHurst:
> The merge is looking good!
> 
> I think it would be better for the multi select to keep focus on the options to select multi options quicker and easier!
> 
> As you can see in the video recording you have to make multi more clicks to select multiple options.

Also wit regards to headerfilters or dropdowns that have a long list of options! Could we include a 'select/deselect all' option at the top of the drop down?

dbauszus-glx:
The multi dropdown will not collapse on select.


https://user-images.githubusercontent.com/22201617/222145689-3be4a0c0-500b-460f-a06c-2b825baf12db.mp4

PR #637, opened 2023-02-27
Add layerfilter to toolbar

dbauszus-glx:
The layerfilter button will be active if the filter param is set in the queryparams. Toggling the layer will flip the boolean param and update the dataview.

### DIFF
--- a/lib/ui/elements/toolbar_el.mjs
+++ b/lib/ui/elements/toolbar_el.mjs
@@ -1,8 +1,9 @@
 export default {
   viewport,
+  layerfilter,
   download_json,
   download_csv,
-  clear_table_filters
+  clear_table_filters,
 }
 
 function viewport(dataview) {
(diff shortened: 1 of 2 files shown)


lib/ui/elements/toolbar_el.mjs
```diff
@@ -17,6 +18,18 @@ function viewport(dataview) {
 
 }
 
+function layerfilter(dataview) {
+
+  return mapp.utils.html`
+    <button class=${`flat ${dataview.queryparams.filter && 'active' ||''}`}
```
eo-uk, line 24: Easier to read this way?:
dataview.queryparams.filter ? 'active' : ''

dbauszus-glx: I don't mind. Feel free to change. I find logical and ternary operator equally easy to read.

eo-uk (approved):

PR #638, opened 2023-02-27
added docstring to fitView

eo-uk:
Added docstring to fitView


dbauszus-glx (changes requested):
Can you add that the extent must be [minX, minY, maxX, maxY]

dbauszus-glx (approved):

PR #639, opened 2023-02-27
Remove default scale 1 and colour

dbauszus-glx:
Icons already have a default scale of 1 which is applied in the style method if no scale is otherwise defined for an icon.

The defaults shouldn't apply a colour to an icon type which has a default.


RobAndrewHurst (approved):
Nice! Not necessary to keep those defaults!

PR #641, opened 2023-02-27
added docstring to addLayer, readability improvements, semi colons

eo-uk:
The var "hooks" is now hooksExist and is a boolean to convey its purpose better.

Replaced !! and tilde operator with >= 0 check to improve readability.

Added docstring and semicolons.


dbauszus-glx (changes requested):
Can you use `new Set(mapp.hooks.current.layers).has(layer.key)` for readability.

dbauszus-glx (approved):
lgtm

PR #642, opened 2023-02-28
Check geom field in geojson layer mod

dbauszus-glx:
The geojson layer mod must check whether a feature has a geometry.


RobAndrewHurst (approved):
Nice! Tested this against a geojson layer! Stepped through the code, and all works as expected!

PR #643, opened 2023-02-28
fixed regex that used to capture single { instead of double, added co…

eo-uk:
Fixed regex that used to capture single { instead of double:
/\{|\{|\}\}/g => /\{\{|\}\}/g

Added comments, stylistic changes.


dbauszus-glx (changes requested):
I don't understand what the fix is for.

What is the error that has been resolved by this fix?

I tested to load this view which now fails to load. https://github.com/GEOLYTIX/mapp/blob/main/views/embedded.html

dbauszus-glx (approved):
Looks good after merge.

PR #644, opened 2023-02-28
workspace dbs as fallback

dbauszus-glx:
If the dbs is defined in the workspace it will be used as fallback if not otherwise defined.

```js
{
 "dbs": "MAPP",
 "templates: {}.
 "locales": {}
}
```

### DIFF
--- a/mod/gazetteer.js
+++ b/mod/gazetteer.js
@@ -130,7 +130,7 @@ async function datasets(req, locale, results) {
       ${filter}
       LIMIT ${dataset.limit || locale.gazetteer.limit || 10}`
 
-    records.push(dbs[dataset.dbs || layer && layer.dbs](q, SQLparams));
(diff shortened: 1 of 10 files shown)


RobAndrewHurst:
@dbauszus-glx 

image

What about this like the mvt_cache where it will update a layers cache file. In that instance will the layer require a `dbs` or should it default to the workspace too?

dbauszus-glx:
> @dbauszus-glx
> 
> image
> 
> What about this like the mvt_cache where it will update a layers cache. In that instance will the layer require a `dbs` or should it default to the workspace too?

Yes... I missed them yesterday.

Should be added in last commit.

image

RobAndrewHurst (approved):
Awesome stuff! this will be very useful going forward! And its backwards compatible!

PR #646, opened 2023-03-01
Query for distinct values in filter doesn't return any values.

dbauszus-glx:
The filter card should be shown to be cancelled.

A warning will be printed in the console.

image


RobAndrewHurst (approved):
Looks good to me!

PR #648, opened 2023-03-02
Update infoj_skip to also check on entry Type and entry Group

simon-leech:
This PR will update the infoj_skip to also check for entry.type and entry.group. 
This is useful as it allows skipping of multiple infoj entries in one go, and also allows for custom types provided via plugins to be skipped too if required.


RobAndrewHurst (approved):
Good idea @simon-leech

PR #650, opened 2023-03-03
enable placeholder for edit options dropdown

dbauszus-glx:
The placeholder should be shown if defined in an entry edit config for the options dropdown.

```
{
 "title": "Optional",
 "field": "option",
 "inline": true,
 "edit": {
  "placeholder": "Select from list",
  "options": [
   "foo",
   "bar"
  ]
 },
}
```


RobAndrewHurst (commented):
Nice! Looks good to me!

PR #652, opened 2023-03-06
assign layer-view

dbauszus-glx:
A layer-view element will be created at the beginning of the layer view method. The layer drawer will be rendered into the layer-view element.

### DIFF
--- a/lib/ui/layers/_layers.mjs
+++ b/lib/ui/layers/_layers.mjs
@@ -5,6 +5,7 @@ import listview from './listview.mjs'
 import filters from './filters.mjs'
 
 // Panels
+import meta from './panels/meta.mjs'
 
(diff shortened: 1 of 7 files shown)


eo-uk:
Works well from inside a plugin using the layer object.

dbauszus-glx:
The methods for the layer.drawer header should be inside the block for the layer.drawer. It would not make sense to create the header if it is not rendered with the drawer.

The meta method has been moved to the panel. The only reason this was not done before was the need to have the meta being drawn first.

The layer.panelOrder is now assigned as a default if not in the layer config. The content array will be sorted according to the data-id of the content element.

RobAndrewHurst (approved):
Looks good to me!

PR #653, opened 2023-03-06
filter fixes and keepPlaceholder

dbauszus-glx:
Decoding of the title should be tried prior to populating the dropdown.

The `selected` class toggle should only apply for multiselect filter.

The callback for the filter interface should shortcircuit if the filter.card exists.

The clearAll button was invisible because of a missing dictionary entry.

`[].flat()` should be used to ensure a value is an array.


RobAndrewHurst (approved):
Nice!

Solves the issue of the multi select styling being present on a single select drop down.

PR #655, opened 2023-03-07
merge instead of assign layer to template

dbauszus-glx:
The layer should be merged with a template instead of assigned. Otherwise the template itself will be modified across the workspace.

For example a template with a named dataview should only be merged into the workspace layer where the template is used not into all shared layer templates.

In this scenario the title: "foo" should only be assigned to the layer which has the legal_dataviews in the templates array. This layer and dataview should be display true. The legal_prison layer should not be displayed and the same dataview should have it's original title.

```js
{
    "dbs": "MAPP",
    "templates": {
        "legal": {
            "format": "cluster",
            "table": "geodata.uk_glx_geodata_poi_legal",
            "geom": "geom_p_4326",
            "cluster_label": "poi_name",
            "qID": "id",
            "infoj": [
              {
                "type": "key"
              },
              {
                "type": "pin",
                "field": "pin",
                "fieldfx": "ARRAY[ST_X(geom_p_4326),ST_Y(geom_p_4326)]",
                "class": "display-none"
              },
              {
                "field": "poi_type",
                "title": "Type",
                "inline": true
              },
              {
                "field": "poi_name",
                "title": "Name"
              },
              {
                "field": "add_one",
                "title": "Address"
              },
              {
                "field": "add_two",
                "skipNullValue": true
              },
              {
                "field": "town",
                "skipNullValue": true
              },
              {
                "field": "locality",
                "skipNullValue": true
              },
              {
                "field": "postcode",
                "skipNullValue": true
              },
              {
                "field": "court_type",
                "title": "Court type",
                "skipNullValue": true,
                "inline": true
              }
            ],
            "dataviews": {
              "legal": {
                "title": "Legal",
                "target": "tabview",
                "query": "legal_dv",
                "queryparams": {
                  "filter": true
                },
                "table": {
                  "layout": "fitColumns",
                  "selectable": true,
                  "autoColumns":true,
                  "columns": []
                }
              }
            }
        },
        "legal_dv": {
            "template": "select poi_name, poi_type, town from geodata.uk_glx_geodata_poi_legal where true ${viewport} ${filter}"
        },
        "legal_dataview": {
            "dataviews": {
                "legal": {
                    "title": "Foo",
                    "target": "tabview",
                    "display": true,
                    "query": "legal_dv",
                    "queryparams": {
                        "filter": true
                    },
                    "viewport": true,
                    "mapChange": true,
                    "toolbar": {
                        "download_csv": true
                    }
                }
            }
        }
    },
    "locales": {
        "UK": {
            "name": "United Kingdom",
            "extent": {
                "north": 63,
                "east": 7,
                "south": 47,
                "west": -15,
                "mask": true
            },
            "view": {
                "lat": 54.3,
                "lng": -4,
                "z": 6
            },
            "minZoom": 4,
            "maxZoom": 18,
            "layers": {
                "OSM": {
                    "display": true,
                    "format": "tiles",
                    "attribution": {
                    }
                },
                "legal_prison": {
                    "template": "legal",
                    "filter": {
                        "current": {
                            "poi_type": {
                                "like": "Prison"
                            }
                        }
                    }
                },
                "legal_court": {
                    "display": true,
                    "template": "legal",
                    "templates": ["legal_dataview"],
                    "filter": {
                        "current": {
                            "poi_type": {
                                "like": "Court"
                            }
                        }
                    }
                }
            }
        }
    }
}
```

### DIFF
--- a/mod/workspace/assignDefaults.js
+++ b/mod/workspace/assignDefaults.js
@@ -20,7 +20,8 @@ module.exports = async workspace => {
 
       layer.key = layer_key
 
-      layer = Object.assign({},
+      // Merge layer templates
+      layer = merge({},
 
         // Assign layer template implicit or from key lookup.
         workspace.templates[layer.template || layer.key] || {},
@@ -50,6 +51,7 @@ module.exports = async workspace => {
 
       if (Array.isArray(layer.templates)) {
 
+        // Merge templates from templates array into layer.
         layer.templates.forEach(template => {
           merge(layer, workspace.templates[template] || {})
         })


RobAndrewHurst (approved):
Tested this, with a workspace that will have layers with the same object names, such as the dataview example provided above.

Also tried an arb use case where a layer uses multiple templates that reference the same object but will have different combinations of values. The last object that is merge will obviously replace the existing properties values where found. This will not be a typical use case but something to consider in the future.

PR #656, opened 2023-03-07
Updated default panelOrder to include dataviews and adding new locations

simon-leech:
Updated the default panelOrder to also define the position of the dataviews and draw drawers


dbauszus-glx (approved):

simon-leech:
I have just reversed the order of the sorting so that undefined drawers are returned to the bottom of the panel order

RobAndrewHurst (approved):
Nice! Looks good to me!

PR #658, opened 2023-03-08
added icon button element

eo-uk:
Added a button element with icon:

```
mapp.ui.elements.iconButton({
    data_id: "button-id",
    label: "Button Label",
    onclick: () => {console.log("foo")},
    iconURL: "<some_icon_url>",
    class: "<custom_classes>",
    style: "<custom_style>;",
    styleIcon: "<custom_icon_style>"
  })
 ```
  
image


eo-uk:
I can move the css to core

dbauszus-glx (changes requested):
style should be just params.style.

The default style should be added to _button.scss

RobAndrewHurst (approved):
Nice!

RobAndrewHurst:
> style should be just params.style.
> 
> The default style should be added to _button.scss

@eo-uk Has done this in the proceeding commits!

dbauszus-glx (changes requested):
Can you add the active class to the button. In order to toggle whether the button is styled as active in the callback method.

I would rename the method to buttonIcon or button_icon.

If there are multiple button methods we would like them to be consistent. Like slider, slider_ab, dropdown, dropdown_multi.

The type, e.g. button, slider, dropdown, etc. should be first followed by a modifier if needed.

The outside element should be type button not div to provide semantic html.

eo-uk:
Outside element is already a button. Pushed the other requested changes.

dbauszus-glx (approved):

PR #659, opened 2023-03-08
merge from entry

dbauszus-glx:
The `objectMergeFromEntry` key will look up an entry type with the value provided.

```js
        {
          "type":"dataview",
          "label": "Rail Stations",
          "target": "tabview",
          "display": true,
          "query": "query",
          "objectMergeFromEntry": "regions_bus_stop_tableview",
          "queryparams": {
            "id": true,
            "group_field": "stat_type",
            "table": "geodata.uk_glx_geodata_transport_rail_stations a",
            "table_geom": "a.geom_p_4326",
            "boundary_table": "geodata.uk_glx_geodata_metrics_region b",
            "boundary_geom": "b.geom_4326"
          },
          "toolbar": {
            "download_csv": true
          },
          "table": {
            "layout": "fitColumns",
            "selectable": true,
            "columns": [
              {
                "field": "count",
                "title": "Count",
                "hozAlign": "right"
              },
              {
                "field": "value",
                "title": "Category"
              }
            ]
          }
        }
```

The the `entry.merge` object will be merged into the entry object.

```js
        {
          "type": "regions_bus_stop_tableview",
          "merge": {
            "label": "Bus Stops",
            "queryparams": {
              "id": true,
              "group_field": "stoptype",
              "table": "geodata.uk_glx_geodata_transport_bus_stop a",
              "table_geom": "a.geom_p_4326",
              "boundary_table": "geodata.uk_glx_geodata_metrics_region b",
              "boundary_geom": "b.geom_4326"
            }
          }
        }
```

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -41,6 +41,10 @@ export default (location, infoj) => {
       // Find the entry to merge.
       let fieldEntry = entry.location.infoj.find(_entry => _entry.type === entry.objectMergeFromEntry)
 
+      if (fieldEntry===undefined) {
+        console.warn('The entry you are trying to merge into does not exist. Please check your configuration')
+      };
+
       // Only merge the entry.merge object.
       fieldEntry 
         && fieldEntry.merge instanceof Object 
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -41,10 +41,6 @@ export default (location, infoj) => {
       // Find the entry to merge.
       let fieldEntry = entry.location.infoj.find(_entry => _entry.type === entry.objectMergeFromEntry)
 
-      if (fieldEntry===undefined) {
-        console.warn('The entry you are trying to merge into does not exist. Please check your configuration')
-      };
-
       // Only merge the entry.merge object.
       fieldEntry 
         && fieldEntry.merge instanceof Object 
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -35,6 +35,18 @@ export default (location, infoj) => {
       fieldEntry && mapp.utils.merge(entry, fieldEntry.value)
     }
 
+    // Merge from another entry.
+    if (entry.objectMergeFromEntry) {
+
+      // Find the entry to merge.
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (changes requested):
Works, well! But if you have a template entry that is not found, you don't get an error or warning! Can we add this in?

```json
"__IDONTEXIST__": {
      "infoj": [
        {
          "type": "regions_bus_stop_tableview",
          "merge": {
            "label": "Bus Stops",
            "queryparams": {
              "id": true,
              "group_field": "stoptype",
              "table": "geodata.uk_glx_geodata_transport_bus_stop a",
              "table_geom": "a.geom_p_4326",
              "boundary_table": "geodata.uk_glx_geodata_metrics_region b",
              "boundary_geom": "b.geom_4326"
            }
          }
        }
      ]
    }
```

simon-leech (changes requested):
This is working well and will be so so useful! 

I committed to this PR a console warning if the entry type you are trying to merge into does not exist, for aiding debugging. 

This might need a review before merging though - so will let @RobAndrewHurst review this?

simon-leech (approved):

RobAndrewHurst:
> I don't think I want a warning as this is completely valid.
> 
> Please have a look at the example in the original post.
> 
> ```js
> "objectMergeFromEntry": "regions_bus_stop_tableview",
> ```
> 
> Where regions_bus_stop_tableview is found I want to merge. Where it isn't found I do not want to merge but keep the entry as is.
> 
> In different layer configurations I may add different `type: 'regions_bus_stop_tableview'`, or none. All situations are valid, non require a warning.

Ok! So long as that's the intention

RobAndrewHurst (approved):

dbauszus-glx:
@RobAndrewHurst yeah. this was done on purpose. Because you might want load different params which are all called the same.

For example you might have a demographics_dataview and want to load params_for_demographic_dataview. It would be hard to have a different key for different params entry in the dataview config. So the idea is that different layer all merge from params_for_demographic_dataview but the actual entry could be different for each layer. The merge entry will overwrite existing key values so it is completely valid to provide the entry and load defaults which would only be overwritten (merged) if the params_for_demographic_dataview is added and found in the layer infoj array.

PR #660, opened 2023-03-08
Add maplibre format with version bump. Update mapboxgl format.

dbauszus-glx:
The maplibre version 1.15.3 which is used to support the mbtiles format cannot be bumped to version since the mapbox schema support was removed with the version 2 release.

The _render, cancel, resize have been removed from the layer render method. The container for the maplibre Map control is appended to the mapview.Map target and will resize accordingly.

The _render, and cancel methods within the render do not seem to have any effect with this version of the maplibre library.

The mbtiles format will not check for the library to be defined in the head but will always import the specific version required from the skypack cdn. A console warning has been added for this format.

It is recommended to use the maplibre format which will check for the library having been loaded prior to the format method call. Version 2.4 will be imported from the skypack cdn if required.

`style.URL` replaces `mbStyle` key/value.

Transform URL methods are required in order to support mapbox `style.URL`.

The transformation method and supporting functions are taken from https://github.com/rowanwins/maplibregl-mapbox-request-transformer with a few modifications.

The `mapboxgl` format has been updated to use a dynamic import of the mapbox-gl v2.13 as fallback.

The `mapboxgl` format still shows some stutter on scroll-wheel zoom which will need to be investigated in a separate ticket.


RobAndrewHurst (approved):
Awesome! Looks good!

PR #661, opened 2023-03-09
set layerfilter as headerfilterparam for all type of filter.

dbauszus-glx:
Previously some filter types used layer and some used layerFilter. Every type should use layerFilter.


simon-leech (approved):
Looks good to me and makes sense to have all named the same way for consistency

PR #662, opened 2023-03-09
parse filter min max as integer for type integer

dbauszus-glx:


simon-leech (changes requested):
This is all working well for me!

However, please could we convert integers into integer with thousand separator? 
So 10000 becomes 10,000 as this is much nicer for the user to read.

dbauszus-glx:
@simon-leech  That is possible but not in this PR. This PR is for a fix where the wrong type is assigned to a value.

Changing the behaviour of an input will be a lot more involved. First an input utility would have to be created. In the case of the range slider the input utility which formats input values would then have to be integrated into the slider utility.

In the case of numeric the filter. The input utiliity would have to be created. The utility integrated into the slider element utility. The configuration for the numeric format would have to be integrated into the filter, and then fed all the way done towards the input utility.

None of this is related to this PR.

This PR is for defining the min and max as integer for integer type filter.

PR #663, opened 2023-03-09
array merge must check that source and target are both arrays

dbauszus-glx:


simon-leech (dismissed):
@dbauszus-glx I have been testing this using the bugs_and_testing/geojson_icons workspace.json file. 

This does not appear to be correctly applying the styling as all thematic values are assigned the default styling. 
Also, no cluster styling is drawn at the bottom of the styling panel, even though a separate cluster styling is provided.

simon-leech (approved):
Looks good to me!

RobAndrewHurst (approved):

PR #664, opened 2023-03-10
add wkt layer format

dbauszus-glx:
The wkt format is pretty much the same as the geojson format. However data is not transferred as json with geojson geometry as string, but as an array of value arrays with the first value being the id, the second the geometry as wkt, and subsequent array values being the properties.

a geojson feature will look like this:
```
{"type":"Feature","geometry":{"type":"Point","coordinates":[-0.18261055,51.4893428]},"properties":{"road_num":"U"},"id":942883}
```

the same feature as wkt:
```
[942883,"POINT(-0.18261055 51.4893428)","U"]
```


RobAndrewHurst:
@dbauszus-glx 

Does styling for this still work the same as geojson layers?

image

I can't seem to get the styling to work correctly.

dbauszus-glx:
@RobAndrewHurst Yes styling is the same as geojson. Graduated themes do not work on either at the moment. I need to add the value aggregation to the OL cluster method.

RobAndrewHurst (approved):
Tested with making my own WKT, and works just as a geoJSON layer does!

PR #665, opened 2023-03-13
Added a plugin hooks to fix layer.view

eo-uk:
Added a hooks array to mapp.plugins that gets executed when layer views are fully loaded, solving the issue where the layer.view element in a plugin does not have children nodes available.

Usage: 

```
mapp.ui.layers.panels.someplugin = (layer) => {
        mapp.plugins.onViewsLoad(() => {
            // layer.view is fully rendered
            console.log(layer.view);
        })
}
```

### DIFF
--- a/lib/plugins.mjs
+++ b/lib/plugins.mjs
@@ -17,3 +17,16 @@ export function svg_templates(plugin) {
       })
   })
 }
+
+export function onViewsLoad(hook) {
+  /**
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
Can I get a sample plugin where this hook is needed and can be tested with?

dbauszus-glx:
Plugin / panel methods should not have access to other panels. Elements should be created in a panel to be returned to the layer view rather than modifying elements in other panels.

PR #667, opened 2023-03-13
a drawer should also be disabled if the immediate parent is disabled.

dbauszus-glx:


simon-leech (approved):
This fixes the issue whereby layers would not be initially disabled if the zoom extent of the initial load was lower than the minimum defined in the tables object. 

Happy to merge

RobAndrewHurst:
Nice! I noticed this issue after this Merge of this PR

https://github.com/GEOLYTIX/xyz/commit/f614ed95bab5596e0b9d4b11dce1d4432b89a06d

RobAndrewHurst (approved):
Good spot!

PR #669, opened 2023-03-15
Added popup panel to elements

eo-uk:
Added a panel that pops up when active and disappears when not with a sliding animation.


dbauszus-glx:
I need some help with this.

I created a simple plugin which takes an array and creates a panel for each item in the array. The panel rendered in a div prior to a div with the buttons. The click of the button toggles the active class on the panel element.

```js
export default (function () {

    mapp.ui.layers.panels.flexIconBtns = layer => {

        const panel = mapp.utils.html.node`
        <div>
        <div>
        ${layer.flexIconBtns.map(btn => {

            btn.panel = mapp.ui.elements.panelPopup({
                "title": btn.label,
                "content": "Test Content"
            })

            return btn.panel
        })}
        </div>
        <div 
        
            style="display: flex; flex-direction: column; gap: 5px;">${layer.flexIconBtns.map(btn => {

                return mapp.ui.elements.buttonIcon(Object.assign(btn, {
                    onclick: e=>{

                        if (e.target.classList.contains('active')) {

                            e.target.classList.remove('active')
                            btn.panel.classList.remove('active')
                           return;
                        }

                        e.target.classList.add('active')
                        btn.panel.classList.add('active')
                    }
                }))

        })}`

        return panel
    }

})()
```

The config for this is:

```js
"flexIconBtns": [
  {
      "label": "foo",
      "icon": "https://geolytix.github.io/MapIcons/adidas/adidas_24px.svg"
  },
  {
      "label": "This is a long title",
      "icon": "http://localhost:3000/latest/public/icons/locator.svg"
  },
  {
      "label": "This is a much much much much much much longer title than I would ever expect",
      "icon": "https://geolytix.github.io/MapIcons/adidas/adidas_24px.svg"
  },
  {
      "label": "From dataurl",
      "icon": "data:image/svg+xml,%3Csvg width='1000' height='1000' viewBox='0 0 1000 1000' xmlns='http://www.w3.org/2000/svg'%3E%3Ccircle cx='500' cy='500' r='350' stroke='%231F964D' opacity='0.8' stroke-width='75' fill='none'%3E%3C/circle%3E%3Ccircle cx='500' cy='500' r='200' fill='%231F964D' opacity='0.8'%3E%3C/circle%3E%3Cpath stroke='%231F964D' opacity='0.8' stroke-width='75' d='M500,150L500,0M500,850L500,1000M0,500L150,500M850,500L1000,500' /%3E%3C/svg%3E"
  }
]
```

dbauszus-glx:
It may be much easier to add the `.popup` to buttonIcon.

This way you can control the behaviour with the `button > .popup` selector.

```js
mapp.utils.html.node`
${mapp.ui.elements.buttonIcon({
    "label": btn.label,
    "icon": btn.icon,
    onclick: (e) => {
        e.target.classList.toggle("active");
    },
    popup: mapp.ui.html`
    <h3>Foo</h3>
    <p>Lorem Ipsum`
})}`
```

eo-uk:
The button and the panel are both meant to be standalone elements. I avoided any logic combining the two in case, for example, we'd like the popup panel to appear when another type of element is hovered.

The panel popup container is to give it relative position (to make it self contained instead of adding the relative to the button)

stopPropagation is to prevent clicks on the panel affecting its parent (button in this case). Without it, button click event can fire on certain clicks.

to remove bottom: -7.5em for example, we'd need to add a similar margin instead. It is an arbitrary number but otherwise the panel and the button overlap. Can make it a param instead?

Multiple open panels overlapping is meant to be addressed in the logic. As you said, the onclick event can clear other panels if that's the desired behaviour.

If you prefer to combine the button and panel into one element I can refactor it to that end.

dbauszus-glx:
Yes, please try to combine the popup with a button. A popup doesn't make sense without an element to toggle the popup. You wouldn't have a popup somewhere and a toggle for the popup somewhere else. But a popup by definition needs to be tied to some control element for the popup.

dbauszus-glx:
While a button can exist without a popup, a btn-popup can not exist without a button. It is hard to manage the clickevents of a panel contained within a button. It's probably easiest to provide a button to the popup btn method and both elements get nested within a container.

```html
<div>
  <button>
    <h3>Title</h3>
    <div>Icon</div>
  </button>
  <div class="btn-popup">
```

```css
.btn-pop: {
  display: none;
}

.active > .btn-pop: {
  display: block;
}

```

dbauszus-glx:
If you set a single value e.g. style it is not required to use a template tag inside a template tag.

This works.
```js
style=${params.style || ''}
```

The template tag is only needed if you need to merge content. e.g.
```js
style=${`color: red; ${params.style || ''}`}
```

I want to provide the panel as an optional element single element rather than providing panel title and content.

e.g.
```js
        ${params.panel && mapp.utils.html`
            <div class="panel">${params.panel}</div>`}
```

The active class toggle should be native to the onclick event. A callback should be provided for additional event logic.
```js
        onclick=${e => {
            e.target.classList.toggle('active')
            params.callback(e)
        }}
```

This way I can provide a logic that toggles sibling events like so if required:
```js
function callback(e) {

            e.target.parentElement.querySelectorAll('button')
                .forEach((el) => {
                    el != e.target && el.classList.remove('active')
                });

        }
```

uhtml is self closing. the </button> at the end is not required.

The .btn-panel should not be relative positioned as it will limit the width of the panel to the width of the button. This can be set in the style if needed.

image

I did a commit with the changes.

This works now pretty smooth.

[Screencast from 30-03-23 17:01:56.webm](https://user-images.githubusercontent.com/22201617/228896366-f5f79dba-06a1-4666-8bfc-7d78fc88e733.webm)

PR #672, opened 2023-03-30
Shortcircuit the mousedown event if longclickmethod is falsy.

dbauszus-glx:


cityremade:
just tested with my custom mapview highlight interaction.
When I set longClickMethod: null I no longer see the cursor waiting 👏

cityremade (commented):
as in comment;

PR #673, opened 2023-03-30
ensure that the in  field filter is deleted, not the filter type

dbauszus-glx:


freddie-wallace (approved):
Tested locally and works as expected.
Thanks.

PR #674, opened 2023-03-31
Sum features properties values for graduate theme on geojson cluster.

dbauszus-glx:


simon-leech (approved):
Tested and all working as expected - happy to merge

PR #676, opened 2023-04-03
Assign style instead of style.icon

dbauszus-glx:
Assigning style.icon instead of style for a theme category style may create an invalid icon style.

Assigning the style.icon to the style.default will not change the default icon. The style itself must be used in the Object assign to overwrite the default icon style with the style from the theme category.

```js
style.default = {
  color: 'red'
  icon: {
    type: 'pill'
    color: 'blue'
  }
}

style.icon = {
  type: 'pill'
  color: 'red'
}
```


simon-leech (approved):
I've tested this on geojson and cluster layer types and seems to be working as expected.  This in particular fixed an issue with geojson icon not being displayed in the legend, so looked good to me.

PR #678, opened 2023-04-04
Workspace root plugins

dbauszus-glx:
An array of plugins defined in the workspace root will concatenated with every locale in that workspace.

In case of multiple locales using the same set of plugins it is possible to define these plugins in the workspace root rather than for every locale.


RobAndrewHurst (approved):
Tested this with a workspace that uses one plugin array on the root of the workspace

And as you can see the plugins load across multiple locales

image

image

PR #680, opened 2023-04-05
Add link to default view from admin panel.

dbauszus-glx:
image


RobAndrewHurst (approved):
Nice! tested locally and deployed! Works as expected!

PR #683, opened 2023-04-05
Check whether the query response in text entry module is null.

dbauszus-glx:


simon-leech (approved):
This looks good to me and fixes the reported error 👍

PR #684, opened 2023-04-05
Merge locale templates

dbauszus-glx:
A JSON template of the same name as the locale key will be merged with the locale.


simon-leech (approved):
Tested and this is working as expected, allows you to have a smaller workspace with separate locale files. 

Secondly, you can also merge a set of default layers too - reducing code duplication.

PR #686, opened 2023-04-12
Cluster format, Geojson properties, WKT fields, and field templates.

dbauszus-glx:
All request parameter are now validated, **once**.

Templates do not have to be specified. A subquery template will be applied if the field name matches a template key in the workspace. e.g. Instead of label.template, cat_template, etc. the template key can be assigned as field value.

## Cluster

All cluster properties can be defined in a cluster config block. The cluster config will be created in the mapp cluster module if not defined. All cluster config values are sent with a request to the cluster endpoint. The XYZ cluster mod does not require to see the cluster config or theme.

```js
  // Assign legacy cluster properties to cluster config object.
  layer.cluster = Object.assign(layer.cluster || {}, {
    label: layer.cluster_label,
    kmeans: layer.cluster_kmeans,
    dbscan: layer.cluster_dbscan,
    resolution: layer.cluster_resolution,
    hexgrid: layer.cluster_hexgrid
  })
```

The `aggregate` value is no longer determined in the cluster layer mod. The cluster layer request will send `'sum'` as aggregate value for graduate type themes if the aggregate is not explicit in the theme config.

A cat array will only be sent if the aggregate value is `'array_agg'`.

The cluster geometry will be sent as array `[x, y]` instead of an object.

`size` has been removed from the cluster request and endpoint as this can be determined.

A cluster location will only ever have a cat and label property. Cluster are verbose and hard to query. Cluster should only be used if the payload would exceed the WKT format. A cluster request will be sent for every mapview change.

## Geojson

Geojson is the most verbose format. All properties must be specified as a `layer.properties[]` array. This was previously an object where the first value is the property key and the value is the property field to look up in the database table. The endpoint will check whether a properties[] array value matches a workspace template and assign the template as field in the in the query to the database.

## WKT

WKT is the most compact format and is recommended if clustering is required and mvt is not an option. WKT does not reload on viewport changes. WKT reloads if the filter or fields change. The WKT dataset will only send properties required for the render. e.g. a field for the current theme or a label. Changing the theme requires a reload of the WKT dataset.

### DIFF
--- a/mod/layer/mvt.js
+++ b/mod/layer/mvt.js
@@ -57,6 +57,28 @@ module.exports = async (req, res) => {
     mvt_fields.push(`${layer.style.label.field} AS label`)
   }
 
+  function getField(theme) {
+
+    if (theme.fieldfx) {
+
+      console.warn('fieldfx is no longer supported for themes. please use field with a template lookup.')
+    }
+
+    if (typeof theme.fields === 'string' && theme.fields) {
+
+      console.warn('theme fields must be an array.')
+    }
+
+    if (Array.isArray(theme.fields)) {
+
+      return theme.fields.map(field => `${req.params.workspace.templates[field]?.template || field} AS ${field}`).join(', ')
+    }
+
+    if (!theme.field) return;
+
+    return `${req.params.workspace.templates[theme.field]?.template || theme.field} AS ${theme.field}`
+  }
+
   const geoms = layer.geoms && Object.keys(layer.geoms)
 
   var geom = geoms && layer.geoms[z] || layer.geom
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
Removed fieldfx from MVT themes.

simon-leech (approved):
I have tested wkt layers on this PR.
I have tested adding fieldfx and get a console warning for cluster layers and this is working correctly. 

I have tested fieldfx on mvt layers, and get a console warning for this. I have updated fieldfx to be defined as field > looking up to a template in the workspace and this is working too. 

I think this is all working as expected :)

PR #687, opened 2023-04-17
Update Geoms warning to Include Layer Name or Key

simon-leech:
I noticed that when you don't provide geom or geoms to the layer - the console warning doesn't specify which layer this is on. 

This is problematic if multiple layers are turned on by default on load, as it's hard to debug. 
This PR adds the layer name is provided, or layer key if not to the console warning - to simply aid debugging as you can quickly locate the layer of issue


RobAndrewHurst (approved):
Looks good to me!

PR #688, opened 2023-04-17
Update Filter Wording

simon-leech:
I noticed that the Filters use - 
'Greater than' and 'Lesser than' which isn't correct English. So I updated to use 'Less than' instead. 

I noticed that when applying a filter - 
the wording is 'Clear all Filter', so I updated to use 'Clear all Filters' instead.

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -135,7 +135,7 @@ async function filter_numeric(layer, entry){
       layer.filter.current[entry.field].gte = Number(e.target.value)
       applyFilter(layer)
     },
-    label_b: 'Lesser than',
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
While you are at it. Can you add the less than and greater than to the filter dictionary to allow for these text entries to be translated.

simon-leech:
@dbauszus-glx good shout! Added the dictionary definitions and updated 👍

RobAndrewHurst (approved):
Looks good to me!

RobAndrewHurst (approved):

dbauszus-glx (approved):

PR #690, opened 2023-04-18
grid aggregation cluster optimisation; dbscan kmeans removal.

dbauszus-glx:
ST_Transform statements have been removed from the cluster module. The module will warn if the SRID is not 3857 for grid aggregation cluster (default).

A cluster.viewportLimit param has been introduced. The viewport will not queried if the mapview map Z is below the limit.

The reason being that a query on millions of points is slower if limited to a large viewport. It takes longer to cluster 5 out of 6 million locations than all 6 million locations.


dbauszus-glx:
kmeans and dbscan cluster algorithm have been removed. These format modes can never perform as well as Openlayers clustering on smaller datasets.

Larger datasets require grid aggregation cluster which is limited to EPSG:3857 point geometries for performance reasons.

dbauszus-glx:
Added validation for the query method type in response to the code scan alert.

simon-leech:
@dbauszus-glx testing this now and noticed that the console warning used layer.key instead of params.layer.key - I committed to this PR for this, so I can continue testing :)

dbauszus-glx:
@simon-leech well spotted, thanks.

simon-leech (approved):
I've tested this PR on cluster layers, I can confirm the console warning is working correctly when the SRID is not 3857. 
The db_scan and kmeans removal is working as expected too. 

Happy to merge

PR #691, opened 2023-04-18
Update to cluster config object

simon-leech:
I noticed that the default for the cluster layer is still assigning cluster_resolution not cluster: {resolution} to be in line with the new method of assignment using a cluster object. 

I also added the assignment to first check for a cluster object and use the values from within that, otherwise to use the older legacy method values. 

I added a console warning if the old method is used to tell the developer to update to use cluster:{}


dbauszus-glx:
I added the warning to this PR. https://github.com/GEOLYTIX/xyz/pull/690

PR #696, opened 2023-04-20
Update Console Warning

simon-leech:
This PR simply amends the console.warning about layer.edit moving to layer.draw. 

The console warning now provides the name of the layer or the layer key, and tells the developer to update to use draw:{}


dbauszus-glx (changes requested):
let's only use layer key not layer.name or layer.key

every layer has a key and it is unambiguous. i don't want to fish a layer name from a combination of templates.

PR #698, opened 2023-04-21
fix for geometry entry modify interaction

dbauszus-glx:
The modify method should receive the entry object as a layer like object instead of the options argument.


simon-leech (approved):
Tested this on both point and polygon geometry editing, and working as expected. Happy to merge

PR #699, opened 2023-04-21
added flexbox css to map legend

eo-uk:
Added css rules to optionally make legends use flexbox instead of grid.


dbauszus-glx (changes requested):
Can you apply this to the legend script?

Ideally I would like the container to be a flex class element. This could be returned by an element util to which you provide an array of elements are argument with options to control the flex behaviour, e.g. the direction.

We want to get away from a legend being grid or flex. But a grid container being grid, and a flex container being flex. And a legend can choose which class container to use.

simon-leech (approved):
I've tested this in both layer config and also in legends for custom views and working as expected 👍

dbauszus-glx (changes requested):
lib/ui/layers/panels/style.mjs

You don't need to overload the call to the legend method. The layer is provided as first argument, all members of the layer such as style.theme.legendOptions can be accessed in the legend method without the additional argument.

lib/ui/layers/legends/categorized.mjs

Instead of the legendOptions argument you can access these from the layer. layer.style.theme.legendOptions. These don't need to be optional since the method would not be called without the layer having a style.theme configuration.

We don't want to have clashing classes.

The container should either be `.grid` or `.flex`. But not `.grid.flex`

`.grid` inside `.flex` or vice versa is fine.

image

dbauszus-glx:
Changed `type` to `layout` and `alignContent:'center'` to `alignContent:'centered'` to be in line with the class.

This way the key values can be directly applied as class names.

dbauszus-glx (approved):

PR #700, opened 2023-04-21
transform feature according to layer srid in drawing interaction.

dbauszus-glx:


simon-leech (approved):
Fixes the described issue- good to go!

PR #706, opened 2023-04-25
Create a geojson layer with features array

dbauszus-glx:
A geojson layer can be added with an array of geojson features instead of a table definition.

The features can be styled with feature properties. It is not possible to highlight or get locations from this layer since it has no source.

It will not be possible to set filters since the features are not gotten from a source.

### DIFF
--- a/lib/layer/format/geojson.mjs
+++ b/lib/layer/format/geojson.mjs
@@ -2,7 +2,11 @@ export default layer => {
 
   layer.format = new ol.format.GeoJSON();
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
I have reviewed this PR and confirm it is working as expected. 

I have also written this up in the custom views section of the Config Docs as a nice way to get geometries from a location as layers in custom view maps 
https://github.com/GEOLYTIX/xyz/wiki/Configuration#adding-geometries-as-layers

PR #709, opened 2023-04-26
Add console warning for draw.delete to layer.deleteLocation

simon-leech:
I noticed that when moving from the old configuration of 

`"edit":{ 
"point": true,
"delete": true
}` 
to 
`"draw":{ 
"point": true,
"delete": true
}` 

The delete method is actually now defined as layer.deleteLocation, so this PR adds a console warning to tell the developer to update from draw.delete to layer.deleteLocation

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -23,6 +23,10 @@ export default async layer => {
 
   if (layer.draw) {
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
Doesn't need the ? check inside the if, but better safe than sorry. Will merge this in.

PR #710, opened 2023-04-26
The filter panel should not add filter from skipped entry.

dbauszus-glx:
If infoj entries with filter which are skipped because they are in the `layer.infoj_skip[]` should not be added to the filter panel drop down.


freddie-wallace (approved):
Understood and working as expected

PR #718, opened 2023-04-27
prepend render container to mapview map target

dbauszus-glx:
If appended to the mapview.Map target element the render canvas will display below the mapview.Map. This can be observed in views where the Map element is not set to 100 view height (vH).

The container is returned from the render method and must match the dimensions of the ol-viewport element.

The mapbox-gl and maplibre-gl can prepend the container element to the mapview.Map target. The container will be below the ol-viewport like this.


eo-uk (approved):
Removed the excess maps from under the panel. Looks good.

PR #733, opened 2023-05-02
Check whether template is object before iterating through keys.

dbauszus-glx:


simon-leech (approved):
This fixes the issue reported - happy to merge

PR #735, opened 2023-05-02
Responsiveness fix for legend flex layout

eo-uk:
Fixed a responsiveness issue where the switch button would be treated as part of contents.


dbauszus-glx (approved):
image

lgtm

PR #737, opened 2023-05-02
MVT Clone Issues (Ongoing and need further help)

simon-leech:
1. If the featureLookup queries returns nothing, the whole layer is drawn. In this PR i've got it so it removes the style drawer and then disables the tickboxes. 
2. You should be able to use the layer style, rather than defining the style in the infoj entry too. In this PR, i've got it so that it takes the layer style, but please can we amend this so it takes the entry style if provided, otherwise uses the layer style (to remove duplicated code). 
3. The data isn't drawn to the layer even though information is returned from the featureLookup query. In my PR, i've got it console logging this to test.

### DIFF
--- a/lib/ui/locations/entries/mvt_clone.mjs
+++ b/lib/ui/locations/entries/mvt_clone.mjs
@@ -72,7 +72,7 @@ export default entry => {
     // Create a style panel as entry.view and append to entry.node.
     entry.view = mapp.ui.layers.panels.style(entry)
     entry.view && entry.node.append(entry.view)
-
+    
     // Add the clone layer to the mapview.Map.
     entry.mapview.Map.addLayer(entry.L)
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
I added a logical assignment for the entry.style from entry.Layer.style.

The show method will simply disable the entry.input and return before the layer.view is ever created.

simon-leech:
I also just committed back to this repository to fix any issue i noticed whereby is the featureLookup query returns null, the checkbox is rightly disabled. But then, if you zoom out beyond the zoom extent and back in, the checkbox becomes enabled but doesn't do anything - as no data. So i added a check for featureLookup in the chkZoom function and return when it's null to prevent this.

dbauszus-glx (approved):
Ah yes, well spotted. check for featureLookup needs to be in chkzoom method as well.

PR #738, opened 2023-05-03
control layer container visibility for 3rd party libraries.

dbauszus-glx:
The layer container for maplibre, mbtiles, and mapboxgl layers should initially be invisible and turned on when the layer is displayed. Otherwise the render canvas will be visible when no layers are displayed.

The container will not be visible again if the layer is toggled off.

Remove GMAP element from default view which was from a google maps test.


RobAndrewHurst (approved):
Looks good to me!

PR #739, opened 2023-05-03
Z filter.

dbauszus-glx:
Setting a field z_field in an MVT or cluster layer config will add a condition to filter geometries where the z_field [int] is equal or larger than the z param provided with query.


simon-leech:
@dbauszus-glx This is brilliant, I made a few fixes and amendments so will wait for @RobAndrewHurst to review this too - but all working beautifully for me now!

simon-leech (approved):

PR #747, opened 2023-05-16
Distinct True filters on filtered layers

simon-leech:
This PR addresses an issue related to when we use layers that are filtered in the workspace to use filter current. 
If you then apply a distinct true filter on an infoj entry you will see any values from the database table (regardless of if they should have been filtered out or not).  
This PR adds a layerFilter flag to the infoj filter entries configured as so - 
`  "filter": {
                        "type": "in",
                        "distinct": true,
                        "layerFilter": true
                    }`

When layerFilter is provided, the current layer filter will be passed into the SQL query to return distinct values from the database. 

However, I have not yet got this working correctly - so please could i get some help on this one?

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -155,25 +155,40 @@ async function filter_in(layer, entry) {
     // Allow layerfilter to be provided to the query.
     if (entry.filter?.layerFilter) { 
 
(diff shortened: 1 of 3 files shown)


dbauszus-glx:
SQL cannot be sent from client to the host. Only parameters to be substituted in a stored template.

PR #749, opened 2023-05-16
Fix for logout not resolving correctly on instances

RobAndrewHurst:


simon-leech (approved):
I've tested this locally and also on a deployed app and all working well. 
Tested logging out via both the logout button in the button panel and via the admin panel logout button and both working well - happy to merge :)

PR #750, opened 2023-05-16
added sass color vars as css root vars

eo-uk:
This is so plugins can access color variables defined in sass for easy and consistent color management.


RobAndrewHurst (approved):
Shap!

PR #751, opened 2023-05-21
Replace version script with javascript version

RobAndrewHurst:
You can test this by executing `node version.js` in the root of the xyz directory.


dbauszus-glx (approved):
lgtm

PR #753, opened 2023-05-22
Infoj Skip Update

simon-leech:
It was noticed that the infoj_skip method wasn't working correctly when you tried to use group in the current codebase. 
To fix this, the code has been amended to check if each property exists n the infoj_skip array, rather than using OR operator.

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -47,10 +47,17 @@ export default (location, infoj) => {
         && mapp.utils.merge(entry, fieldEntry.merge)
     }
 
-    // Skip entries from infoj_skip array;
-    if (new Set(entry.location?.layer?.infoj_skip).has(entry.key || entry.field || entry.query || entry.type || entry.group)) {
-      continue;
-    }
+  // Skip entries from infoj_skip array;
+  if (
+    entry.location?.layer?.infoj_skip &&
+    (entry.location.layer.infoj_skip.includes(entry.key) ||
+      entry.location.layer.infoj_skip.includes(entry.field) ||
+      entry.location.layer.infoj_skip.includes(entry.query) ||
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
You are right. Checking whether the set of infoj_skip values has a value from a logical condition will not work since the a later condition such as group will never be checked.

Doing checks in a logical sequence feels a little clunky. Can you test this with checking whether infoj_skip is array. Then check for some from an array of values including group. The some checks whether the value from array is included in the set from infoj_skip.

```js
    // Skip entries from infoj_skip array;
    if (Array.isArray(entry.location?.layer?.infoj_skip)) {

      // Check whether some val from this array is in infoj_skip set.
      if ([entry.key, entry.field, entry.query, entry.type, entry.group]
        .some(val => new Set(entry.location?.layer?.infoj_skip).has(val))) continue;
    }
```

simon-leech:
@dbauszus-glx Thanks for the comments, I've used that code block above to amend infoj.mjs and working well :) 
I've also amended filter.mjs to use the same logic to ensure if skipping group this is also skipped in the filter here too. 


Ready for review again

dbauszus-glx (commented):
I think this check needs to be the other way around.

```js
//use 
.filter(entry => Array.isArray(layer.infoj_skip) && ![entry.key, entry.field, entry.query, entry.type, entry.group].some(val => new Set(layer.infoj_skip).has(val)))

//instead of
.filter(entry => !layer.infoj_skip.some(val => new Set([entry.key, entry.field, entry.query, entry.type, entry.group]).has(val)))
```

simon-leech:
This is all working for me now - thanks @dbauszus-glx for switching the filter check :)

dbauszus-glx (approved):

PR #754, opened 2023-05-22
Provider current filter to distinct values query template

dbauszus-glx:
The distinct values query template should allow for a SQL filter.

In order for the SQL filter to be created in the query module the layer and locale key params must be provided with the query request.

The filter module has been altered to provide the locale, layer, and current filter (json) params.


simon-leech (approved):
This is working well for me - fixes the bug reported! 

Happy to merge :)

PR #755, opened 2023-05-22
fixed some svg icons not appearing due to invalid css rules

eo-uk:
Some legend icons were not appearing due to rendering of invalid background-image rules. Wrapping the rule in double quotes fixes this.


simon-leech (approved):
I've tested this across a range of styles and layer formats, geojson, wkt, cluster, icon arrays, svgs, defaults and seems to be working well for me. 

Tested on legend plugins too including the split variable theme and seems to have fixed the issue with the legend there too!

PR #757, opened 2023-05-22
Check whether a geojson layer exists in workspace.

dbauszus-glx:


simon-leech (approved):
Tested and can confirm this is working as expected - returns an error instead of crashing the instance.

PR #769, opened 2023-05-26
remove proxy; google maps format & gazetteer

dbauszus-glx:
The proxy endpoint has been removed. Calls to third party API should be made from the client with the API key secured against a trusted domain. Restricted API access should be selected for Google Maps API.

The Here Isoline entry requires an apiKey to work without proxy.

```js
"type": "isoline_here",
"params": {
  "apiKey": "***"
}
```

Likewise the Here Isoline drawing object requires an apiKey.

```js
"isoline_here": {
  "apiKey": "***"
}
```

The streetview entry requires a Google key.
```js
"type": "streetview",
"key": "***"
```

The Google gazetteer requires a key for which the Google Places API is enabled.
```js
"gazetteer": {
  "provider": "GOOGLE",
  "key": "***"
}
```


dbauszus-glx:
requestHeader: null will prevent the default content-type request header to be added.

```js
  mapp.utils.xhr({
    url: `https://maps.googleapis.com/maps/api/streetview/metadata?location=${lnglat[1]},${lnglat[0]}&source=outdoor&key=${entry.key}`,
    requestHeader: null
  }).then(response => {

      if (response.status !== 'OK') return;

      const src = `https://maps.googleapis.com/maps/api/streetview?location=${lnglat[1]},${lnglat[0]}&source=outdoor&size=300x230&key=${entry.key}`

      node.append(mapp.utils.html.node`<img src=${src}>`)

    })
```

RobAndrewHurst (approved):
Nice! Works well!

Good to finally get rid of the proxy module!

PR #772, opened 2023-05-30
check whether locales are available in view mod

dbauszus-glx:
The view mod will check whether a user with given roles has access to at least one locale. The login view is returned with a message if locales.length is falsy.

### DIFF
--- a/mod/user/login.js
+++ b/mod/user/login.js
@@ -73,6 +73,8 @@ async function view(req, res, message) {
     msg: message || ' '
   })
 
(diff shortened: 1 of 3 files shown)


mod/user/login.js
RobAndrewHurst, line 1: @dbauszus-glx can we remove this line?

RobAndrewHurst: Works well!

image

RobAndrewHurst (changes requested):

RobAndrewHurst (approved):

PR #773, opened 2023-06-01
allsettled templates instead of crashing the process

dbauszus-glx:
The assignTemplates mod attempts to load templates from the source.

The promise should be rejected if the src has not a valid getFrom method.

The promise.all will throw an exception if one promise is rejected. The load method should be changed to allSettled.

The logger, `LOGS: "templates"`, can be used to log the array of promises and whether they have been fulfilled.

The reason for a rejected promise is most likely a SRC param missing, e.g. `SRC_CLIENT`

image

### DIFF
--- a/mod/workspace/assignTemplates.js
+++ b/mod/workspace/assignTemplates.js
@@ -62,7 +62,7 @@ module.exports = async (workspace) => {
   );
 
   const templatePromises = Object.entries(workspace.templates)
-    .map((entry) => new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
The resolveFrom method should reject a promise which failed to retrieve a template from src.

image

dbauszus-glx:
The assign templates method must filter for template values to assign to the workspace templates object.

The loadPlugins method should also settle for all and not throw duplicate error messages for a plugin load failing.

RobAndrewHurst (approved):

PR #774, opened 2023-06-02
Change user approval expiry check

dbauszus-glx:
User expiry can be set with the env variable `APPROVAL_EXPIRY = "30"`.

The value in days indicates when the account approval will expire.

This can not be done automatically but will only be checked once a user tries to logon.

Previously the check was against the time of approval. Hence it was not possible to extend this but only remove the expiry by removing the approval time stamp. This solution was done in order to prevent adding an additional field to ACL schema.

The approach was quite limited. With this PR a field `expires_on:integer` must be added to the ACL schema if the APPROVAL_EXPIRE environment variable is set.

A day integer will be set when a user registers. Resetting the password will reset the day integer.

A the expiry date will be checked against the current date on login. The approval will be removed if the account has expired. The login process will abort.

The admin view will show the date in red if today or in the past. An admin can clear the date or set a new expiry date.


dbauszus-glx:
The expiry_on date will not be set on password reset if already cleared.

dbauszus-glx:
There is an issue with the date input toggle. It does not seem possible to focus the editor input unless you click into a different row first. I opened a ticket with Tabulator in this regard. https://github.com/olifolkerd/tabulator/issues/4210

simon-leech:
[User Table Admin - unable to set expiry](https://trello.com/c/ofW54TLS/32-user-table-admin-unable-to-set-expiry)

RobAndrewHurst (approved):

PR #779, opened 2023-06-06
Replace id with dataset id for slider ui elements

dbauszus-glx:
This is to prevent `Duplicate form field id in the same form` warnings.


simon-leech (approved):
That error has been removed with this PR - happy to approve

PR #781, opened 2023-06-08
Gazetteer

dbauszus-glx:
The gazetteer endpoint will be removed.

Gazetteer queries to a layer will use a stored query gaz_query through the query module.

```js
module.exports = `
  SELECT 
  \${label} AS label,
  \${qID} AS id
  FROM \${table}
  WHERE \${qterm}::text ILIKE \%{term}
  \${filter}
  LIMIT \${limit};`
```

This allows for the role and filter logic being applied through the query module instead of having to be replicated in the gazetteer module.

The query module has been expanded to replace wildcard character in substitute parameter strings.

The `wildcard` param will be reserved. Provided with a query a string value, eg. `wildcard=*` will replace any * in the substitute parameter value with a wildcard character `%`.

The gazetteer module has been removed.

mapp.ui.gazetteer requires a gazetteer object with a gazetteer provider and/or datasets being defined. The autocomplete input will be appended to gazetteer target in the ui method. 

Multiple dataset entries can be added to the gazetteer configuration. The `layer` key must be defined for the location from the layer to be selected.

`maxZoom` can be set to limit the zoom for the flyto() location method.

A `leading_wildcard` flag can be set to add a wild card character before the search term.

A `limit` can be set to limit the results. The default is 10.

The `limit`, `leading_wildcard`, and `maxZoom` keys can be set in the gazetteer config itself or in the dataset entry. Dataset entry key values will override the gazetteer block config for the dataset. A `table` must be defined if not implicit on the layer config.

```js
"gazetteer": {
 "maxZoom": 10,
 "leading_wildcard": true,
 "datasets": [
  {
   "layer": "retailpoints",
   "qterm": "store_name",
   "leading_wildcard": true,
   "limit": 5
  },
  {
   "layer": "oa_metrics",
   "qterm": "oa_code",
   "title": "OA Code",
   "table": "geodata.uk_glx_geodata_oa_metrics",
   "maxZoom": 10
  }
 ]
}
```

The oninput method of the gazetteer input will call gazetteer search method from `mapp.utils.gazetteer`. e.g. mapp.utils.gazetteer.datasets for a dataset/layer gazetteer. The query will be re-run with the input value on focus.

A gazetteer can also be defined on a layer. The `mapp.ui.layer.panel.gazetteer()` method will return an autocomplete input to be appended into the layer.view.

It is possible to define multiple datasets to search on different fields in the layer table.

The `table` must be specified if not implicit from the layer config.

`limit` and `maxZoom` values can either be set in the dataset config or in the gazetteer block itself. Dataset key values will override gazetteer block entries.

```js
"gazetteer": {
 "table": "geodata.uk_glx_open_retail_points",
 "placeholder": "Search for Retail Point",
 "limit": 5,
 "datasets": [
  {
   "qterm": "store_name",
   "title": "Name"
  },
  {
   "qterm": "postcode",
   "title": "Postcode"
  }
 ]
}
```

Additional gazetteer method can be assigned as mapp plugins. A sample plugin for a gazetteer using the Google Places API can be found [here](https://github.com/GEOLYTIX/mapp/blob/main/plugins/googleMaps.md).


simon-leech:
[Filter current not being assigned to gazetteer](https://trello.com/c/Rgwg7cEE/3-filter-current-not-being-assigned-to-gazetteer)

simon-leech:
[Gazetteer: Layer-Specific gazetteer not working](https://trello.com/c/UqvlJNw0/42-gazetteer-layer-specific-gazetteer-not-working)

simon-leech:
[Gazetteer - Ability to link to streetview from gazetteer search result location](https://trello.com/c/Sd1lIWoh/55-gazetteer-ability-to-link-to-streetview-from-gazetteer-search-result-location)

simon-leech:
[Gazetteer zoom config](https://trello.com/c/Ob2deK85/41-gazetteer-zoom-config)

RobAndrewHurst (changes requested):
Nice! Works well!

Question! Should we default the placeholder text if non is provided?

dbauszus-glx:
No default. This can be provided from the calling interface, ie. the default view script. The reason being that it should be possible to use the control without a placeholder and that the default script wants to call the placeholder from a dictionary.

RobAndrewHurst (approved):

RobAndrewHurst:
> No default. This can be provided from the calling interface, ie. the default view script. The reason being that it should be possible to use the control without a placeholder and that the default script wants to call the placeholder from a dictionary.

Nice! Didn't think of it that way!

PR #782, opened 2023-06-13
Allow for cluster layer to fade between change:resolution event and render.

dbauszus-glx:
Setting the `layer.cluster.fade` flag will fadeout the layer by 20% with a 50ms interval.

Once data is loaded and features are assigned to the layer source the layer opacity is faded in by 20% with a 50ms interval.

The interval length can be controlled by setting the ms as the `layer.cluster.fade` value.

### DIFF
--- a/lib/layer/format/cluster.mjs
+++ b/lib/layer/format/cluster.mjs
@@ -1,3 +1,5 @@
+import { ListBucketAnalyticsConfigurationsCommand } from "@aws-sdk/client-s3"
+
 export default layer => {
 
   layer.highlight = true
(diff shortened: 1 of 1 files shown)


simon-leech:
[Fade toggle for clustering](https://trello.com/c/fvY17fA8/89-fade-toggle-for-clustering)

simon-leech (approved):
This is working well, tested with both fade:true and fade:value and both working for me. 

added to the config docs https://github.com/GEOLYTIX/xyz/wiki/Configuration#v450

PR #783, opened 2023-06-14
make cluster label for wkt optional

dbauszus-glx:
The WKT cluster should not require a label. The ID will be used instead.


simon-leech (approved):
This is working well for me - happy to merge :)

PR #784, opened 2023-06-14
location.layer should immediately re-style when pin is drawn. this wi…

dbauszus-glx:
…ll remove style:null features.


simon-leech:
[Re-draw layer on pin](https://trello.com/c/RQd9HelS/90-re-draw-layer-on-pin)

simon-leech (approved):
This is working well for me - happy to merge

PR #804, opened 2023-06-27
fixed gazetteer error when datasets layer is missing in mapview 

eo-uk:
Fixed the gazetteer error when a layer defined in datasets caused an error on input if same layer was not added to the mapview.


RobAndrewHurst (approved):
LGTM!

dbauszus-glx:
Is there a valid reason for the gazetteer or dataset to not have a layer assigned, ie. Is the return an indication of a config issue?

In that case a warning should be logged to indicate the issue rather than silently returning.

PR #809, opened 2023-06-29
hooks.mjs parse() rewrite

RobAndrewHurst:
#### 🚨 **Problem:**
There's a risk of a **Regex Denial of Service (ReDoS)** attack due to a [weakness in the regular expressions](https://sonarcloud.io/organizations/geolytix/rules?open=javascript%3AS5852&rule_key=javascript%3AS5852) used in the `lib/hooks.mjs` module. This is because the regular expressions might take a  long time to run for some inputs (super-linear runtime) due to excessive backtracking. 

#### 💡 **Solution:**

1. **Using URL API:**  
   Instead of regular expressions, we now use the **URL API** for handling URL parameters.

2. **Processing URL Parameters:**  
   We create a URL object from the current window's location with the URL API. Then, we loop through each parameter in the URL, processing them one by one. This is a clean and efficient way to handle the information in the URL.

3. **Decoding Parameters Properly:**  
   We replaced the `decodeURI` method with **`decodeURIComponent`**. This new method is better suited for extracting information from URL parameters, making it more accurate.

#### 🛡️ **Benefits:**

- **Security**: The changes strengthen the code against ReDoS attacks.
- **Accuracy**: Picking information out of the URL is now more precise.

### DIFF
--- a/lib/hooks.mjs
+++ b/lib/hooks.mjs
@@ -11,46 +11,41 @@ const hooks = {
   filter,
 }
 
-export default hooks
+export default hooks;
 
 function parse() {
+  
+  const url = new URL(window.location.href);
 
-  // Take hooks from URL and store as current hooks.
-  window.location.href.replace(/[?&]+([^=&]+)=([^&]*)/gi, (match, key, value) => {
-
+  // Iterate over the search parameters
+  for (const [key, value] of url.searchParams.entries()) {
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
Looks great...

I like the use of a web API over a regex.

PR #813, opened 2023-07-04
added option to horizontally display flex legend items

eo-uk:
Added an optional "horizontal" parameter to the graduated legend options that displays items horizontally.


dbauszus-glx:
I will need to see a sample legend configuration to review this.

dbauszus-glx (changes requested):
This can only work with legend.layout = 'flex'.

Following that example it would make sense to use key that describes the effect, ie. legend.orientation = 'horizontal'

Why assign some variables from from the layer object which is provided as argument? This makes it more confusing to understand the code.

I know what `layer.style.theme.legend` is. I have no way of knowing what `legendOptions` are but have to scroll to the assignment. We are trying to reduce ambiguity which is effectively achieved by not using placeholder variables which blow up the code.

```js
  const legendOptions = layer.style.theme.legend;

  const theme = layer.style.theme
```

There ought to be a better way than declaring an empty grid array variable and then use a forEach() array method to push into the grid array if a map() method returns an array.

simon-leech (approved):
All working well for me - thanks!
Config Docs updated here too
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#type-graduated

PR #816, opened 2023-07-12
Multiple execution of infoj process from updateCallbacks array

dbauszus-glx:
The infoj variable defined locally with let in the location view method has been renamed to location.viewEntries to better reflect the nature of the variable.

We prefer variables to be assigned to an object rather than being declared in the closure of a function.

Locations where the layer is flagged to toggle edits assigned an updateCallback method which would recreate the location.viewEntries. This caused the location.viewEntries to be created twice in quick succession. Once from the location.update method which calls the `updateInfo` event, and once from the callback method assigned to remove edits when toggled.

### DIFF
--- a/lib/ui/locations/view.mjs
+++ b/lib/ui/locations/view.mjs
@@ -76,31 +76,25 @@ export default location => {
     // Remove edit from infoj entries
     function removeEdits(){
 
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):
Looks all good to me!

PR #818, opened 2023-07-13
Check whether entries are editable for toggle condition

dbauszus-glx:
The toggle location view edits button should only be added to the location view header if there are editable entries.

Edit roles maybe assigned by roles when the layer is decorated. In that case the toggle button might be shown but no entry is actually editable.

### DIFF
--- a/lib/ui/locations/view.mjs
+++ b/lib/ui/locations/view.mjs
@@ -71,7 +71,7 @@ export default location => {
       }}>`)
 
   // Edit icon.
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
This must also be tested with new locations.

simon-leech (approved):
All working for me! 
Tested this using entry roles and working for a view only role (no editable entries) and also an edit role (editable entries). I also tested on existing locations and new locations.

PR #819, opened 2023-07-13
Fix for persistent mvt_clone entry layer.

dbauszus-glx:
This PR addresses two issues with the mvt_clone entry.

No entry method should re-process unless necessary.

The node variable to hold the entry panel elements to be return to the entry.node was misleading.

The panel node/elements are now constructed as entry.panel.

If the entry.panel exists it will be returned to the entry.node. Meaning the entry.panel and all associated properties of the entry persist during a location [view] update.

### DIFF
--- a/lib/ui/layers/panels/style.mjs
+++ b/lib/ui/layers/panels/style.mjs
@@ -116,13 +116,11 @@ export default layer => {
       layer.mapview.Map.getTargetElement().addEventListener('changeEnd', () => {
         const z = layer.mapview.Map.getView().getZoom();
 
-        const chkBox = layer.view.querySelector('[data-id=labelCheckbox]')
-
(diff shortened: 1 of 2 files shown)


simon-leech (changes requested):
The MVT Clone is no longer persisting, the checkbox removes it, clearing the locations or removing a location removes it and zooming out beyond the layer extent too - this is all good!

However, I see an error in the console when you have an mvt_clone turned on, and then zoom in or out 

image

dbauszus-glx:
@simon-leech This should be fixed in the latest commit.

I was querying the labelCheckbox inside the zoom check event. However, that checkbox is already assigned to a variable outside the event method closure.

simon-leech (approved):
Thanks @dbauszus-glx  - tested again and all good with me!

PR #821, opened 2023-07-14
Layer / template warnings

dbauszus-glx:
A warning will be issued if the getFrom method cannot be determined from the src.

image

A warning will be issued from the addLayer method when the layer format is undefined or unknown. The layer will not be added to the mapview.layers.

image

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -1,7 +1,7 @@
 export default async layer => {
   
-  // Assign layer render.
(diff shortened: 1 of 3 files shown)


simon-leech (approved):
Looks good to me - tested by creating incorrect file path using unknown keys and also incorrect file path and errors appeared. 

I have pushed a minor change to the console warning just to add `This is likely due to an incorrect file path` to make it very clear to the configuration developer what the fix is :)

image

PR #823, opened 2023-07-17
Add roles filter check for location.get method.

dbauszus-glx:
A role filter assigned to a layer would filter out locations from the data request for the layer, but not for the location.get() method.

This would make it possible to get a location to which a user should not have access according to the roles filter.


dbauszus-glx:
The roles object will returned empty from the Roles.check() method if the user has roles but the roles do not match the roles assigned to the layer config. The check method will now return false if there is a mismatch.

dbauszus-glx:
A role filter object must be added to the filter key in the role object.

A `null` role will simply make the layer available for a given role. Nothing will be merged.

The role object will be merged in the layer decorator.

When iterating through the infoj entries. All roles will be checked and merged. This includes negated roles. Previously only the first found role was merged.

```json
          "roles": {
            "foo": null,
            "!point": {
              "draw": {
                "polygon": true
              }
            },
            "draw_role": {
              "draw": {
                "point": true
              }
            },
            "fascia_filter": {
              "filter": {
                "fascia": {
                  "in": ["Budgens", "Waitrose"]
                }
              }
            }
          },
```

RobAndrewHurst (changes requested):
Can we ensure that fields are validated with validateRequestParams?

dbauszus-glx:
> Can we ensure that fields are validated with validateRequestParams?

Yeah. I will look at this now. Also need to make sure that the role logic is documented proper.

RobAndrewHurst:
> > Can we ensure that fields are validated with validateRequestParams?
> 
> No. We MUST ensure that the field key is validated. This is a potential secure vulnerability. I will look at this now. I also need to make sure that the role logic is documented proper.

Lekker! Thanks!

dbauszus-glx:
The filter field parameter validation should now work.

I tested a bunk filter which got rejected with the correct warning.

http://localhost:3000/api/query?template=wkt&reduce=true&locale=UK&layer=retailpoints&table=geodata.uk_glx_open_retail_points&fields=retailer,store_name&filter=%7B%22retailer()%22:%7B%22like%22:%22b%22%7D%7D

```
"retailer()" didn't pass SQL parameter validation
```

The roles.filter method wasn't working neither. The roles.filter method should only return filters that apply. This was broken when the role as whole being a filter was moved into the `role.filter{}` object.

dbauszus-glx:
WIKI for workspace layer.roles{}

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#roles-1

dbauszus-glx:
Roles.reduce method has been removed from roles mod as it doesn't do anything.

RobAndrewHurst (approved):
**Revised Access Check in mod/location/get.js for Better Error Handling**


I identified an issue with the current access check in the `mod/location/get.js` module that resulted in a type error when attempting to convert undefined or null values to objects. The following piece of code was causing this problem:

```javascript
  if (!Object.keys(roles).length && layer.roles) return res.status(403).send('Access prohibited.')

```

This error occurred when the `layer.roles` property was not defined, causing the `Object.keys()` function to throw an error.

To fix this issue, I added checks for `roles` in the `layer` object. This addition ensures the code does not attempt to convert undefined or null `layer.roles` to an object, thus preventing the aforementioned error.


```javascript
if (roles && layer.roles && !Object.keys(roles).length ) return res.status(403).send('Access prohibited.')
```

Subsequently, I added roles to a layer, then attempted to access the layer via the GET API with a user that does not have the correct role. Here is the result I got, which aligns with the expected behavior:

Access prohibited

cityremade (commented):
When I set a role and open a link with hooks to the layer and the selected location I am getting this message in red: 
Screenshot 2023-08-03 at 11 44 02
Screenshot 2023-08-03 at 11 44 17

No error messages when the link has only the restricted layer hook.

dbauszus-glx:
The everyNegatedRole check is important for these cases.

"!test | [foo]" true
"!test | [test, foo]" false

RobAndrewHurst (approved):

PR #826, opened 2023-07-17
classList option for layer, layer-group, style, filter, and draw panels

dbauszus-glx:
Adding `classList: "expanded"` to the layer config will add this classList to the layer view drawer.

Adding `groupClassList: "expanded"` to the layer config will add this classList to the layer group drawer.

Adding `classList: "expanded"` to either the style, filter, or draw config block will add this classList to the corresponding panel drawer.


simon-leech (approved):
This is a really nice feature, thank you @dbauszus-glx ! 

I have amended the infoj.mjs file to check for entry.expanded and console warn to change this to `groupClassList: "expanded` for consistency :)

PR #830, opened 2023-07-19
Enable logger for DBS; Retire KEY_LOGFLARE

dbauszus-glx:
The env `KEY_LOGFLARE` has been retired.

The configuration for the logger module will be done with the env `LOGGER`.

For logging to logflare the apikey and source must be defined in the `LOGGER` env.

```
"LOGGER": "logflare:apikey=***&source=***-***-***-***-***"
```

For logging to a PostgreSQL dbs the corresponding `DBS_*` env must be set.

The account for the dbs connection string must have write permission.

The table schema for logging in PostgreSQL must be like so:

```SQL
CREATE TABLE IF NOT EXISTS table_name
(
   process varchar,
   datetime bigint,
   key varchar,
   log text
);
```

The LOGGER env can be set like so:

```
"DBS_NEON": "postgresql://dbauszus-glx:***@ep-curly-base-242741.eu-central-1.aws.neon.tech/neondb?sslmode=require",
"LOGGER": "postgresql:dbs=NEON&table=schema.table_name"
```

### DIFF
--- a/mod/utils/logger.js
+++ b/mod/utils/logger.js
@@ -4,40 +4,67 @@ const { nanoid } = require('nanoid')
 
 const process_nanoid = nanoid(6)
 
-module.exports = (msg, log) => {
+const logout = {
+  logflare,
+  postgresql
+}
 
-  if (!logs.has(log)) return;
+const dbs_connections = require('./dbs')()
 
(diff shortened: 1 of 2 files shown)


RobAndrewHurst (approved):
Nice works well!

PR #835, opened 2023-07-25
Update to default gazetteer query

dbauszus-glx:
It must be possible to query a dataset without specifying the datasets array in the gazetteer config. The datasets array is strictly for additional queries.

At it's most basic the gazetteer configuration will look as follows:

The `layer` will define the table to be queried and from where to select a found location.

The `qterm` defines the field which will be queried.

```json
"gazetteer": {
  "layer": "retailpoints",
  "qterm": "store_name"
}
```

Any current layer filter as well as role restrictions will apply to the query.

Additional configuration keys are:

* `table` - Use this table instead of the layer table in the query.
* `placeholder` - Set the gazetteer input placeholder. Can not be used in datasets[].
* `label` - The field which will be returned in the results. By default this is the `qterm`, however it is possible to query one field but display a different field as label in the gazetteer results.
* `title` - The title is shown next to the records in the gazetteer results. Will default to `layer.name`, or `provider`.
* `leading_wildcard` - Whether the search term should be prefixed with a wildcard.
* `limit` - Limits the maximum number of results per search/dataset. Defaults to 10.

## datasets[]
Additional queries are made for each dataset defined in the datasets array. A `placeholder` or `provider` can not be defined in a dataset. The `layer` and `limit` parameter will be taken from the gazetteer configuration if not implicit in the dataset configuration.

```json
"gazetteer": {
  "layer": "retailpoints",
  "qterm": "store_name",
  "leading_wildcard": true,
  "limit": 5,
  "datasets": [{
    "title": "Postcode"
    "qterm": "postcode"
  }]
}
```

image

## provider
An external gazetteer can be queried if a `provider` is defined in the gazetteer configuration. A layer must **NOT** be defined together with a provider.

### DIFF
--- a/lib/ui/Gazetteer.mjs
+++ b/lib/ui/Gazetteer.mjs
@@ -47,18 +47,22 @@ export default gazetteer => {
       return;
     }
 
-    if (Array.isArray(gazetteer.datasets)) {
+    // An external gazetteer provider is requested
+    if (gazetteer.provider) {
 
-      mapp.utils.gazetteer.datasets(e.target.value, gazetteer)
+      // Check whether the gazetteer provider is available.
+      if (!Object.hasOwn(mapp.utils.gazetteer, gazetteer.provider)) {
+
+        console.warn('Requested gazetteer service not available');
+      } else {
+
+        // Query the external gazetteer provider
+        mapp.utils.gazetteer[gazetteer.provider](e.target.value, gazetteer)
+      }
     }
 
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
XYZ Wiki documentation
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#gazetteer

simon-leech (approved):
This is working all good for me - tested and approved :)

PR #836, opened 2023-07-26
The distinct values query must not use the current layer filter.

dbauszus-glx:
A current IN filter would invalidate the results from the distinct values query making it impossible to alter the filter through the interface.

Role filters will still be applied since they are set in the XYZ Query API.


simon-leech (approved):
This is working well - works nicely to distinguish between what the user can see initially and then the ability to change this on the front end. 

@dbauszus-glx Is this written up in the documentation? As if my understanding is correct the IN filter will enable the user to select other options. 
Does the NI filter NOT allow the user to edit the filter (equivalent to a view on the db?) ?

PR #838, opened 2023-07-26
Style labels

dbauszus-glx:
In a similar fashion to `style.theme{}` and `style.themes{}` do we support now `style.label{}` and `style.labels{}`

A dropdown to change the label will be displayed for `style.labels{}`.

The display status of the label will remain when the label is changed.

A checkbox to toggle the label display will be shown above the dropdown.

The layer decorator will assign the first label from the labels list or the label where the key matches a `style.label` string value.

[Wiki documentation](https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#label-1)

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -67,19 +67,26 @@ export default async layer => {
   }, layer.filter);
 
   // Set layer opacity from style.
-  layer.style?.opacity && layer.L.setOpacity(layer.style.opacity);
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
This is a great feature! All working well for me and tested on a range of layer types, tested both label and labels are still working and the associated options. 
Really like the count + field label option - that's a neat one!

PR #839, opened 2023-07-26
Collapse the layer drawer when the layer view becomes disabled.

dbauszus-glx:
The layer view module will check whether a layer is outside it's zoom level range and disable the layer view element.

The layer view drawer should also be collapsed at the same time.


simon-leech (approved):
This is working as expected -collapses layers when the zoom level is beyond the tables values

PR #843, opened 2023-07-28
Long click hover popup

dbauszus-glx:
The longClickMethod assigned to the highlight interaction has been moved to the mapview.

`mapview.allFeatures()` receives the event and mapview as arguments. The event pixel is used to lookup all features and provide a popup to get the feature.

It is possible to override the method with a plugin, thus allowing to alter the popup style if desired.

The `mapview.popup()` method will return true without an argument to indicate that a popup is present in the mapview.

The highlight interaction (`pointermove()`) will break if a popup is present in the mapview.

The highlight interaction (`click()`) will close the popup if present in the mapview.

### DIFF
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -12,6 +12,8 @@ import locate from './locate.mjs'
 
 import popup from './popup.mjs'
 
+import allFeatures from './allFeatures.mjs'
+
 import _highlight from './interactions/highlight.mjs'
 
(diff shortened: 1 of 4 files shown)


dbauszus-glx:
The highlight click method will close any existing popup and return.

The long click method has been moved prior to the popup check.

A long click method will close and create a new popup.

simon-leech (approved):
Tested this across a range of layer types - all working well and tested that the cluster popup is closed when long-click is done. 

Happy to merge :)

PR #844, opened 2023-07-28
Set point style zindex to infinity for highlight feature

dbauszus-glx:
A highlighted feature should always be rendered on top [`zIndex: Infinity`].


simon-leech (approved):
This is all good - happy to merge :)

PR #845, opened 2023-07-28
Validate logger env

dbauszus-glx:
The logger module must check whether the LOGGER env exists and whether the split string value is a valid logout method.


simon-leech (approved):
This fixes the described issue - if no LOGGER is provided then the instance will no longer crash :)

PR #847, opened 2023-07-31
Create a default locale with empty layers array.

dbauszus-glx:
XYZ should not use a default locale with the OSM layer if no locale or locales array are configured in the workspace.

The default locale should be a locale with an empty layers array. The _default view can load an OSM tile layer if the layers array is empty.

The api module should turn `"null"` string params to `null`, `"true"` string params to `true`, and `"false"` string params to `false`.

The api module turns array strings into array params. This should happen in the same iteration with all the other req.params methods.

`workspace.plugins[]` will warn. The correct way to assign default plugins is by defining a default locale with plugins.

`workspace.locales` is optional. The `workspace.locale` will be assigned as `locale` in the workspace locales object.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -97,7 +97,6 @@ module.exports = async (req, res) => {
 
       // Slice square brackets of string and split on comma.
       req.params[key] = req.params[key].slice(1, -1).split(',')
-      return;
     }
   })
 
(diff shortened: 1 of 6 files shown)


RobAndrewHurst (approved):
Looks all good to me!

PR #850, opened 2023-08-01
querycheck & queryparams

dbauszus-glx:
The infoj process will assign queryparams from the layer, and locale, making the assign_queryparams plugin redundant.

With either the querycheck or run flag set in an entry with a query the query will be executed from the infoj method. The entry method will be called after the query response has been assigned as entry.value.

The dataview entry method will assign entry.value as entry.data and will perform checks whether the entry.value is null.

The query will no longer be called from the `json`, `text`, and `textarea` type methods but only from the infoj method.


simon-leech (approved):
I've tested this on an instance with lots of assign_queryparams and queries and all working well - really nice!

PR #851, opened 2023-08-02
Layer show should only be called when not shown.

dbauszus-glx:
The draw methods called the layer show if layer is already shown causing an unnecessary reload.

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -147,7 +147,5 @@ async function trash() {
       id: this.id
     }))
 
-  this.layer.reload()
-
   this.remove()
 }
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -31,13 +31,8 @@ function remove() {
   this.layer.mapview.interaction.type !== 'highlight'
     && this.layer.mapview.interactions.highlight()
 
-  if (this.layer.format === 'mvt') {
-
-    this.layer.L.changed()
-  } else {
-
-    this.layer.reload && this.layer.reload()
-  } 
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
The reload on deleting a location is caused by the location.trash() method calling a layer.reload() and then the location.remove() which in turn also calls a layer.reload().

The reload has now been removed from the trash method.

lib/location/decorate.mjs
```diff
@@ -147,7 +147,5 @@ async function trash() {
       id: this.id
     }))
 
-  this.layer.reload()
```
RobAndrewHurst, line 150: Don't we require this reload when we delete a location from a mvt layer? Such as a catchment statistics. Otherwise the geometry will still remain on the canvas without reloading the layer?

dbauszus-glx: Yes it is. The reload is in the location remove method which is called from the location trash(delete) method.

RobAndrewHurst: So it does delete it from the dataset but it still remains on the canvas!

RobAndrewHurst: https://github.com/GEOLYTIX/xyz/assets/43169158/32223440-b625-4576-b479-bdfcab165c88

dbauszus-glx: This should be resolved in latest commit. MVT layer were not reloaded.

RobAndrewHurst: @dbauszus-glx Shap!

RobAndrewHurst (approved):

PR #855, opened 2023-08-02
Lazy load RTL plugin for maplibre

dbauszus-glx:
image

Solves the issue with RTL label in mapbox vector tile styles.


RobAndrewHurst (approved):
Nice!

Looks good to me!

PR #856, opened 2023-08-03
Consolidate geojson and wkt formats as vector format

dbauszus-glx:
The geojson and wkt layer formats have now been consolidated as vector layer format.

`layer.properties` for geojson are no longer required and will warn. The feature properties will be determined as a field array the same way as it is already done for wkt.

The `formatFeatures()` method can be used to create Openlayer features from wkt or geojson features.

`layer.setSource(features)` will formatFeatures and set the source. This is called from the `layer.reload()` or if `layer.features[]` have already been assigned to the layer.

The WKT now supports `layer.features[]` as well.

The cluster source has been moved into the `layer.setSource()` method. The clustering is now also supported if features are assigned as `layer.features[]`.

The changeEnd method will always be assigned and checks for the cluster property inside the event execution. This allows to turn clustering on and off.

The reduced flag will now be checked from template, or query param.


RobAndrewHurst (approved):
### Code Review: Consolidate geojson and wkt modules! 🔥 

#### General Feedback 🌍
The module efficiently leverages OpenLayers for handling geo-spatial layers in both GeoJSON and WKT formats. The organization of the functions and their respective responsibilities is clear, but maybe there are some areas where we can further optimize and enhance clarity.

#### Points of Appreciation 🙌
1. **Modular Function Design**: Both `geojson` and `wkt` functions are streamlined, keeping the Single Responsibility Principle in mind.
2. **Use of Constants**: Utilizing `formatFeatures`, `formatGeojson`, and `formatWKT` as constants provides easy maintainability. It's a great way to avoid magic strings and numbers.
3. **Helpful Warnings**: Emitting warnings for deprecated or unexpected properties is a very user-friendly touch and will help config developers going forward. 

#### Suggestions for Improvement 🚀
1. **Code Comments**: 
    - While the code is mostly self-explanatory for someone familiar with OpenLayers, more inline comments, especially around complex logic, could potentially enhance clarity?

2. **Error Handling**: 
    - There seems to be minimal not much error handiling, especially around the XHR request. Maybe we could consider adding try-catch or other error-handling mechanisms in and amongst the module?

3. **Use of Ternary**:
    - The line `layer.L = new ol.layer[layer.vectorImage && 'VectorImage' || 'Vector']` uses a combination of logical AND and OR for a ternary-like decision. This could be confusing to those intially reading the module.

#### Closing Thoughts 🌟
This is a well-constructed module, reducing the complexity and reducing our code footprint. Reducing the amount of lines we have to maintain is what we always aim for!
Lekker Stuff!

PR #858, opened 2023-08-04
Move hover to style; Enable multiple hover.

dbauszus-glx:
The `hover{}` configuration has been moved from the layer into the style block.

It is possible to configure multiple hover as `hovers{}`

The hover is the same as the label assignment in the layer decorator.


dbauszus-glx:
`mapp.layer.featureHover(feature, layer)` will be assigned as default `layer.style.hover.method()` if non is provided in the hover config.

simon-leech (approved):
Tested this and really like it - much cleaner config wise and like having hover options!

PR #859, opened 2023-08-07
Assign layer, locale queryparams to layer dataview

dbauszus-glx:
The layer and locale queryparams should be assigned in the `lib.ui.Dataview()` method. This will ensure that layer dataviews have the queryparams assigned.


simon-leech (approved):
This allows the queryparams to be assigned to the layer dataview - happy to merge :)

PR #860, opened 2023-08-07
Check whether a query entry should be skipped

dbauszus-glx:
The skip flags which check whether an entry should be skipped according to the entry.value have been moved into a method to be called after a query response has been assigned as entry.value.


dbauszus-glx:
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#skipentry

simon-leech (approved):
Thanks! All good with me

PR #862, opened 2023-08-08
added logging of failed query

eo-uk:
Failed queries will be logged to the console for debugging.

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -172,8 +172,11 @@ module.exports = async (req, res) => {
   }
 
   const rows = await dbs(query, SQLparams, req.params.statement_timeout || template.statement_timeout)
-
-  if (rows instanceof Error) return res.status(500).send('Failed to query PostGIS table.')
+  
+  if (rows instanceof Error) {
+    console.log("Attempted Query:", query);
+    return res.status(500).send('Failed to query PostGIS table.');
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
console.error should be used to log errors.

the error should be logged in the module.

if the error occurs in the dbs module then it should be logged in the dbs module.

dbauszus-glx (changes requested):
The catch clause in the dbs mod doesn't work.

```js
console.error(msg);
logger(msg);
```

Stepping into the logger call will step into the mod calling the dbs method and resolve the node process if the request is resolved. The logger module will never be reached.

We don't want to log twice or thrice `logger('whatever')` should log whatever in the logger module. This will be logged to console if no other logger output has been defined and to console as well as logger output if the logger output has been defined in the env.

As minimum this must be tested againt the supported logger outputs postgresql, and logflare.

dbauszus-glx (changes requested):
Logging to logflare fails straight of the bat with a circular JSON exception.

Logging to stdout now logs everything.

The logic should be if a key is provided as argument only log the msg if the key is defined in env.LOGS. If no key is provided always log the msg.

dbauszus-glx (approved):

PR #865, opened 2023-08-08
Remove style highlight default

dbauszus-glx:
The empty style.highlight has been removed from the defaults.

The `layer.Style()` method and the mapview `highlight` interaction will no longer call for a re-style of the layer if there is nothing to restyle.


RobAndrewHurst (approved):
Nice!

I like the new use of the spread operator for object merging!

PR #867, opened 2023-08-09
Add a question mark icon when tooltip is provided

simon-leech:
I am aware that tooltip has been possible within the infoj configuration for a long while. 
However, I noticed that when you provide a tooltip, there is no visual cue for the user to let them know if they hover over a title, they can see additional information. 
This PR simply adds the question mark SVG already defined in the XYZ framework to the end of the title - as a visual cue for the user to hover to see additional information. 
I have also updated the docs to add tooltip. 

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#tooltip
image

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -101,16 +101,12 @@ export default (location, infoj) => {
 
     if (entry.title) {
 
-      // if a tooltip is provided, then add a question mark to the title. 
-      let tooltip = '';
-      // This is done so the user is aware that there is a tooltip.
-      if (entry.tooltip) tooltip = mapp.utils.html.node`<div style="height:90%; width:90%;margin-left:1%;" class="mobile-display-none mask-icon question-mark"></div>`
-
       entry.node.append(mapp.utils.html.node`
         <div
           class="label"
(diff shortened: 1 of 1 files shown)


RobAndrewHurst:
The question mark feels too close to the title

dbauszus-glx (changes requested):
I like it.

Can you change div for span.

html.node requires access to the node web api which isn't required for assembling a node.

uhtml is self closing. You can use html`<span>foo` instead of html`<span>foo</span>

Margins and sizing should be provided as em instead of percent to reflect the font size.

The height of the span should be controlled as line-height.

```js
      entry.node.append(mapp.utils.html.node`
        <div
          class="label"
          style="${`${entry.css_title || ''}`}"
          title="${entry.tooltip || null}">${entry.title}
          ${entry.tooltip ? mapp.utils.html`<span style="line-height: 1; margin-left: 0.4em;" class="mobile-display-none mask-icon question-mark">${entry.tooltip}`: ''}`)
```

image

I added a commit with proposed changes.

RobAndrewHurst (approved):

dbauszus-glx (approved):

PR #869, opened 2023-08-11
Ensure that only an existing negated role is checked.

dbauszus-glx:
At current the check will pass if the negated role is undefined, ie. doesn't exist.


simon-leech (approved):
Working well for me and fixes the described issue.

PR #870, opened 2023-08-14
Remove layer defaults from workspace mod.

dbauszus-glx:
The defaults json has been removed from the mod/workspaces folder.

A default `layer.srid` will be assigned in the layer format lib module.

`layer.style.default` will be assigned in the `layer.Style` method.

The mod query will prioritise srid from URL param and use `layer.srid` as fallback.


lib/layer/Style.mjs
```diff
@@ -135,20 +188,4 @@ export default layer => feature => {
...
-  // Check whether the feature referenced in mapview.locations
-  if (layer.mapview?.locations[`${layer.key}!${feature.properties.id}`]) {
-
-    // The feature geometry should not be displayed.
-    if (layer.style.selected === null) return;
-
-    if (layer.style.selected) {
-
-      mapp.utils.merge(feature.style, layer.style.selected, {
-        zIndex: Infinity,
```
RobAndrewHurst, line 148: is the zIndex required on grid layers?

image

dbauszus-glx: No it's not required as the grid layer will be removed https://github.com/GEOLYTIX/xyz/pull/871

RobAndrewHurst: Gotcha!

RobAndrewHurst (approved):

Issue #306, opened 2020-07-08
Dynamic themes with geostatistics


dbauszus-glx:
We want to be able to use dynamic themes based on field statistics of viewport elements.

For vector formats this is relative straightforward.

In https://github.com/GEOLYTIX/xyz/pull/871 we have introduced a field_stats object on the theme.

A plugin can add a custom module which imports the simple-statistics library.

```js
const simple_statistics = import('https://unpkg.com/simple-statistics@7.8.2/index.js')

export default (async function () {

    mapp.utils.simple_statistics = await simple_statistics

    mapp.layer.themes.graduated_jenks = function (theme, feature) {

        // Check whether the field_stats field values have changed.
        if (theme.updated) {

            // Generate jenks bins based on the number of cat in the cat_array
            theme.field_stats[theme.field].jenks = mapp.utils.simple_statistics.jenks(theme.field_stats[theme.field].values, theme.cat_arr.length)

            delete theme.updated
        }

        // rest of graduate theme logic but values are checked against the jenks array not the cat_array values.
    }
})()
```

This can only work for vector layer, not for MVT layer. There are multiple separate issues in regards to the MVT format which must be resolved before dynamic themes are possible with MVT layer.

dbauszus-glx:
Going forward the [geostats]((https://github.com/simogeo/geostats)) module will be used for different classification algorithm.

dbauszus-glx:
The hard bit is to style layers once all tiles are loaded. This can be achieved with the rendercomplete event which must be called only once. The rendercomplete event should only be called by the first `source.tileloadstart` event after a `map.movestart` event.

```js
let features = []
let jenks
let moveStart

source.on('tileloadstart', function () {
  layer.L.setStyle(addFeature)

  if (!moveStart) return;
  moveStart = false
  features = []
  _xyz.map.once('rendercomplete', renderComplete)
})

function renderComplete() {

  if (!features.length) return;
  const series = new geostats(features)
  jenks = series.getEqInterval(10)
  layer.L.setStyle(styleJenks)
}

_xyz.map.on('movestart', () => {
  moveStart = true
})

function addFeature(feature) {
    
  const val = feature.get(field)
    
  val && features.push(val)

  // just lines style.
  return new ol.style.Style({
    stroke: new ol.style.Stroke({
      color: '#333',
      width: 1
    }),
    fill: new ol.style.Fill({
      color: _xyz.utils.Chroma('#fff').alpha(0).rgba()
    })
  })
[...]

PR #871, opened 2023-08-14
Remove grid format, mod, and legend; optional layer qid; vector format viewport; featureFormats util & theme.field_stats

dbauszus-glx:
The grid format has been removed as it is currently bugged. #873 

In order to replace the grid format the vector format must be configurable to reload the data within the viewport when the viewport changes. Vector layer will reload the data from source with a viewport if the `viewport: true` flag is set on the layer. The `layer.reload()` will be triggered from the mapview `changeEnd` event.

Vector layer will use `layer.tableCurrent()` instead of `layer.table` and not load data from source if there is no current table.

Vector feature properties should be spread on the vector feature and not be assigned as `feature.properties.properties`.

The grid format had a dynamic theme.

Dynamic themes require stats for the field used in theme. These stats can only be calculated from a series of field values which will be populated when the features are processed. The `theme.updated` flag will be set allowing the theme method to only process the required field statistics once prior to deleting the flag. #306 

The `featureFormats` have been removed from the vector module and added as mapp.utils.

The optional layer qID bug #874 has been addressed in this issue.


RobAndrewHurst:
The styling on WKT layers doesn't seem to be applying.

image

dbauszus-glx:
@RobAndrewHurst Styling should be fixed in https://github.com/GEOLYTIX/xyz/pull/871/commits/56e27d7b12db20b535c3bafd2e4c072ba879991c

The properties object must not be spread when being assigned to the feature.

RobAndrewHurst:
@dbauszus-glx I noticed a bit of code duplication in the featureFormats.mjs module.

There was an if statement 

```js
if (layer.style?.theme?.field_stats) {
        layer.style.theme.updated = true;
        layer.style.theme.field_stats[layer.style.theme.field] = {
            values: []
        };
    }
```

which I then put into a function `initializeFieldStats`

which is now referenced in both the geojson and wkt functions.

Let me know if you are happy with that!

RobAndrewHurst (approved):
## Code Review for PR #871

😃 I've reviewed the PR and more specifically the module relating to processing GeoJSON and WKT formats using the OpenLayers library. Here's my feedback:

### 1. Clarity & Readability :star2:
The code is clear and well-structured. Functions are named descriptively, which makes the module easy to understand even for someone unfamiliar with the domain specifics.

### 2. Code Duplication :mag:
There's a pattern in both `geojson` and `wkt` functions where we push a value into `field_stats`. While this isn't strictly duplicate code due to the different data structures being processed, it may be worth keeping an eye on this if further operations get added in the future. Consider further refactoring if this becomes more complex.

### 3. Documentation :memo:
Consider adding more comments to explain the purpose of each function, the expected format of its parameters, and any other quirks or nuances that might help a developer unfamiliar with this module.

### 4. Use of Utility Function :+1:
The introduction of the `initializeFieldStats` utility function is a commendable move. This ensures that any initialization related logic is contained within its own function, thus keeping the main functions clean and adhering to the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).:+1:

### In Conclusion
The module is well-structured and makes good use of utility functions to minimize redundancy. Great work overall! 🚀

eo-uk (approved):

eo-uk:
vector.mjs:101 throws below exception with certain url params:

Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getFeatures')

PR #880, opened 2023-08-17
remove streetview from default location.entries

dbauszus-glx:
The streetview entry is dependent on an external API and key and has been moved to the plugin for that API.


RobAndrewHurst (approved):
## Code Review for Object Property Check on `mapp.ui.locations.entries`

😃 I've taken a look at the PR to check the presence of a method on `mapp.ui.locations.entries`. Here's my feedback:

### 1. Use of `Object.hasOwn` :+1:
Great use of `Object.hasOwn` to check for the presence of an own property. It's a safer and more modern way to check compared to `Object.prototype.hasOwnProperty`. This makes the code both future-proof and resistant to potential prototype pollution issues. Good stuff oak🌲!

### 2. Warning Message Clarity :mag_right:
The warning message provides necessary details about the missing method. 

### 3. Use of `continue` :bulb:
The use of `continue` ensures that if a certain condition isn't met, the loop skips to its next iteration. This is a good practice, ensuring that we don't execute further logic for this particular loop iteration.

### In Conclusion
Very Lekker stuff!! 🚀

eo-uk (approved):

PR #884, opened 2023-08-21
Create legendIcon for an icon array.

dbauszus-glx:
The legendIcon method will now return a canvas element for an icon array of multiple SVG.

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#icon

```js
"icon": [
  {
    "svg": "https://geolytix.github.io/MapIcons/pins/mint.svg",
    "scale": 2,
    "legendScale": 0.35,
    "anchor": [0.5, 1]
  },
  {
    "svg": "https://geolytix.github.io/mapp/icons/emblem.svg",
    "scale": 0.03,
    "anchor": [0.5, 1.6]
  }
]
```


RobAndrewHurst (dismissed):
Code Review for `legendicons.mjs`
😃 I've reviewed the legendicon.mjs module related to SVG icons and canvas rendering using OpenLayers. Here's my feedback:

1. Clarity & Readability :star2:
The logic of the function is mostly clear with an understanding of the domain specifics. The separation of icon, line symbol, and poly symbol creation logic helps in readability.

2. Documentation :memo:
While there are comments explaining major blocks, consider adding more detailed comments on the purpose of specific logic branches and expected parameter formats. This can further aid in the readability and maintainability of the function.

3. Error Handling & Edge Cases :warning:
The function does not have explicit error handling, especially for missing mandatory parameters or if an SVG doesn't load. Introducing these would make the function more robust and predictable.

5. Use of Utility Functions :+1:
Breaking down complex conditions or repeated logic into smaller utility functions might enhance readability and testability. The approach to wait for all icons to load before proceeding is pretty good :).

In Conclusion
The function showcases structured logic, especially in handling different types of symbols. Overall, very lekker!! 🚀

RobAndrewHurst:
image

Just thinking on this! What if we have a use case that will create an icon from two different icon templates. In this case 'triangle' and 'circle'. This won't work because if we provide an icon array the icon.legendStyle requires an Icon to have a icon.svg.

image

is this a desired use case we will ever want?

@dbauszus-glx @simon-leech

dbauszus-glx:
The `legendIcon` method must call the `svgSymbols` type method to create an icon url prior to rendering the icon to the canvas context.

The template icons plugin must be async to ensure that templates are loaded before the legends are created.

The plugin execution in the default view script must map the plugins array. Promise.all() can be used to await that any async plugin executions are resolved before continuing in the default script execution.

RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/assets/43169158/862ffb0b-4cf5-4e7d-a785-549c961048e7

@dbauszus-glx 
Seeing some interesting behaviour when toggling between two different themes the icons disappear.

You can replicate this with the icon_array/workspace.json workspace on bugs-testing.

simon-leech (changes requested):
I am seeing the same issues as @RobAndrewHurst : 
- When i swap thematic, some of the legend icons do not display (the first time i swapped the cluster icon did not show, when i swapped back to original theme, no legend icon was displayed). 
- I also see that when you use the template_pin it is too large for the legend icon space, i know that you can manually set the legendscale, but just thinking about this will be a big config change - if a pin is 24px is this now too large then or should this fit into the space?
image

dbauszus-glx:
This wants to be re-reviewed. The legendIcon now set a pixelRatio of 1 ignoring the window scaling.

Images already loaded will not be loaded again with the same loadEvent being assigned multiple times.

Empties will be filtered out from the icons array.

RobAndrewHurst (approved):

simon-leech (approved):

simon-leech:
This is now working for me :)

PR #885, opened 2023-08-22
Add fetch mod; Remove module mod; Update mongo provider.

dbauszus-glx:
The module mod has been removed. This was only used for testing mongodb queries as a module.

There is a plan to implement dynamic module loads in the future but this will be dependent on the experimental node loader.

The mongodb provider has been updated to work without the need to install mongoose drivers. The mongodb provider must not be accessible from the client and should only be used to load a workspace from a mongo data-api.

The `collection`, `database`, `dataSource`, `apiKey`, and data api `resource` must be provided in the workspace env.

```
"WORKSPACE": "mongodb:collection=dev&database=workspace&dataSource=xyz-mapp-dev&apiKey=***&resource=https://data.mongodb-api.com/app/data-vwsmf/endpoint/data/v1/action/find"
```

The fetch mod has been added to replace the module mod.

The fetch mod will work in a similar way to the query mod. Templates are provided as objects.

The resource, and options are provided to the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/fetch).

```json
"shipwrecks_filter": {
  "resource": "https://data.mongodb-api.com/app/data-vwsmf/endpoint/data/v1/action/find",
  "options": {
    "method": "POST",
    "mode": "cors",
    "cache": "no-cache",
    "credentials": "same-origin",
    "headers": {
      "Access-Control-Request-Headers": "*",
      "Content-Type": "application/ejson",
      "apiKey": "***"
    },
  "body": {
    "dataSource": "xyz-mapp-dev",
    "database": "sample_geospatial",
    "collection": "shipwrecks",
    "filter": {
      "feature_type": "Wrecks - Visible"
    }
  }
}
```

Logger require has been removed from the workspace mod as it isn't used in this mod.


RobAndrewHurst (approved):
Code Review for `mongodb.js`
😃 I've taken a look at the mongodb.js module that is used handling a search request. Here's my feedback:

1. Clarity & Readability :star2:
The function flows logically, converting search parameters to a request body and then making the fetch request. The transformation from URL search parameters to an object is straightforward, and the overall structure of the function is linear, which is great for maintainability.

2. Code Duplication :mag:
The function appears streamlined and does not show signs of code duplication. It's focused on its primary task of converting parameters and making a fetch request.

3. Documentation :memo:
Consider adding comments detailing the expected format of searchparams and a brief summary of what the function does. This can be particularly helpful for developers who might not be immediately familiar with the specific functionality.

4. Parameter Validation :exclamation:
Consider introducing validation for mandatory parameters like resource, apiKey, and others. It might be helpful to have default values or error messages when expected parameters are not provided.

In Conclusion
The function showcases a clear approach to making a POST to mongoDB based on search parameters. It's looking lekker! We will keep refining as we get more and more usecases for this 🤓! 🚀

PR #891, opened 2023-08-24
Enable maxlength edit config for type text

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#text

Setting maxlength in the edit config will set the same attribute on the text input.


simon-leech (approved):
Looks good to me - tested on text and textarea inputs. 
A future improvement could perhaps be a popup for the user if they hit the maximum character length 👍

PR #896, opened 2023-08-29
Check for feature.set method

dbauszus-glx:
Vector tile layer features do not have a set method.

### DIFF
--- a/lib/mapview/interactions/highlight.mjs
+++ b/lib/mapview/interactions/highlight.mjs
@@ -287,12 +287,12 @@ export default function (params) {
     // Delete the highlight id from current layer.
     delete mapview.interaction.current.layer.highlight
 
-    // Unset cached style to allow for highlight style.
-    mapview.interaction.current.F.set('Styles', null, true)
-
     // Style the feature itself if possible.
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
Fixes the described bug - happy to merge!

simon-leech (changes requested):
Similar error is seen on highlight in the highlight.mjs file. 
Line 291 
 // Unset cached style to allow for highlight style.
    mapview.interaction.current.F.set('Styles', null, true)

simon-leech (changes requested):
@dbauszus-glx Thanks for that - this is now working for non-mvt layers. 
However, on mvt layers there is a new error. 
This originates from line 204 in highlight.mjs this code
    // Unset cached OL Styles object.
    feature?.F?.set('Styles', null, true).

The error occurs on higlight of an mvt layer - see image. 
image

dbauszus-glx:
@simon-leech I moved the second feature set method into the block which checks whether the method is available.

simon-leech (approved):
This is looking good to me now - no further errors on either style or highlight on mvt / non-mvt layers. Happy to merge!

RobAndrewHurst (approved):

PR #898, opened 2023-08-30
Dynamically adjust the position tabulator dropdown elements. Tabulator 5.5.2

dbauszus-glx:
UL dropdown elements must be positioned fixed in order to be visible outside of their element.

The Tabulator container maybe scroll-able. The fixed elements must be re-positioned in the horizontal scroll event of the Tabulator table element.

The maxHeight should be adjusted in the click event on the dropdown header.

**Tabulator has been bumped to 5.5.2 and is imported as ESM from unpkg instead of skypack.**

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -73,6 +73,13 @@ export default (params) => {
       <div class="head"
         onclick=${(e) => {
 
+          const bounds = e.target.getBoundingClientRect()
+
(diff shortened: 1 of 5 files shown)


dbauszus-glx:
SASS has also been bumped to latest point release 1.66.1

cityremade (approved):
yeah I can see dropdown resized and scrolled when the table is lower.

simon-leech (approved):
This is working well with scrolling the tabulator tables - happy to merge :)

PR #899, opened 2023-08-31
User Expiry Chceck

RobAndrewHurst:
**Bug Report**:

We discovered a bug affecting **non-admin users**. Even when they had no `expires_on` date set (i.e., it was cleared), they were **unable to log in** unless an expiry date was assigned.

---

This has now been fixed with an additional check to see if the users `expires_on` is null.

```js
if (user.expires_on !== null && user.expires_on < new Date()/1000)
```

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -200,9 +200,9 @@ module.exports = async (req, res) => {
 
     if (template.err) return res.status(500).send(template.err.message)
 
-    if (!user && (template.login || template.admin)) return login(req, res, 'Route requires login')
+    if (!user && (template.login || template.admin)) return login(req, res, 'login_required')
 
-    if (user && (!user.admin && template.admin)) return login(req, res, 'Route requires admin priviliges')
+    if (user && (!user.admin && template.admin)) return login(req, res, 'admin_required')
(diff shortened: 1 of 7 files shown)


simon-leech (approved):
I have pushed a commit to this PR that sets the date on the edit to the expires on to the current date, as previously it came up as 01/01/1970. 

I have tested the logic of the expire, admin assigns a new expiry date, re-register and user is re-approved

dbauszus-glx (changes requested):
Non of this documented and therefore impossible to review proper. I will add documentation but already noticed that the process fails straight off the bat with a new ACL.

The verification process must return with an appropriate message if the no admin account is present in the ACL. https://github.com/GEOLYTIX/xyz/pull/899/commits/14507b1198e927e86eb9e96e00e3ab8e930ddcba

I will continue the review, but this cannot be merged as it is.

dbauszus-glx (approved):
Seems to work well enough now.

There is an issue with the date editor which must be addressed with the Tabulator implementation. https://github.com/olifolkerd/tabulator/issues/4286

PR #900, opened 2023-09-01
Do not delete interaction when finishing

dbauszus-glx:
The `mapview.interaction` is assigned after the current interaction is finished, `mapview.interaction?.finish()`

The finish method can call a callback prior to deleting the current interaction.

If the callback method were to enable the highlight interaction this would cause an issue.

The highlight interaction would be enabled and immediately deleted once the callback execution finished and the event loop jumps back into the finish method, deleting now the current highlight interaction and not itself.

The interaction must not be deleted. This is legacy logic from a time where we (object) assigned the interaction and it's parameter argument on initialisation. The interaction is now set with the params argument spread in.


RobAndrewHurst (approved):
Nice! I am happy with this!

PR #903, opened 2023-09-01
Set source to null on vector reload with features === null

dbauszus-glx:
The vector layer source will remain unchanged if a single feature is deleted. The vector layer reload will return null and the setSource method will return with the single feature remaining in the layer source.


simon-leech (approved):
This fixes the described bug when deleting the only location in a layer - happy to merge!

PR #904, opened 2023-09-04
Roles Check

simon-leech:
I noticed that with the previous logic around layer.roles merging, if the user has a single role found in the roles object, then the method loops through all roles in the layer.roles object and merges everything. 
This is incorrect, it should only merge the relevant roles object. 
I have refactored the roles to hopefully make it a little more readable, and this should fix the described issue. 

Please note, i've left the console logs in for testing - but we need to remember to remove as a final commit when we are happy!

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -1,8 +1,8 @@
 export default async layer => {
-  
+
   // Decorate layer format methods.
   await mapp.layer.formats[layer.format](layer);
-  
+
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
Looks good. I renamed roleName to negatedRole to make this more clear and removed the logs.

RobAndrewHurst (approved):
Nice! 

Happy with the logic we added, so that its understood by all :)

PR #905, opened 2023-09-05
Ensure that the draw callback exists.

dbauszus-glx:
This is to ensure that it is possible to add role based editing when there is no drawing method defined otherwise.


RobAndrewHurst (approved):

simon-leech (commented):
All good - fixes the issue!

RobAndrewHurst (approved):

simon-leech (approved):

PR #909, opened 2023-09-08
Allow for empty roles update

dbauszus-glx:
It must be possible to set the roles value to an empty array (remove all roles).

The cellEdited method rolesEdit in the user admin view should check whether the value has changed before sending an update request.


simon-leech (approved):
Fixes the described bug!

RobAndrewHurst (approved):

PR #912, opened 2023-09-12
unnest pseudo elements for input range slider.

dbauszus-glx:
image

I unnested the pseudo elements for the input range slider.


RobAndrewHurst (approved):
Nice! Fixes the bug 🐛

PR #913, opened 2023-09-13
Enhancing Modal Interactivity

RobAndrewHurst:
#### **Changes Proposed**

This pull request aims to fix the issue where form elements inside the modal are not responsive to clicks due to the `e.preventDefault()` calls in the `'mousedown'` and `'touchstart'` event listeners.

#### **📝 Description of Changes**

I have modified the `'mousedown'` and `'touchstart'` event listeners to bypass the `e.preventDefault()` call when the event originates from a form element. This change facilitates the natural functioning of form elements, permitting users to interact with them as intended, without hindering the existing drag functionality of the modal.

Here are the updated sections of the event listeners:

```javascript
params.modal.addEventListener('mousedown', (e) => {
  if (e.target.matches('input, select, textarea, button')) return;

  e.preventDefault();
  params.modal.style.cursor = 'grabbing';
  window.addEventListener('mousemove', shiftEvent);
  window.addEventListener('mouseup', stopShift);
});

params.modal.addEventListener('touchstart', (e) => {
  if (e.target.matches('input, select, textarea, button')) return;

  e.preventDefault();
  params.modal.style.cursor = 'grabbing';
  window.addEventListener('touchmove', shiftEvent);
  window.addEventListener('touchend', stopShift);
}, {
  passive: true,
});
```

#### **🎯 Benefits**

- **Usability**: Enhances user experience by allowing interaction with form elements inside the modal.
- **Accessibility**: Improves accessibility by ensuring all elements function as intended.
- **Maintainability**: Keeps the drag functionality intact, ensuring that other functionalities remain unaltered.

#### **🔎 How to Test**

To test the updates:
1. Open a modal containing form elements.
2. Try clicking into form elements such as input fields and buttons to ensure they are responsive.
3. Drag the modal to ensure that the drag functionality still operates correctly.

Looking forward to your feedback! 💬

### DIFF
--- a/lib/ui/elements/modal.mjs
+++ b/lib/ui/elements/modal.mjs
@@ -31,7 +31,6 @@ export default (params) => {
 
   // Function to handle the start of a drag event, setting up the necessary event listeners for moving and ending the drag
   function handleStartEvent(e) {
-    e.preventDefault();
     params.modal.style.cursor = 'grabbing';
     
     const moveEvent = e.type === 'mousedown' ? 'mousemove' : 'touchmove';
--- a/lib/ui/elements/modal.mjs
+++ b/lib/ui/elements/modal.mjs
@@ -35,6 +35,9 @@ export default (params) => {
 
   // Function to handle the start of a drag event, setting up the necessary event listeners for moving and ending the drag
   function handleStartEvent(e) {
+
+    if (e.target.matches('input, textarea')) return;
+
     params.modal.style.cursor = 'grabbing';
 
     const eventsMap = {
--- a/lib/ui/elements/modal.mjs
+++ b/lib/ui/elements/modal.mjs
@@ -18,8 +18,8 @@ export default (params) => {
       ${params.content}`);
 
   //set the width and height of the modal from where it gets defined.    
-  const { width = '50%', height = '50%' } = params;
-  params.modal.style = { ...params.modal.style, width, height };
+  params.modal.style.width = params.width || '50%';
+  params.modal.style.height = params.height || '50%';
 
   // Function to handle the modal closure
   function closeModal(e) {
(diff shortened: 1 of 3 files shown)


simon-leech (approved):
This is a really nice feature! 
Opens up the modal element to be used for more much than currently possible! 
I've tested with a form element and can confirm that the inputs and buttons are all working as expected within the modal, i can still move the modal and close it.
Happy to merge 🚀🔥

dbauszus-glx (changes requested):
The mousedown and touchstart event method should not prevent defaults. The exclusion of certain elements can then be omitted.

The events should assign the same named method instead of duplicated inline script.

The event methods must only be added once not multiple times whenever the same modal dialog is opened.

It must not be possible to move the dialog completely off-screen.

The dialog should not wrap at the edge off the screen.

RobAndrewHurst:
### Update on RFC ✍️ 

#### **Named Methods**

- **`closeModal`**: I created a function to separate what handles the modal closure into its own named method.
- **`addEventListeners`**: Also created a distinct function for adding event listeners to facilitate drag functionality.

#### **Restrict Off-Screen Movement**

Enhancements were made to the `shiftEvent` method to maintain the modal within the viewport boundaries:
- **Minimum and Maximum Boundaries**: Defined boundaries for the `right` and `top` properties of the modal to prevent it from being moved off-screen.

#### **Avoid Dialog Wrap**

To prevent the dialog from wrapping at the screen's edge, I set constraints on its movement:
- **Boundary Limits**: Set minimum and maximum limits for the `right` and `top` values, ensuring the dialog does not wrap around the screen edges.

#### Addition of Comments 💬 
- I have also added in additional comments to the module for other developers to make further sense of things 🤓 Please let me know if some need more clarification.

#### Still an Issue.
- Ensuring that the events only get added once when the modal gets created! @dbauszus-glx Could you possibly assist me with this logic. I have tried a couple of things but to no avail.

dbauszus-glx (approved):
I have changed the drag event from a mousemove to an actual drag event.

The render performance is better since there is just one check at the end of the event instead of checks during the move event.

This also doesn't require attach and detach events within events.

PR #915, opened 2023-09-18
featureSource & getFeatures

dbauszus-glx:
The mvt layer snapSource has been renamed to featureSource to better represent what it is. It a source which loads features as vector features instead of render features.

Vector layer must have a style object if added through a plugin.

The vector layer source must be accessible from the layer object since the actual source might be an intermediary cluster source.

The vector layer will assign the xhr to the layer object to indicate whether a layer is currently loading data from the [data]source.


dbauszus-glx:
The snap source should have a remove method taking care of issues associated with snapping to the MVT featureSource.

RobAndrewHurst (approved):

PR #916, opened 2023-09-19
Short circuit style assignment for icon with clusterScale

dbauszus-glx:
Eventually it should be possible to theme cluster features with an aggregation. However this will require some work on the cluster mod to provide features with a similar makeup to cluster created from the vector source.


simon-leech (approved):
This is working for me ,  a neat way to ensure that the cluster scale is used on thematics.

PR #919, opened 2023-09-20
Set label and hover; then replace children.

dbauszus-glx:
Introduces setLabel and setHover keys for themes. 

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#setlabel

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#sethover

Please note the use of `.replaceChildren`. Instead of fuffing around with modifying elements it is preferred to recreate a panel from the creator method and then replace the children of the existing panel.

Also fixes a bug where the index lookup for categories in the categorical theme legend wasn't correct.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -79,6 +79,16 @@ export default async layer => {
       : layer.style.themes[layer.style.theme || Object.keys(layer.style.themes)[0]];
   }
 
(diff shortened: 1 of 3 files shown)


simon-leech (approved):
Loving this feature, working well for me! Happy to merge!

RobAndrewHurst (approved):

PR #922, opened 2023-09-20
Vector layer

dbauszus-glx:
The `type:"vector_layer"` will assign the query response as `features:[]` to the entry object. The entry will passed to the vector layer decorator to create a vector layer associated with the entry.

```json
{
 "label": "Nearest to Catchments",
 "type": "vector_layer",
 "format": "wkt",
 "srid": "4326",
 "display": true,
 "query": "uk_site_catchment",
 "queryparams": {
  "id": true,
  "reduce": true
 }
}
```

Style and themes are possible but will require some test data to work with.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -1,6 +1,10 @@
 export default layer => {
 
+  // Assign default SRID if nullish.
   layer.srid ??= '3857'
+
+  // Assign empty style object if nullish.
+  layer.style ??= {}
 
   layer.reload = () => {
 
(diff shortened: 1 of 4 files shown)


dbauszus-glx:
@simon-leech all valid points. I haven't done the documentation yet because I haven't yet decided whether this will be a type in it's own right or consolidated with the mvt_clone to become type:layer.

Label, themes etc are not yet supported since I will need a public use case for this.

The format is the format. If WKT, then it must be st_astext(), if geojson then it must be st_asgeojson(). The default stored queries can be used but need to be tested against label, themes, etc.

simon-leech (approved):

simon-leech:
Understood @dbauszus-glx I've opened a separate issue for those outstanding asks but the main functionality is working!

PR #923, opened 2023-09-21
Filter Labelling Bug

simon-leech:
This PR addresses two bugs: 

1. When you provided an infoj entry with a `filter.title` this title was only applied to the filter provided after you have clicked on the dropdown option - not also to the dropdown title   (see screenshot provided).
``` json 
{
      "title": "Hello",
      "field": "field",
      "type": "integer",
      "filter": {
        "title": "TEST ME",
        "type": "integer"
      },
      "inline": true
 }
```
image
image
image

2. The dropdown.mjs module was incorrectly overwriting entry.title, which meant that updating the dropdown value led to the infoj title value also being updated. To correct this a local variable `dropdown_title` has been created, ensuring that the two issues are resolved.

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -3,15 +3,15 @@ export default (params) => {
   params.selectedTitles = new Set()
   params.selectedOptions = new Set()
 
+  let dropdown_title;
   // Create array of li elements for dropdown.
   const ul = params.entries.map((entry) => {
-
     try {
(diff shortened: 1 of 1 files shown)


RobAndrewHurst:
Point 2. was due to this assignment on the dropdown.mjs `entry.title = decodeURIComponent(entry.title)`

dbauszus-glx (approved):
image

works for me.

RobAndrewHurst (approved):

PR #921, opened 2023-09-20
added refreshing of dataviews when a location is edited

eo-uk:
With this PR, dataviews no longer need to be manually refreshed after a location is edited.

PR #924, opened 2023-09-21
Fix for Editing Locations without Dataviews

simon-leech:
This PR https://github.com/GEOLYTIX/xyz/pull/921 introduced updating for dataviews when locations were edited. 
However, i noticed this errors if a layer has no dataviews. 
As such,
`  Object.values(location.layer.dataviews).forEach(dv => dv.update());`
Has been updated to 
    `  if (location.layer?.dataviews) {
        Object.values(location.layer.dataviews).forEach(dv => dv.update());
      }
    })`

To ensure no error occurs when a location is edited and a layer does not have any dataviews to update.


dbauszus-glx (approved):
I tested this with and without dataviews and it works.

PR #926, opened 2023-09-21
Distinct Edit Options

simon-leech:
It should be possible to get the distinct values from the db for editable options in a dropdown for the user.
This is required to prevent the config needing to be manually updated each time a new option is provided in the db, and is also important for consistency - as currently the filter allows this.

This PR is a draft as not yet ready for review.
However, in my current code:

using toggleLocationViewEdits you have to click it twice for the dropdown containing the distinct values to be rendered for display.
using toggleLocationViewEdits when you press save and the value is stored back to the db, the edit mode button is set to inactive, but the dropdowns remain instead of swapping back to non-editable entries.

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -3,15 +3,15 @@ export default (params) => {
   params.selectedTitles = new Set()
   params.selectedOptions = new Set()
 
+  let dropdown_title;
   // Create array of li elements for dropdown.
   const ul = params.entries.map((entry) => {
-
     try {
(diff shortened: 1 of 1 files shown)


eo-uk:
Changed it so initially a placeholder message is rendered (eg "Loading") and then when api call finished, the actual options dropdown replaces the message.

dbauszus-glx:
That dropdown shouldn't be shown while loading. The dropdown should be rendered into a container when ready to use.

eo-uk:
Dropdown isn't shown while loading, the loading message is.

PR #929, opened 2023-09-22
Cloudinary image and documents method consolidation; Array value location update.

dbauszus-glx:
The `images` and `document` entry types have been consolidated to the cloudinary entry type method to reduce duplication.

The cloudinary method will now immediately update the location making it unnecessary alter the entry content after the update.

It is now possible to set array values in a location update.

The `set_field_array` query has been removed since it is no longer required.

The `skipFalsy` flag will now skip on an empty array.


RobAndrewHurst:
I have found an issue where if you place a new point, you can see the add image/document buttons in the infoj. But once you clear your selection and reselect the point again the buttons to upload will no longer be there.

https://github.com/GEOLYTIX/xyz/assets/43169158/9031708b-a591-4373-9478-a6484577f238

This is also true for any previously existing points on the scratch layer.

lib/location/decorate.mjs
```diff
@@ -42,18 +42,18 @@ async function update() {
 
   const newValues = {}
 
-  this.infoj
-    .filter(entry => typeof entry.newValue !== 'undefined')
-    .forEach(entry => {
+  for (const entry of this.infoj.filter(entry => typeof entry.newValue !== 'undefined')) {
```
eo-uk, line 45: Recommend separating filtering array to a descriptive variable prior to the for loop for better readability

RobAndrewHurst:
I found an issue with the delete of a cloudinary asset. If I upload multiple images, and I delete one. They all get removed.


https://github.com/GEOLYTIX/xyz/assets/43169158/0c47babd-17ee-452d-87c7-f82539df4d9e

dbauszus-glx:
@eo-uk Thanks for pointing out the newValues assignment. This was rubbish. We don't want to assign object in a forEach loop. And the number should not be checked here. There is a specific entry module for integer, and numeric [float] values. This was rubbish as well though.

I map the newValues object from entries like so.
```
  const newValues = Object.fromEntries(this.infoj
    .filter(entry => typeof entry.newValue !== 'undefined')
    .map(entry => [entry.field, entry.newValue]))
```

The numeric entries module has been updated to be in line with how entries modules should be like. The prefix and suffix will not be shown with the input.

@RobAndrewHurst The remove failed because I checked for the set length instead of its size. Hence it always removed everything.

The was also an issue with the location view logic. Edits were removed after the view was built. The edits must be removed before the infoj is populated. The toggleEdit and removeEdits should not fly around but should be assigned to the location object.

dbauszus-glx:
@RobAndrewHurst documents don't have a src but must remove the href from set. Also the response from xhr is nonsense and can be removed. The xhr should be awaited though.

simon-leech (approved):
This is working well for me now - with both toggleLocationViewEdits and without. It now respects that flag and only swaps to edit mode when toggled. Tested adding and removing uploads and working well!

RobAndrewHurst (approved):

PR #930, opened 2023-09-22
Edit Distinct Options

simon-leech:
It should be possible to get the distinct values from the db for editable options in a dropdown for the user.
This is required to prevent the config needing to be manually updated each time a new option is provided in the db, and is also important for consistency - as currently the filter allows this.
Providing the code sample below will run the `distinct_values` query on the db and populate the options dropdown when loaded. Prior to loading a Loading... message will appear to the user. 
```
"edit": { 
"distinct":true, 
"options":[]
}
```

### DIFF
--- a/lib/ui/locations/entries/text.mjs
+++ b/lib/ui/locations/entries/text.mjs
@@ -1,48 +1,66 @@
 export default entry => {
-
   // val maybe the input element with the entry.value.
   let val = entry.value;
 
   if (entry.edit) {
 
     if (entry.edit.options) {
+      
+      // If distinct get the values from the field on the table.
+      if (entry.edit.distinct) {
 
-      val = options(entry)
+        // Create empty container to be populated by the API call
+        const containerEl = mapp.utils.html.node`<div class="val-container">Loading...</div>`;
+        
+        // start async API call to load options
+        handleDistinctOptions(entry, containerEl);
+
+        val = containerEl;
+
+      } else {
+
+        val = options(entry);
+      }
+
     } else {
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
`edit.distinct` isn't required. The options array must be set for options edit. This can be empty indicating that the distinct options must be queried. The distinct options are then stored as options. This will require to query distinct options only once.

The `filter.title` must be removed from the dropdown method as this will break dropdowns not coming from the filter panel. The filter title should be escaped in the filter panel where the dropdown method is called.

The options method should always create container and then render into the container. It's easier to set container, render options into container, or get options if empty, and then render options into the container.

simon-leech:
image
The dropdown isn't populated with values, but you can select and then save them and then the selected value will update.

RobAndrewHurst:
Adding in a placeholder on the entry doesn't seem to be working. 

Also you can't unselect an option once selected. (i.e go back to the placeholder or no value)

dbauszus-glx:
@RobAndrewHurst this should now be working. NULL values must not be filtered out from the distinct values.

simon-leech:
@RobAndrewHurst Good spot - sorry I missed this! I've just tested again and can confirm i'm able to use placeholder on the dropdown, and also set the value back to null :)

RobAndrewHurst:
The placeholder is now working, but I still can't deselect what I have selected. This should happen by clicking on the option you have already selected, which will result in the placeholder displaying in the drop down.

RobAndrewHurst (approved):

RobAndrewHurst (approved):

PR #931, opened 2023-09-25
Feat icon updates

eo-uk:
Event listener in locations view.mjs was relying on icon css class to retrieve the save button for hiding/showing. I added the btn-save class so its icon can be changed without breaking the listener's logic. Also updated the save and edit button icons from a wrench and cloud to pen and checkmark respectively.

We need to make sure code elsewhere isn't broken as a result of icon class change, including any plugins.


dbauszus-glx (approved):
image

lgtm

simon-leech (approved):
This is looking good to me - like the new icons!

PR #932, opened 2023-09-26
Provide filter instead of entry

dbauszus-glx:
Assign the title and field to the entry.filter object and provide the filter object as argument to the filter methods instead of the entry object.


simon-leech (approved):
This is working for me, and the change from `entry` to `filter` aids readability too - nice work!

PR #937, opened 2023-09-26
Assign location style to layer type default style.

dbauszus-glx:
Without a default style the location.style will be assigned as default style.

This applies to the `vector_layer` and `mvt_clone` type entries.

The `location.style` is assigned in the listview.mjs `mapview.locations` setter proxy.

Assigning an svg type in the entry default style will allow to create coloured icons from the `layer.style` fill and strokeColor.

```json
"style": {
 "default": {
  "type": "dot"
 }
}
```

### DIFF
--- a/lib/ui/locations/entries/mvt_clone.mjs
+++ b/lib/ui/locations/entries/mvt_clone.mjs
@@ -16,7 +16,13 @@ export default entry => {
   }
 
   // The entry must have a style object.
-  entry.style ||= entry.Layer.style
+  entry.style ??= entry.Layer.style
+
+  if (entry.style.default) {
+
+    // Assign the location style to the default style.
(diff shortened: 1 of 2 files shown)


simon-leech (changes requested):
I don't think these things are related but I'm seeing some issues with mvt_clone. 

- The `mvt_clone` is returning all data from the cloned layer, not just the data that is returned by the query. 
- Swapping styles of an `mvt_clone` layer returns an error in the console from dropdown.mjs,

simon-leech:
I think these issues are being addressed in other PRs - so i will approve the functionality in this one!

simon-leech (approved):

RobAndrewHurst (approved):

PR #938, opened 2023-09-26
zoom interaction callback from default script

dbauszus-glx:
The default script must delete the current interaction in the zoom interaction callback and check in a timeout whether the highlight interaction should be assigned.

The callback will also be called if the zoom interaction is finished by another interaction, e.g. draw.


RobAndrewHurst (approved):

RobAndrewHurst (approved):
Awesome! I thought this was the case!

RobAndrewHurst (approved):

PR #939, opened 2023-09-27
Dataview Update

simon-leech:
This PR addresses an issue whereby if a layer has `dataviews` but they are all unchecked, ie not turn on. The `dv.update()` fails with update is not a function. 
This PR simply checks and updates only dataviews that are display true - ie turned on.


dbauszus-glx (approved):
looks good, just needs the console.log() removed before merge.

simon-leech:
Pushed commit to remove console log

PR #940, opened 2023-09-27
Ensure that the modify callback enables the highlight interaction.

dbauszus-glx:
The interaction callback must delete the interaction before a timeout to check whether no other interaction is current and then enabling the highlight interaction.


simon-leech (approved):
This is working for me - I am now able to edit a location geometry and immediately select other locations from the mapview. Happy to merge

PR #941, opened 2023-09-27
Ensure that featurelookup returns nought from style method.

dbauszus-glx:
The style method must return nought if no lookup feature is found.


dbauszus-glx:
Please ignore the code smell on this PR. This is what caused the issue to appear in the first place.
Putting the feature lookup conditional into a method cause the method to return into the style method instead of from the style method to filter the feature out.

simon-leech (approved):
This fixes the two described bugs provided - 
1. That mvt_clone was returning all data from that layer, rather than just the query respnse.
2. That mvt_clone style thematic was not working.
Happy to merge!

PR #944, opened 2023-10-02
Numeric Entry Module Refactor

RobAndrewHurst:
### 🐞 Bug Report

**Issue:** 
When assigning 'integer' or 'numeric' as the type for an entry, the system does not format the entry's value appropriately.

**Impact:** 
This can lead to incorrect data presentation and potential issues when users expect a specific format.
None of the values would have a `maximumFractionDigit`
image
**Proposed Solution:** 

I have refactored the module to make it easier to follow the logic, and added in a few helper functions to break it all up. I have also included a function `formatEntryValue` that will format the data accordingly which will solve the bug.

---

### DIFF
--- a/lib/ui/locations/entries/numeric.mjs
+++ b/lib/ui/locations/entries/numeric.mjs
@@ -17,15 +17,17 @@ function formatEntryValue(entry) {
   if (isNaN(entry.value)) return;
 
   if (entry.type === 'integer') {
-  const options = {
-    // Number of decimal places is always 0 for integers
-    maximumFractionDigits: 0
-  };
-  entry.value = parseFloat(entry.value).toLocaleString('en-GB', options);
+
+    const options = {
+      
+      // Number of decimal places is always 0 for integers
+      maximumFractionDigits: 0
+    };
+    entry.value = parseFloat(entry.value).toLocaleString(entry?.formatterParams?.locale || 'en-GB', options);
   } else {
 
     // Convert the value using toLocaleString with the formatterParams defined
-    entry.value  =  parseFloat(entry.value).toLocaleString(entry?.formatterParams?.locale || 'en-GB', entry?.formatterParams?.options)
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
The refactor looks good but there doesn't seem to be an obvious way to control the numeric decimal points. It should be possible to define the maximumFractionDigits in the entry.

dbauszus-glx:
What is the bug to test for?

RobAndrewHurst:
> What is the bug to test for?

I have just updated the Initial comment for this! Sorry I left it in draft until I got to writing what the problem was.

simon-leech (approved):
This is all working for me and fixes the bug. 

I've pushed a commit to add `entry.precision` which will be used for the number of decimal places, or will default to 2 for numeric types. 

Added to the config docs here too https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#numeric-1

RobAndrewHurst:
> This is all working for me and fixes the bug.
> 
> I've pushed a commit to add `entry.precision` which will be used for the number of decimal places, or will default to 2 for numeric types.
> 
> Added to the config docs here too https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#numeric-1

Nice one!!

dbauszus-glx:
@simon-leech Have a look at the use of toLocaleString with options as defined for the numbers in tabulator. A similar format should be used.

simon-leech:
@dbauszus-glx Amended this to use `toLocaleString` and `formatterParams` as per our Tabulator formatting. 

``` json 
  {
      "title": "X",
      "field": "field",
      "type": "numeric",
      "formatterParams": {
        "options": {
          "maximumFractionDigits": 5
        },
        "locale": "en-GB"
      },
      "inline": true,
      "prefix": "£"
    },
```

dbauszus-glx (approved):
looks good now. thanks for updating the wiki.

PR #947, opened 2023-10-04
Enable dataview entry with data from querycheck on update.

dbauszus-glx:
This is to fix an issue where a dataview would be disabled without data and stay disabled during the update event even if data is returned from the query in the queryCheck.


simon-leech (approved):
Thanks @dbauszus-glx - fixes the described bug!

PR #948, opened 2023-10-04
Fix Source clear bug

RobAndrewHurst:
### 🐞 Bug Fix

**Issue:** 
When zooming past the minimum zoom level of an MVT layer, the system attempted to clear the layer's source. However, it was only referencing the source directly rather than accessing it through `layer.source`.

**Solution:** 
Fixed the reference issue by ensuring that we are referencing the source via `layer.source` when attempting to clear it.

---

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -40,7 +40,7 @@ export default layer => {
 
     const tableZ = layer.tableCurrent()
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
Ah crap. I touched this recently and must have missed this.

Looks good. Thanks for the fix.

simon-leech (approved):
Thanks @RobAndrewHurst - all good and fixes the bug!

PR #942, opened 2023-10-02
Edit Mode with Null Geometry

simon-leech:
This PR fixes a bug whereby if a `location.entry.infoj` geometry type field was present and had edit true, but the field was initially empty, in edit mode the edit button would appear. However, this threw an error as you cannot edit a null geometry. 

Instead what this PR does is check if the entry.value is null and if entry.draw is provided, and if so it will show the user the ability to draw the geometry instead - fixing the issue. 

It also adds the ability for `draw.groupClassList` to be provided, allowing the draw button to be shown iniitally and removing a user click. 

``` json 
{
      "label": "Display Catchment",
      "field": "field_geom",
      "fieldfx": "ST_asGeoJson(field_geom)",
      "type": "geometry",
      "group": "Catchment",
      "display": true,
      "srid": "4326",
      "edit": {
        "geometry": true,
        "modifyBtnOnly": {
          "label": "Edit Catchment"
        }
      },
      "draw": {
        "groupClassList": "expanded",
        "polygon": {
          "label": "Draw Catchment"
        }
      },
      "style": {
        "strokeColor": "#A4036F",
        "strokeWidth": 2,
        "fillOpacity": 0
      }
    }
```

PR #949, opened 2023-10-04
Remove duplication of create button when entry geometry is null

simon-leech:
This PR removes a bug introduced. If the location has no `entry.value` and the location is in edit mode, the draw button will appear instead of modify - as fixed in this PR. https://github.com/GEOLYTIX/xyz/pull/942

However, accidentally we were calling the draw method twice for this, once outside the `if(entry.edit?.geometry)` and once within it - using an else when the value was null. This has the unintended side effect of generating the draw button twice when edit mode is toggled on a location with a geometry type of null value. 

This PR simply removes the `else` so that only one button is generated.

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -75,12 +75,7 @@ export default entry => {
 
       list.push(modifyBtn)
     }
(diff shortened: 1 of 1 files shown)


KieronFerrey (approved):
Tested  and the issue is fixed

dbauszus-glx (approved):
lgtm

PR #950, opened 2023-10-05
Vector cluster

dbauszus-glx:
Cluster layer are effectively vector layer with a query that aggregates features in the data source. This is different to clustering on a vector layer where features are aggregated by Openlayers in regards to their pixel distance in the [OL] cluster source.

This PR adds query templates for grid and hexgrid aggregation to the default templates.

In the future it must be possible to request cluster feature from [data]sources better suited to aggregation than PostGIS.

Cluster feature (count property > 1) must not have an ID of any or properties of any single feature within the cluster. The cluster feature itself has no knowledge about the features making up the cluster. This is to preserve bandwidth and keep the data payload at a minimum. The data structure for cluster is well known array with the first entry being a coordinate [x,y] pair (cluster feature will only ever be points), the second being the count, the third the ID, followed by the fields.

The first feature ID in a cluster will be prefixed with an exclamation mark to indicate the cluster ID. This is required for highlighting purposes. Cluster are vector features. Not like MVT render features.

### layer.params

In order to support different queries with different parameter it must be possible to support a custom set of parameter provided with the data request. This a requirement for themes defined in the client and not available in the workspace.

### mapp.layer.fade(layer, out)

The `cluster.fade` flag is now a universal (vector) `layer.fade` flag. The method itself was ported to mapp.layer.fade(layer, out). The flag itself holds the timeout. There is only one method not a fadeOut and a fadeIn method. It wouldn't make sense to fade either way at the same time.

### aggregates

In future it should be possible to define field aggregates as request parameter to return numeric and or arrays from the cluster request.


dbauszus-glx:
There is nothing I want to do about the code duplication these are two similar templates. We are not starting template in template for this.

RobAndrewHurst:
## ✨ Improved Fade Animation

### 📝 Description

This pull request enhances the fade animation function (`fade`) by implementing [linear interpolation](https://en.wikipedia.org/wiki/Linear_interpolation) to achieve a smoother transition between opacities on a the cluster layer format. The fade will occur during the zoom in and out as well as changing the view port.

### 🔄 Changes Made

1. **🎯 Target Opacity Calculation:** Depending on whether fading in or out, the target opacity is calculated. If fading out, the opacity is reduced by 20% or brought to a minimum of 0. If fading in, the opacity is increased by 20% or limited to a maximum of 1.

2. **⏱️ Animation Duration and Timing:** The animation duration is set to 100 milliseconds (adjustable), and the start time of the animation is recorded.

3. **📊 Opacity Update Function:** The `updateOpacity` function is used to calculate the current time and elapsed time since the animation started.

4. **🌟 Interpolation and Opacity Update:** If the elapsed time is less than the animation duration, interpolation is used to smoothly transition the opacity between `currentOpacity` and `targetOpacity`. The `requestAnimationFrame` is used to schedule the next frame of the animation. When the animation duration has elapsed, the layer's opacity is set to the `targetOpacity`, concluding the animation.

5. **🚀 Starting the Animation:** The animation is triggered by calling the `updateOpacity` function.

6. **⏳ Setting New Timeout:** A new fade timeout is set for the next step of the fade animation after the specified duration has passed, ensuring the animation continues until the desired opacity is reached.

### 🧪 Testing

Please ensure to thoroughly test the updated `fade` function to confirm that the fade animation is now smoother and visually appealing.

### 📌 Additional Notes

- The animation duration can be adjusted as needed for different animation speeds.

RobAndrewHurst (approved):

simon-leech (approved):
Tested this out and working for me!

PR #951, opened 2023-10-05
Do not merge HTMLElement

dbauszus-glx:
The deepmerge method must check for items which are instanceof HTMLElement.

Technically objects, these cannot be merged as it will cause the stack to max out and crash the party.

I'll add warning with the element and return false indicating that this is not an object from the object check method.


simon-leech (approved):
This seems to be working correctly for me

PR #952, opened 2023-10-06
Flag for suffix_date for cloudinary image upload

dbauszus-glx:
Setting the `suffix_date` flag in a cloudinary `type:images` entry will add `@date` to the image name.

```
                    {
                        "title": "Pictures",
                        "skipFalsyValue": true,
                        "field": "images",
                        "type": "images",
                        "cloudinary_folder": "test",
                        "suffix_date": true,
                        "edit": true
                    },
```

This is to prevent images with the same name overwriting existing images.

image


RobAndrewHurst (approved):
Nice looks all good to me!

PR #956, opened 2023-10-06
Drawing elements must enable highlight interaction callback

dbauszus-glx:
It is not required to check whether the btn contains the active class. This can always be removed since the callback indicates that the interaction is finished. Likewise the interaction must be removed from the mapview. Otherwise the highlight interaction will not be activated in the timeout.


simon-leech (approved):
Tested and working for me on drawing / editing 😊

PR #957, opened 2023-10-09
Move public/js/queries to workspace/templates

dbauszus-glx:
[query] templates are for the host and should not be stored as js/public. There is no reasons for [query] templates to be accessible  by the client.

The folder has been renamed as templates sitting with the assignTemplates module.

Unused templates have been removed:
count_locations
filter_aggregate
get_last_location
labels
nnearest_locations.sql


RobAndrewHurst (changes requested):
Can we put back the get_last_location query as I use this in API testing.

dbauszus-glx:
@RobAndrewHurst I added the get_last_location template

RobAndrewHurst (approved):

PR #958, opened 2023-10-10
Location mvt cache

dbauszus-glx:
XYZ/MAPP should support different datasources. This will require to replace location methods first with query templates and eventually with fetch templates.

The SQL mvt cache queries should not be baked into the location methods.

The SQL query to find and remove intersecting tiles should be stored as one query and not in multiple places.

The query will be executed from the MAPP library when needed. The method is added to the location (when decorated) and will shortcircuit on a layer which does not have an mvt_cache (SQL) defined.

The tiles must be removed before and after an location update. After a new location is added to the datasource. When a location is removed from the datasource.


RobAndrewHurst:
Is the mvt_cache template used anywhere anymore?

I might be wrong, but I can only see reference of the mvt_cache_delete_intersects.

I guess you can call the template via the api.

dbauszus-glx:
Yeah. It is used to create or truncate a cache table by providing locale and layer as params. I just used/tested it yesterday. It's called manually hence there is no reference. Need to add this to the wiki though. Thanks for flagging this.

RobAndrewHurst:
Nice just tested it too!

PR #960, opened 2023-10-11
Ensure table is defined when table=true passed

simon-leech:
If you pass `queryparams: {"table": true}` to an `entry`, there is a bug where is it currently defined as `undefined`. 

This PR addresses that issue by ensuring it correctly calls `_this.location?.layer?.tableCurrent()`

### DIFF
--- a/lib/utils/queryParams.mjs
+++ b/lib/utils/queryParams.mjs
@@ -1,5 +1,5 @@
 export default _this => {
- 
+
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
lgtm; should make more use of the logical and assignment.

RobAndrewHurst (approved):

PR #962, opened 2023-10-12
Update queryParams.mjs 

simon-leech:
This PR is directly related to the fix in this one #960  . However, i omitted the check for if `table: true` is passed from a `layer.dataview`. This PR adds the check for this as well. So now `table:true` is supported whether the dataview comes from an `entry` or a `layer`.


KieronFerrey (approved):

dbauszus-glx (approved):
Looks good. I only changed to use const instead of let.

PR #964, opened 2023-10-13
fixed rtl plugin error on loading it twice

eo-uk:
RTL Text Plugin errors when multiple maps are on the screen due to being called multiple times. This check fixes the issue.


dbauszus-glx (changes requested):
The plugin should only be loaded after the import in promise. Before the promise resolves. Otherwise is would be possible to load an older version of the maplibre library from the head which would break when trying to set a newer version of the plugin.

dbauszus-glx (approved):
works well now.
image

PR #965, opened 2023-10-16
Ensure tables response is always a string

simon-leech:
In the `location.get` get, it is currently getting the first table in the `tables` object that is not null. However, null is converted to an empty object in XYZ, so presently this returns an error. 
This PR simply adds a check to ensure that the response has to be a string, ensuring objects and null are omitted. 

Example here 
``` 

// example 
const layer2 = {
    "tables": {
    "10": {},
    "11": "schema.table"
  }
}

//  CURRENT METHOD
const layer2table = layer2.table || Object.values(layer2.tables).find(table => !!table)
console.log(layer2table);

// PR CHANGE
const layer2tableNEW = layer2.table ??= 
    Object.values(layer2.tables).find(table => typeof table === 'string')

console.log(layer2tableNEW);
```

### DIFF
--- a/lib/location/get.mjs
+++ b/lib/location/get.mjs
@@ -27,10 +27,10 @@ export default async function (location, list = location.layer.mapview.locations
     return;
   }
 
+
   // Get dbs table from layer.
-  location.table = location.table
-    || location.layer.table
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
lgtm

PR #967, opened 2023-10-17
assign null value in merge util. fire changeend event on start.

dbauszus-glx:
The merge util incorrectly assigned an empty object instead of null.

The changeEnd event must be fired when the mapview is initialised. Otherwise the event would not be fired with no layer on display. The table check would be omitted in that case.

### DIFF
--- a/lib/utils/merge.mjs
+++ b/lib/utils/merge.mjs
@@ -12,6 +12,11 @@ export default function mergeDeep(target, ...sources) {
     // Iterate over object keys in source.
     for (const key in source) {
 
+      if (source[key] === null) {
+        target[key] = null;
+        continue;
+      }
+
       if (isObject(source[key])) {
 
         // Assign empty object on key.
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -244,7 +244,9 @@ export default (mapview) => {
   })
 
   // changeEnd
-  mapview.Map.getTargetElement().addEventListener('changeEnd', () => {
+  mapview.Map.getTargetElement().addEventListener('changeEnd', changeEnd)
+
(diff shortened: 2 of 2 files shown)


lib/utils/merge.mjs
```diff
@@ -5,25 +5,30 @@
...
 
         // Call recursive merge for target key object.
         mergeDeep(target[key], source[key]);
 
       } else {
 
-        // Assign key if not an object itself.
-        Object.assign(target, { [key]: source[key] });
+        // source[key] is neither null nor object.
+        target[key] = source[key]
```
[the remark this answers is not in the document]
dbauszus-glx: This should be resolved by checking the key against the target object prototype.
image

RobAndrewHurst:
image

Found an issue where if you turn a layer on and then zoom past the allowed min zoom the layer will still render. 

Would the desired behaviour be that the canvas is cleared of that layer?

dbauszus-glx:
@RobAndrewHurst should be fixed in last commit.

RobAndrewHurst:
Nice one! Thanks @dbauszus-glx

lib/layer/format/vector.mjs
```diff
@@ -66,7 +66,11 @@ export default layer => {
 
     const table = layer.tableCurrent()
 
-    if (!table) return;
+    if (!table) {
+
+      console.log('foo')
```
RobAndrewHurst, line 71: We will want to remove this

RobAndrewHurst: Removed it!

RobAndrewHurst (approved):
Looks all good to me

simon-leech (approved):
This is working for me and fixes the described bug. Happy to merge!

PR #971, opened 2023-10-18
Error if incorrect configuration

simon-leech:
This PR console errors and returns if a layer configuration incorrectly specifies within the cluster object `cluster.resolution` and `cluster.object`. 
Presently this will render a layer on a map that is unusable - it cannot be selected, highlighted and no warnings occur. 

Instead of trying to amend the configuration in the Framework, we will just return an error to the configuration developer.


dbauszus-glx (approved):
excellent

PR #973, opened 2023-10-19
Add cloudfront logger; signedURL short circuit; Await templates before merge.

dbauszus-glx:
I added the `cloudfront` logger key which logs the response status and url for the cloudfront request.

The signedURL for cloudfront must be returned before the the actual signed URL request.

The custom_templates are awaited before being merged.


RobAndrewHurst (approved):
Very nice!

PR #975, opened 2023-10-19
Fix fetch

eo-uk:
Requests to /api/fetch without a template param were crashing the app. This will send back 400 instead.


dbauszus-glx (approved):
lgtm

simon-leech (approved):
All looks good to me too!

PR #976, opened 2023-10-20
Allow for an infotip render method

dbauszus-glx:
The featureHover callback will check for a render method to render the response as html.

The infotip method allow content as HTMLElement.

The infotip method will not process content objects but will log this content to the console.

The infotip method will warn if string content passed to the innerHTML contains braces.


cityremade (approved):
Just tested and it works as expected :)
Thank you.

PR #977, opened 2023-10-20
Template caching

dbauszus-glx:
tbc


RobAndrewHurst:
image

After multiple tests with 34k requests in 2mins against a testing environment with over 50 locales, all with different amount of layers. The requests to the cached workspace did not fail.

Approved. ✅

PR #980, opened 2023-10-23
Fix hover/hovers and label/labels bug

### LINKED ISSUE #979: Hover/hovers and label/labels
**Describe the bug**

**What?**

Providing both a hover and hovers, and/or label and labels within the layer.style causes a bug whereby the dropdown value is empty as a single hover/label does not need a title.

**Expected behavior**

I’m not sure we should allow both hover/hover and label/labels, if feels messy. I would rather we just console warn if both are provided and remove the single entry.

simon-leech:
Fixes #979 . 

If both `hover` and `hovers` are provided, a console warning is provided and `hover` is removed.
If both `label` and `labels` are provided, a console warning is provided and `label` is removed.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -14,6 +14,7 @@ export default async layer => {
     zoomToExtent,
   });
 
+  // Warn if outdated layer.edit configuration is used.
   if (layer.edit) {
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):

dbauszus-glx (approved):
lgtm

PR #981, opened 2023-10-23
Vector Cluster Changes

### LINKED ISSUE #978: Cluster layer without cluster.resolution
**Describe the bug**
A cluster layer without 'cluster.resolution' fails as it tries to multiply by NaN. 

**To Reproduce**
1. Create a cluster layer without a cluster object 
2.Turn on the layer 
3. See error

**Expected behavior**
If no 'cluster.resolution' is provided, it should be given a default value of 0.1, and then a console warning should appear to tell the configuration developer to explicitly add a value.

simon-leech:
Fixes #978 

If a layer is format cluster, and has no resolution or resolution is not number, it currently fails. 
This PR addresses the issue, checking for this and setting a default value and a console warning if so.

### DIFF
--- a/lib/layer/format/_format.mjs
+++ b/lib/layer/format/_format.mjs
@@ -15,5 +15,6 @@ export default {
   mvt,
   cluster: vector,
   geojson: vector,
-  wkt: vector
+  wkt: vector,
+  vector
 }
--- a/lib/layer/format/vector.mjs
+++ b/lib/layer/format/vector.mjs
@@ -190,6 +190,17 @@ function clusterConfig(layer) {
   // If cluster.resolution is used, the layer srid must be set to 3857.
   if (layer.cluster.resolution) {
 
+    // Check if resolution is numeric.
+    if (typeof layer.cluster.resolution === 'number') {
+      // Assign resolution as float.
+      layer.params.resolution = parseFloat(layer.cluster.resolution);
+    } 
+    // Otherwise, warn and return.
+    else {
+      console.warn(`Layer: ${layer.key}, cluster.resolution must be a number.`)
+      return;
+    }
+
     // Check if srid is set to 4326, not allowed for cluster layer
     if (layer.srid === '4326') {
       console.warn(`Layer: ${layer.key}, srid 4326 is not allowed for cluster.resolution layers.`)
@@ -202,9 +213,6 @@ function clusterConfig(layer) {
 
     // Format is cluster if resolution is set.
     layer.format = 'cluster';
-
-    // Assign resolution as float.
-    layer.params.resolution = parseFloat(layer.cluster.resolution);
 
     // Assign default template.
     layer.params.template ??= layer.cluster.hexgrid ? 'cluster_hex' : 'cluster';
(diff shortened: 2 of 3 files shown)


RobAndrewHurst:
FYI this was removed in this PR.

image

https://github.com/GEOLYTIX/xyz/pull/950/files

dbauszus-glx:
A vector layer may just have a cluster.distance instead of cluster.resolution.

dbauszus-glx (changes requested):
Format:cluster is a legacy format.

Checks where added to the vector format decorator which adds the params for `format:cluster` legacy layers.

It is completely valid to create a cluster vector layer with the params already configured. This is the desired configuration.

This breaks with this PR.

```js
"format": "vector",
"cluster": {
  "resolution": 0.2,
  "viewportLimit": 8
},
"params": {
  "viewport": true,
  "z": true,
  "template": "cluster"
},
```

simon-leech:
@dbauszus-glx Thanks for this, made some changes. 

1. If the layer has no distance or resolution, console.error and return. 
2. If the layer has cluster.resolution, then set the layer.params.viewport and layer.params.z to be true (needed for cluster queries).
3. If the layer has cluster.resolution, set layer.params.resolution to layer.cluster.resolution. 
4. If the layer has cluster.resolution, set layer.params.template to either params.template or cluster by default. 

cluster.hex_grid has been removed as unneeded, you can just pass layer.params.template="cluster_hex". 

Let me know if this is all ok, and I'll make sure the config docs are up to date!
``` js
"format": "vector",
"cluster": {
  "resolution": 0.2
},
"params": {
  "template": "cluster" // "cluster_hex"
}
```

simon-leech:
Added default srid of 3857 to vector format layers with layer.cluster.resolution layers, otherwise its 4326.
Added checks on srid too, updated other checks.

simon-leech:
Thanks for the help @RobAndrewHurst !
This is now ready for re-review 
@dbauszus-glx @RobAndrewHurst @eo-uk

RobAndrewHurst (approved):

dbauszus-glx (approved):
Please review my last commit.

The default SRID for all vector layers is 3857 and should be assigned first.

`layer.cluster?` doesn't need to be conditional within the if `layer.cluster` block.

The layer.cluster.resolution block should be within the layer.cluster block.

There shouldn't be two layer.cluster.resolution blocks.

And then finally, all the clusterConfig can be put into a separate method to clean up the vector format method.

https://github.com/GEOLYTIX/xyz/pull/981/commits/91e5734105fce6700d01510cb894a31d2045b201

simon-leech:
Thank you @dbauszus-glx ! 
I just made one change which is to check if cluster.resolution is numeric before performing `parseFloat()` on it - and warn if not. (I know its a bit silly for it to be non-numeric but always useful!)

simon-leech:
I've tried to update the config docs to explain these changes as simply as possible.
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#vector-wkt-geojson-cluster

PR #982, opened 2023-10-24
Merge dataset object into gazetteer defaults

dbauszus-glx:
The dataset object should be spread into an object with the gazetteer defaults possible accessed within the search method.

The dataset object received as argument by the search method will be passed to a callback.

The layer should be defined as a const.

It should be possible to define a custom `query` template.

The label param wasn't provided for the default gaz_query.


cityremade (approved):
now I can pass arguments to the gazetteer callback (entry, dataset) where dataset contains dataset properties.

PR #987, opened 2023-10-30
Use ID if field is undefined in theme.

dbauszus-glx:
Use the feature ID from the properties if no distinct field is defined for the distributed theme.

### DIFF
--- a/lib/layer/themes/distributed.mjs
+++ b/lib/layer/themes/distributed.mjs
@@ -1,31 +1,29 @@
 export default function(theme, feature) {
 
   if (!theme.lookup) {
-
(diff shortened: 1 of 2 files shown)


cityremade:
It works correct when .field is undefined however in the reported case it's the value of the defined field which is null.
If we have nulls in the column used by the theme all features get the first colour. Should null be a value in the theme that would get its own colour?

cityremade (changes requested):
Needed support for theme field defined but of null value.

cityremade (approved):
so as I said I am approving this as correct. The rest of the issue is on the data science side.

PR #988, opened 2023-10-30
sqlfilter / json url parameter encoding

dbauszus-glx:
JSON does not want to be encoded as stringified URI. A JSON value may contain an ampersand (&) character. The JSON value should be stringified and URLComponent encoded.

The api module script should not be required to decode URI parameter. This should happen by default.

Hence the parameter value is already assumed to be encoded by the time the value is added to [filter] SQLParams array. The skip option for the wildcard is therefore unnecessary.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -111,13 +111,6 @@ module.exports = async (req, res) => {
 
   req.params.template = req.params._template || req.params.template
 
-  // Decode string params.
-  Object.entries(req.params)
-    .filter(entry => typeof entry[1] === 'string')
-    .forEach(entry => {
-      req.params[entry[0]] = decodeURIComponent(entry[1])
-    })
-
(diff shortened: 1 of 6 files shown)


simon-leech (changes requested):
@dbauszus-glx Thanks for this - been doing a little bit of testing. 
This is working for `filter.current` which contains a % or &. 
This is working for the legend filter which contains a % or &.

However, the filter module is not working correctly. It throws an error in the console 
image

simon-leech:
@dbauszus-glx Just pushed a fix to this i think - but needs testing by others.

simon-leech (approved):
This is all working for me now !

RobAndrewHurst (approved):
Looks all good to me!

simon-leech (approved):

PR #989, opened 2023-10-31
Allow a Query to Set the Edit Options

simon-leech:
This PR is a simple change, in PR #930 we added the ability to use an empty `edit.options` array that would then run the `distinct_values` template on the database. 
This PR simply allows the configuration developer to optionally provide an `entry.edit.query` that will instead be used to populate the `entry.edit.options`. 
This is required as is it possible the options come from a different column / different table entirely, especially if a table is initially empty no options will be provided.


dbauszus-glx (approved):
lgtm

PR #992, opened 2023-11-06
no_result configuration for gazetteer.

dbauszus-glx:
Setting `no_result` in the gazetteer config will display the value as no result message instead of the default from the dictionary.

Setting `no_result` as null will display no message.

### DIFF
--- a/lib/utils/gazetteer.mjs
+++ b/lib/utils/gazetteer.mjs
@@ -15,6 +15,7 @@ export function datasets(term, gazetteer) {
         label: gazetteer.label,
         title: gazetteer.title,
         limit: gazetteer.limit,
+        no_result: gazetteer.no_result,
         leading_wildcard: gazetteer.leading_wildcard,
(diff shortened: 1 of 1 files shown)


cityremade (approved):
It does the job for me. I can set this flag to null on each dataset to prevent 'no result' messages.

simon-leech (dismissed):
@dbauszus-glx Please can you add this to the Config Docs? 
 https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#gazetteer

RobAndrewHurst:
Adding in the `no_result` property in the root of the gazetteer doesn't remove the message

image

image

dbauszus-glx:
@RobAndrewHurst should be fixed in last commit.

RobAndrewHurst (approved):
Very nice easy change!

Works for both an entire gazetteer, or a specific dataset.

PR #993, opened 2023-11-07
Sync fields; Geometry edit updates; Boolean Fields

dbauszus-glx:
The `location.syncFields(fields)` will get the values and update entries for the fields argument.

The `location.update()` and the geometry entry update methods will call the syncFields metthod for their `dependents[]` fields.

The geometry entry update method will rebuild the remove and rebuild the location `viewEntries`.

The numeric entry will not format the entry.value but only format the render if the entry is not editable.

The boolean, numeric, and text entries will assign the newValue if rebuild with editing enabled.

The `.link-with-img` css nesting has been fixed for the lineheight.

The geometry entry method will shortcircuit id not editable with the `hideNoEdit` flag set in the entry config.

```js
{
 "type": "geometry",
 "field": "geom_p_4326",
 "fieldfx": "ST_AsGeoJSON(geom_p_4326)",
 "edit": {
  "modifyBtnOnly": true
 },
 "hideNoEdit": true,
 "dependents": [
  "pin"
 ]
}
```

The boolean entry will not return a disabled checkbox but an icon inline with the label if not editable.


simon-leech (changes requested):
Testing this one out - 
✅ hideNoEdit works well and is a great user improvement for points where geometry checkbox doesn't do anything
✅dependents are now being updated for geometry editing
✅ dependents are updated for non-geometry editing
✅ boolean fields now show a tick or cross when true or false, and edit still works. 

❔ Should null values from the db in integer and numeric types be parsed to a number - at the moment they show as NaN which isn't very good for the user?

dbauszus-glx:
Added to config wiki.
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#geometry

`null` or `undefined` numeric `entry.value` should not be displayed. i added those to the NaN check in commit.

simon-leech (approved):
Thanks @dbauszus-glx working for me - happy with this one!

PR #994, opened 2023-11-09
Cancel drawing interaction with escape key.

dbauszus-glx:
The `Esc/Escape` key will cancel the drawing interaction.

Right click to remove the last vertex will executed conditions [array] methods. This allows to undo the last section from routes in the measure_distance plugin.


RobAndrewHurst:
image

After finishing the interaction how do we get rid of the drawn geometry?

RobAndrewHurst (approved):

PR #995, opened 2023-11-09
layer.view() method must return layer

dbauszus-glx:
This will allow to compose the method with other methods before or after in a plugin.

Also the listview must not import the view method but use mapp.ui.layer.view.


simon-leech (approved):
All working for me!

Issue #966, opened 2023-10-17
type:datetime editing toIsoString transforms time to UTC


dbauszus-glx:
The datetime editing input requires an IsoString input for the current value. This causes a shift from local time to UTC.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

dbauszus-glx:
Resolved in PR.

PR #997, opened 2023-11-10
Locations/entries/date - now takes user local time;

cityremade:
Resolved problem of displayed value, unix epochs and formatting for the input of type date.

Works well for me - I see my local time and the unix epoch saved in the db is correct.
Let me know if it works for you as you are both in different timezones.

### DIFF
--- a/lib/ui/locations/entries/date.mjs
+++ b/lib/ui/locations/entries/date.mjs
@@ -1,47 +1,40 @@
 export default entry => {
 
-  let val;
+  const dateString = new Date(entry.newValue || entry.value * 1000)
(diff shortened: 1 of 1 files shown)


cityremade:
https://github.com/GEOLYTIX/xyz/issues/966

dbauszus-glx (approved):
Excellent. Works well.

Things I changed in review is const use instead of let.

Naming from d to dateString and t to timeString.

I removed the commented code.

Value should be null if non editable and null.

Value must be taken from newValue if available in edit mode. The view may rebuild from the update of dependents with edits not yet saved.

PR #998, opened 2023-11-10
Get nonlanguage [view] templates

dbauszus-glx:
View templates (e.g. reports) may not be language templates and should be got from the getTemplates module.

The view endpoint should return 400 instead of crashing the process if template is undefined.


simon-leech (approved):
This fixes the described bug for me - custom views are no longer crashing

PR #999, opened 2023-11-13
Layer should be merged into workspace values

simon-leech:
This PR simply reverses the merging order of the getLayer.js, 
The object (e.g layer) should be merged into the template not the other way around. 
This is so we can overwrite values defined in a layer template via the workspace
``` json
"template": {
"name": "NEW NAME"
}
```
This would overwrite the `name` value defined on the `template`, allowing for flexibility from within teh workspace.

### DIFF
--- a/mod/workspace/getLayer.js
+++ b/mod/workspace/getLayer.js
@@ -33,21 +33,24 @@ module.exports = async (params) => {
     return new Error('Unable to validate layer param.')
   }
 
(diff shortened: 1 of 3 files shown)


dbauszus-glx (changes requested):
I think the template and locale merge was the wrong way as well. Can you test the last commit.

RobAndrewHurst (approved):

PR #1001, opened 2023-11-14
Assign layer as request param for queries

dbauszus-glx:
Queries which have the layer param need to access the layer object which may be constructed from templates. The query module gets the layer from the getLayer mod. The layer object will replace the req.params.layer string and be accessible in query methods.

**A query may access templates as referenced by fields. The templates must be available as string and not be loaded from a src.**


simon-leech (approved):
This fixes the described bug for me now :)

PR #1002, opened 2023-11-14
get and merge templates in a for of loop

dbauszus-glx:
The async getTemplate and subsequent merge must happen in a for...of loop in the getLayer method. Otherwise the merge would happen after the layer is already returned from method.


RobAndrewHurst (approved):
Looks all good to me!

PR #1003, opened 2023-11-14
layer param and geom in geojson query

dbauszus-glx:
The layer param and geom were not referenced correct.


RobAndrewHurst (approved):
Fixes the issue!

RobAndrewHurst (approved):
Fixes the issue!

PR #1004, opened 2023-11-14
Use origin request param instead of process.env.ALIAS

dbauszus-glx:
The ALIAS environment variable was introduced to alias redirected request.

For example the host host `xyz-blog-dev.vercel.app` maybe routed through `geolytix.dev`. In this case the origin will be unknown can be parsed from the request.header.referer like so:

```js
  const host = `${req.headers.origin 
    || req.headers.referer && new URL(req.headers.referer).origin 
    || process.env.ALIAS 
    || req.headers.host}${process.env.DIR}`
```

The user admin view script should set the login URL parameter flag if no user is provided.

The user mod routes which require a user should return the login view.

The ALIAS is required for cosmetics on the admin email.

image

The verify link does not have a referer since the link is opened directly in the browser, not through a view.

Protocol shouldn't be required though since can follow a host in link.

Without the alias the admin will get the vercel.app domain. Which should still work.


RobAndrewHurst:
When approving a user on the admin panel, I am not seeing an email being sent to the user confirming that they are approved.

PR #1006, opened 2023-11-15
Layer Restrictions

simon-leech:
This PR changes a few things: 
1. Layer panels are no longer completely disabled when the table is restricted, instead only the layer toggle is disabled. 
2. A magnifying glass is added to the layer panel that will zoom to the extent of a non-restricted layer, or zoom the mapview in to the extent required for a zoom restricted layer. 
image

This will hope aid user experience as the disabled toggle signifies the layer cannot be turned on currently, and the magnifying zooms the user in to the right level - or they can just manually zoom in.

### DIFF
--- a/lib/ui/layers/view.mjs
+++ b/lib/ui/layers/view.mjs
@@ -113,9 +113,10 @@ export default (layer) => {
   }
 
   // Create a div for the magnifying glass icon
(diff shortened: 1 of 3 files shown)


dbauszus-glx (changes requested):
I like this. Very smart solution.

image

Can you change the zoomDiv? This should be a button element since it is button. We try to use semantic html throughout. Also the inline style should not be required. Especially the colour is problematic for UI themes.

Working good though.

dbauszus-glx (approved):
This looks good now. I only changed the closure of the button element.
image

PR #1007, opened 2023-11-16
renamed color var, added new color var, changed bg color of drawer

eo-uk:
Changed color palette, renamed the color-lighter variable to color-light-secondary and added another variation. Adjusted drawer background for a subtler background.


dbauszus-glx (approved):
Looks really good.

image

simon-leech (changes requested):
I much prefer the colour scheme but i've got a few thoughts - happy to be overruled though! 

1. The layer key is not apparent anymore.

image
image

2. On the tabview, there's no longer a distinction between the tabview tabs and the information, which results in a vertical line that just ends and isn't super clear why. 
image

image
3. I don't think there's enough distinction between layer groups and individual layers. What about if layer groups have a bold header and layers were not bolded? 
image
image

RobAndrewHurst (approved):
Looks good!

simon-leech (approved):
Thanks for addressing those bits I requested - this is looking brilliant now! much cleaner! 

I think this is fine to be merged once the tests all pass :)

PR #1000, opened 2023-11-13
Confirm changes on close and edit toggle

dbauszus-glx:
The edit toggle and close location button will now ask whether changes should be saved.

image

PR #1008, opened 2023-11-16
Fix toggling edit mode with no changes

simon-leech:
This PR addresses an issue that came about as a result of #1000 . 
The issue was that when you toggle the edit mode button but there is no changes, edit mode is not removed.


eo-uk (approved):

dbauszus-glx (approved):
That works well. Thanks for fixing.

PR #1009, opened 2023-11-16
mvt clear source and refresh

dbauszus-glx:
This is an ugly one. Only way I found to resolve this issue is by waiting a second after the insert statement before reloading the layer. Otherwise the insert geometry is not available and broken tiles are cached.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -60,6 +60,8 @@ export default async layer => {
         layer.draw?.defaults || {}))
     })
 
+    await sleep(1000)
+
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
The created mvt, seemed to only display at certain zoom levels..

https://github.com/GEOLYTIX/xyz/assets/43169158/459c4beb-43b8-4cab-85c8-24e6ac1f7a6b

dbauszus-glx:
I am now checking for features loaded in mvt tiles and reload the layer if the feature has not been loaded.

RobAndrewHurst (approved):

PR #1011, opened 2023-11-20
Zoom Toggle Fix

simon-leech:
Steps to reproduce the bug prior to this PR: 

1. Click zoom to button on a zoom restricted layer
2. Click zoom to button again (zooms you out and turns layer 'off')
3. Zoom in using box zoom
4. Layer appears but toggle isn't orange

This was due to missing setting the layer to display false when the zoom to button was pressed twice


dbauszus-glx (approved):
works for me.

PR #1013, opened 2023-11-20
remove mvt_cache reference; create mvt query

dbauszus-glx:
MVT tiles are now generated by the mvt query template.

mvt_cache has been disabled.

The mvt query as well as the vector query will take an array of fields. This allows for custom theme queries without the need to update the layer in the workspace.

`style.cache:null` will prevent the style from the feature to be returned from the style method.

The `value_only` template param returns the first value only from a response.


RobAndrewHurst (approved):
Very nice!

The performance is excellent!

PR #1014, opened 2023-11-21
strcturedClone template in getLayer

dbauszus-glx:
The template object must be structuredCloned() prior to merging in the getLayer mod.

The objcet.hasOwn() check should continue the for loop for ease of understanding.


dbauszus-glx:
The geom check should not happen whenever a layer is loaded. This should be done in the client format method.

```js
// Check for layer geom[s].
if ((layer.table || layer.tables) && (!layer.geom && !layer.geoms)) {
  console.warn(`Layer: ${layer.key},has a table or tables defined, but no geom or geoms.`)
}
```

simon-leech (changes requested):
This fixes the described error with templates that were unrelated being merged together. 

I'm still seeing issues with templates array merging. Sample config here as an example 
``` json 
 "test": {
          "name": "Test",
          "group": "Competitors",
          "template": "test",
          "display": true,
          "templates": [
            "streetview_entry",
            "style_theme",
            "style_dataview"
          ]
}
```

In this example, streetview_entry is simply an `infoj` entry and `style_dataview` just changes the layer dataview title. Neither of these are being shown but no errors or warnings.

dbauszus-glx:
An already merged template will be returned without having to be merged again.

Please note the reversed order between layer > template but templates[] > layer

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#template-2

simon-leech (approved):
I think this all working for me now - thanks @dbauszus-glx 
Config Docs make sense to me too, so i'm happy with this!

PR #1015, opened 2023-11-22
Location queries

dbauszus-glx:
The mod/location endpoints have been removed in favour of queries.

This has the benefit that all the layer checks, role checks, template retrieval, parameter substitution is managed in one module rather than in multiple places.

This is a requirement to have different location stores in the future. A location could be stored and updated in an Object store by using fetch modules.

The location update query was insecure by passing values from the body directly into the SQL string. All values should be substituted in the database instead.


RobAndrewHurst (approved):
Looks all good to me!

PR #1016, opened 2023-11-23
Check query template roles

dbauszus-glx:
The query needed a little cleanup and commenting.

The query module will now check whether a query template has a roles object and whether the roles are matched by the user roles.

```js
"foo": {
  "template": "select 'foo';",
  "roles": {
    "Bar": true
  }
}
```


eo-uk:
Restricts query to role as intended, will approve after the type confusion check is resolved.

RobAndrewHurst (approved):

PR #1022, opened 2023-12-01
Add new template for account approved no admin

simon-leech:
This PR just adds a template for the account approved no admin. Previously this was defined as text - instead this now uses a template. This means that a custom template can now overwrite this.


dbauszus-glx (approved):
lgtm

PR #1023, opened 2023-12-01
Add error for 4326 mvt layer

simon-leech:
Adding a check on srid and returning an error if its 4326 as no longer supported.


dbauszus-glx (approved):
I changed this to warning since there is no actual error exception. I also moved the default assignment and check on srid not '3857' rather than is '4326'. Other projections would also fail.

PR #1027, opened 2023-12-04
Provide layer key as snap option

dbauszus-glx:
Instead of `snap:true` it should be possible to provide a layer key.

```js
"polygon": {
  "snap": {
    "layer": "iuc"
  }
}
```
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#snap

### DIFF
--- a/lib/mapview/interactions/snap.mjs
+++ b/lib/mapview/interactions/snap.mjs
@@ -25,6 +25,12 @@ export default function(mapview) {
     ...mapview.interaction.snap
   }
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
@dbauszus-glx I noticed if you provide a bogus layer key you get no warning or error, and the layer itself is defined as the snap layer. I think this should be explicit with a warning, so I pushed a commit to do this - let me know what you think. 

The functionality is all working for me though!

RobAndrewHurst (approved):

PR #1029, opened 2023-12-05
wkt_properties feature format and mvt queries

dbauszus-glx:
It should be possible to request feature properties for the viewport separate from the tile (geometries).

The `wkt_properties()` featureFormat method processes wkt features without geometries. The feature properties are assigned with the feature [id] key to the `layer.featuresObject{}`.

The openlayers `layer.L.changed()` is used to re-style the layer after the feature properties are loaded. The removes the need to hook into the tile load and rendered methods.

Only the feature properties need to be loaded if the current layer filter are changed.

Having filter statistics will allow to create dynamic themes and legends which can use viewport statistics of loaded features.

The `mvt_geom` query will ignore the any filters. This will allow to create tile caches for shared geometries.

The `wkt_properties:true` flag will enable this mode for mvt layer.

The [tile] `cachSize` can be configured as an integer on the layer.

### DIFF
--- a/lib/layer/Style.mjs
+++ b/lib/layer/Style.mjs
@@ -18,7 +18,18 @@ export default layer => {
       if (Styles) return Styles
     }
 
-    feature.properties = feature.getProperties()
+    if (layer.featuresObject) {
+
+      let id = feature.get('id').toString()
+
+      feature.properties = layer.featuresObject[id]
+
+      if (!feature.properties) return;
+
+    } else {
+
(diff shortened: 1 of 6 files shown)


simon-leech (changes requested):
image

When using a `format: mvt` layer with `setLabels` on themes, I am ending up with multiple labels drawn to the mapview.

RobAndrewHurst (approved):

PR #1030, opened 2023-12-06
syncFields check

simon-leech:
This PR adds a single line of code to address a small bug. 
If syncFields is called for dependents that do not yet have a value - the response will be null. 
If the response is null, then 
``` js 
this.infoj
    .filter(entry => typeof response[entry.field] !== 'undefined')
``` 
will error as the response contains nothing. 
As such, I simply added a check on if response is null, just return before the filtering to remove this error.

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -97,9 +97,9 @@ async function syncFields(fields) {
       id: this.id,
       fields: fields.join(),
     }));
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):

dbauszus-glx (approved):
i added a check whether response is error.

PR #1032, opened 2023-12-07
Make Admin Panel accessible on mobile

simon-leech:
This PR allows the admin panel to be accessed on mobile - removing the `mobile-display-none` class from the button. 

The admin table has been amended to make use of ` responsiveLayout:"hide"` in Tabulator to hide columns on mobile, so admins can see the important information. 

Hiding order as follows: 

- Email (never hide)
- Verified (never hide)
- Approved (never hide)
- Admin (never hide)
- API (hide first)
- Failed attempts (hide second)
- Language (hide second)
- Roles (never hide)
- Approved by (hide first)
- Expires On (hide first)
- Blocked (hide second)
- Delete (hide second) 

We will want to test this on a deployed instance on a range of devices and screen size emulators - and we might want to tweak which columns we think are most important.

### DIFF
--- a/public/views/_default.js
+++ b/public/views/_default.js
@@ -417,7 +417,6 @@ window.onload = async () => {
     btnColumn.appendChild(mapp.utils.html.node`
       <a
         title=${mapp.dictionary.toolbar_admin}
-        class="mobile-display-none"
         href="${mapview.host + '/api/user/admin'}">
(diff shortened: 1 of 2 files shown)


dbauszus-glx (approved):
tested with mobile emulator in dev tools and working as expected.

PR #1034, opened 2023-12-11
Assign entry.field from query response in infoj

dbauszus-glx:


simon-leech (changes requested):
@dbauszus-glx Thanks for this - a simpler implementation than #1031 . 
However, this implementation isn't quite right. 
The issue with the implementation below is that this assumes all `infoj` entries with `query` also contain `queryCheck:true`, but this is not the case for the `infoj` entry containing just a `query` and a `field`, as there's no benefit to having `queryCheck:true` as this is used to disabled checkboxes if not data is provided.

``` js 
  if (entry.query) {

      // Assign queryparams from layer, and locale.
      entry.queryparams = Object.assign(
        entry.queryparams || {},
        entry.location.layer.queryparams || {},
        entry.location.layer.mapview.locale.queryparams || {})

      // Check whether query returns data.
      if (entry.queryCheck || entry.run === true) {

        // Stringify paramString from object.
        const paramString = mapp.utils.paramString(mapp.utils.queryParams(entry))

        // Run the entry query.
        mapp.utils
          .xhr(`${entry.host || entry.location.layer.mapview.host}/api/query?${paramString}`)
          .then(response => {

            // Assign query response as entry value.
            entry.value = entry.field? response[entry.field] :response;

            // Check whether entry should be skipped.
            if (skipEntry(entry)) {

              // Remove the entry.node from location view.
              entry.node.remove();
              return;
            }

            // Create element to be appended into empty entry.node
            const el = mapp.ui.locations.entries[entry.type]?.(entry)

            el && entry.node.append(el)
          })

        continue;
      }

    }

```

dbauszus-glx:
@simon-leech That is correct. Only queries with the `run` or `queryChecks` flag will be executed in the infoj iteration.

simon-leech (approved):
@dbauszus-glx
I just did a commit to update the console warning to explain how to resolve this warning too.
 All working for me happy to merge!

PR #1035, opened 2023-12-12
Update location_get query

dbauszus-glx:
The fields param and entry.key check in the `location_get` query template were legacy and not actually used.

The `location_get` query should check for templates matching the field name.

The location `get()` method should return if the location_get query fails.


RobAndrewHurst (approved):
Looks all good to me!

PR #1037, opened 2023-12-14
entry type:tab chkbox

dbauszus-glx:
The entry `type:tab` should return the chkbox element not a document fragment containing the chkbox. The chkbox element util returns a node.


RobAndrewHurst (approved):
Nice! Fixes the event issue

PR #1039, opened 2023-12-18
Basic authorization header

dbauszus-glx:
It should be possible to provide a base64 encoded authentication header `--user` with a request.

eg.

```console
curl http://localhost:3000/api/user/key --user "dennis.bauszus@geolytix.co.uk:mypassword"
```

The `auth.js` module will short circuit and await the `fromACL.js` response which is either a user object or error.

The `fromACL.js` module is taken from the post method from the login module which now requires the `fromACL` module.

The ACL module itself has been formatted to return the ACL query method.

The ACL module can now be required without having to be executed.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -123,6 +123,11 @@ module.exports = async (req, res) => {
   // The authentication method returns an error.
   if (user && user instanceof Error) {
 
+    if (req.headers.authorization) {
+
+      return res.status(401).send(user.message)
+    }
+
     // Remove cookie.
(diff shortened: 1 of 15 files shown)


dbauszus-glx:
The remote_address check has been updated 
to prevent `Unexpected constant truthiness on the left-hand side of a `||` expression`.

```js
  const remote_address = /^[A-Za-z0-9.,_-\s]*$/.test(req.headers['x-forwarded-for'])
   ? req.headers['x-forwarded-for'] : undefined;
```

RobAndrewHurst (approved):
Nice one!

I tested this out using this small script

```js
import axios from 'axios';

const url = 'http://localhost:3000/latest/api/user/key';
const username = 'myUser';
const password = 'myPassword';

const auth = {
  username: username,
  password: password
};

axios.get(url, { auth })
  .then(response => {
    console.log('Response:', response.data);
  })
  .catch(error => {
    console.error('Error:', error.message);
  });
```

Works well 👍

PR #1040, opened 2023-12-19
Merge locale into template

dbauszus-glx:
The template being merged into the locale means it is impossible to override anything in the template with locale specific config.

It should be possible to null an object entry in the workspace.locale.

A layer which is null should not be returned in the locale layer array.

### DIFF
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -104,6 +104,11 @@ async function locale(req, res) {
   
   // Check layer access.
   locale.layers = locale.layers && Object.entries(locale.layers)
+
+    // filter layers which are null
+    .filter(layer => layer[1] !== null)
+
+    // check layer for user roles
     .filter(layer => !!Roles.check(layer[1], roles))
     .map(layer => layer[0])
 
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
I have tested this using various config options: 
**On single locale instance without default locale**
1. Overwriting name, roles, table/tables via the workspace.locale to a template from a file. This correctly updated the name and overwrite the roles object of the template.
``` json 
"template": { 
"name": "TEST",
"roles":{ 
"test": null 
} 
} 
```

2. Merging templates array via the workspace.locale to a template from a file. 
``` json 
  "template": {
          "templates": [
            "template_1",
            "template_2"
          ]
        }
```
This correctly merges the json from template_1 and template_2 onto the template. 

**On multi locale instance with default locale**
1. Overwriting name, roles, table/tables via the workspace.locales locale to the default locale. 
``` json 
   "layers": {
        "template": {
          "name": "TEST TEST"
        }
      }
```
2. Merging templates array via the workspace.locales to a default locale to a template from a file. 
``` json 
 "layers": {
        "template": {
          "name": "TEST TEST",
          "templates": [
            "template_1"
          ]
        }
      }
```
This correctly merges the json from template_1 and overwrites the name of the template in the default locale for that particular locale. 

3. Hiding a layer using null in a specific locale from a default locale layers object. 
``` json 
"layers": { 
"template": null
}
```

PR #1041, opened 2023-12-19
nodemailer config from TRANSPORT environment variable

dbauszus-glx:
The transport variable must be split for the user, pass, host config for the nodemailer.

```
"TRANSPORT": "smtps:user:pass@host"
```

### DIFF
--- a/mod/utils/mailer.js
+++ b/mod/utils/mailer.js
@@ -10,28 +10,38 @@ let transport
 
 module.exports = async params => {
 
-  if (!process.env.TRANSPORT && !process.env.TRANSPORT_HOST) {
-    console.warn('No transport method set for mail.')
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
Shouldn't we not maybe just rule out the 'TRANSPORT' key?

simon-leech:
I agree with @RobAndrewHurst , it makes the most sense to me to just have one option enabled which is the 3 keys required. Could we just add a warning if an instance contains the TRANSPORT key?

dbauszus-glx:
@RobAndrewHurst @simon-leech Either will work for me.

simon-leech (approved):
This looks good to me - happier for it to be 3 keys and warnings on incorrect / missing!

RobAndrewHurst (approved):

PR #1042, opened 2023-12-19
Abort dataset gazetteer XHR

dbauszus-glx:
The abort was previously inside the search method which has no reference to the dataset on subsequent iterations.


simon-leech (approved):
This is working well for me and I'm not seeing duplicated results. 
I noticed while testing with searching for coordinates that there was a spelling mistake in `lib\ui\Gazetteer.mjs` so I corrected that and pushed it.

PR #1043, opened 2023-12-20
simple-statistics; featureFormats & featureFields

dbauszus-glx:
The [simple-statistics](https://github.com/simple-statistics/simple-statistics) library has been added as a core dependency. This is loaded as `mapp.utils.stats`

The `featureFormats` module has been moved from `mapp.utils.featureFormats` to `mapp.layer.featureFormats` to be in line with other mapp.layer modules.

The featureFormat methods process layer features for styling.

`featureFields` modules has been added to mapp.layer.

`featureFields.reset(layer)` will reset the `layer.featureFields`

Field values are keys within the `layer.featureFields{}` object.

`featureFields.process(layer)` will process the fields required for a theme and re-render the legend.

`featureFields.distribution{}` methods can be expanded to support different distributions.

`featureFields.distribution.jenks` will process the featureField values array for the theme field to calculate the jenks based on the length of the graduated theme cat_array. The cat values will be updated to represented the natural breaks before the layer and legend are rendered.

The categorized theme legend has been expanded to respect featureField count for the categorical theme field if the distribution is set to `count`.

The mvt query has been updated to support a query without fields.


RobAndrewHurst:
image

I noticed a behaviour that if you zoom in to a certain extent you get more categories than what might be rendered. As you can see in the screen shot there are 8 Categories, but only 3 have values.

Is this desired or do we want the number of categories to limit itself based on what should be rendered

dbauszus-glx:
> image
> 
> I noticed a behaviour that if you zoom in to a certain extent you get more categories than what might be rendered. As you can see in the screen shot there are 8 Categories, but only 3 have values.
> 
> Is this desired or do we want the number of categories to limit itself based on what should be rendered

The properties are loaded on a bounding box not exact extent for performance reasons. there will always be a few more features on the outside edges.

dbauszus-glx:
Graduated cat objects with an undefined value should not be rendered in the legend.

RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/assets/43169158/9f70bfd8-51d2-4e81-a5c4-46a0556349c1

The theming of the vector doesn't seem to correlate with what's rendered on the map and what's rendered as the key past a certain zoom level. Not too sure if the zoom level does have anything to do with it.

RobAndrewHurst (changes requested):
Uploading Screen Recording 2024-01-15 at 11.49.39.mov…


The theming of the vector doesn't seem to correlate with what's rendered on the map and what's rendered as the key past a certain zoom level. 

Can we ensure that the theming of the vector correlates to the key?

dbauszus-glx:
@RobAndrewHurst This should be fixed now. The graduated value check should be on smaller or equal than. Otherwise the largest value which is equal to the largest jenks bin will never be styled.

RobAndrewHurst (approved):

RobAndrewHurst:
> @RobAndrewHurst This should be fixed now. The graduated value check should be on smaller or equal than. Otherwise the largest value which is equal to the largest jenks bin will never be styled.

Nice one! Thanks @dbauszus-glx 

@simon-leech @AlexanderGeere Do you want to take a look?

dbauszus-glx:
> @dbauszus-glx We noticed that if you zoom in on a dot, even though you can only see one, there's still two categories on the side Screenshot 2024-01-16 133351
> 
> Is this desired?

No it shouldn't let me add a check whether a break value is the same as previous.

AlexanderGeere (approved):

simon-leech (changes requested):
I've added the jenks distribution to Config Docs here -
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#distributionjenks
The distribution is working well for me ✅

❓Should we allow the configuration developer to pass formatters to the thematic? I think it would be beneficial to allow the removal of the many decimal places and add thousand separators. 

I've added the count distribution to Config Docs here- https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#distributioncount
❌ I can't seem to get the distribution count working - I keep hitting an error in the categorized module about undefined styles.
image

simon-leech (approved):

simon-leech:
@dbauszus-glx  Thanks for this - this is all working for me now and i've updated the Config Docs to explain the configuration required! Love it! 🎉🎉

PR #1044, opened 2023-12-20
mvt and wkt query break with no fields being requested / vector format SRID warning

dbauszus-glx:
Requesting a wkt or mvt layer with no theme or label is bugged due to the fields array being undefined.

Vector format layer require an SRID configuration.


simon-leech (approved):
Happy to merge - all good with me !

PR #1046, opened 2023-12-21
Don't merge workspace.locale with itself

dbauszus-glx:
Bug happens when a workspace is loaded with only workspace.locale and no locales.


simon-leech (approved):
This stops  the workspace merging into itself for me :)

PR #1052, opened 2024-01-02
fixed popup item hover background color

eo-uk:
Hover background color for popup items was indistinguishable. This changes it to another color.


dbauszus-glx (approved):
works as expected. tested on draw confirmation and cluster select popup li elements.

PR #1055, opened 2024-01-03
Email true queryparams

simon-leech:
Just a little change here - I've added the ability to pass `email:true` in the queryparams object and this is then passed to the query as `%{email}`, the same way as `id: true` works.


dbauszus-glx (approved):
Works as expected.

PR #1056, opened 2024-01-03
Allow object default layer filter

dbauszus-glx:
Previously the layer.filter.default had to be defined as a SQL string.

It is now possible to define the default filter as a filter object.

The SQL filter module will check for the filter type and return the filter argument if string.

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#default


simon-leech (approved):
This is working well for me! 
Tested using it on a layer, hitting switch all on the style only shows the default options applied, as does a filter on that field :)

PR #1060, opened 2024-01-11
Tooltip Tweaks

simon-leech:
Couple of minor tweaks on the tooltip: 
1. If the title is a long title - the tooltip svg expands to the line height which makes it massive, this PR sets the default height to 1.5em and width to 24px (the svg width). 
2. Added tooltip_height key which allows the configuration to pass a height for the tooltip svg if desired. 
3. Removed mobile display-none so the tooltip can be visible on mobile.

This should fix the issue with a long title and a huge SVG, and should also prevent the title spilling to two lines when the text does not require it - as the width is now 24px as per the SVG and not 90px as it previously was.


dbauszus-glx (approved):
We don't want to have inline style in the script if unnecessary.

I added the .tooltip class to the label.

```css
  & .label {
    align-items: center;
    display: flex;
    font-weight: bold;
    height: 1.5em;
  }

  & .label>.tooltip {
    -webkit-user-select: none;
    user-select: none;
    height: 100%;
  }
```

The tooltip icon must not be selected.

The tooltip icon should take the height from the label in which it is nested.

The introudction of tooltip-height as another property is not required.

The height can be set through the css_title. Likewise the gap can be set which is a better descriptor than a margin.

```js
"tooltip": "## ~Foo bar £££",
"css_title": "height: 2em; gap: 1em;",
```

uhtml doesn't require closing tags.

uhtml doesn't require an empty string tag.

This is valid. uhtml will ignore a falsy html fragment provided to the `html.node` tag method.

```js
const tooltipSpan = entry.tooltip && mapp.utils.html`<span>FOO`
```

AlexanderGeere:
Using: 
```Json
{
      "title": "American Golf Score",
      "field": "ag_score",
      "inline": true,
      "group": "Store Information",
      "type": "integer",
      "tooltip": "The model score that aligns with the blueprint score, the higher the number the better the score."
 }
```

Yields: 
Screenshot 2024-01-15 110619

Is this desired behaviour with no styling supplied?

simon-leech:
@dbauszus-glx  This fixes the tooltip height getting massive. However with no styling applied - a configuration of this yields this just like @AlexanderGeere says above. 

``` json 
 {
            "title": "Data Data Data Data Data",
            "field": "field",
            "type": "integer",
            "inline": true,
            "tooltip":"This is a tooltip for additional info, this is super super super long but it should be fine and shouldn't be a problem"
          }

```
image

dbauszus-glx:
The height for the icon must be set implicit to prevent the wrapping of the label. A column gap should be used instead of a gap.

simon-leech:
That's look good to me now - tested and working as expected!

AlexanderGeere (approved):

PR #1062, opened 2024-01-12
Clear highlight interaction on mouseout.

dbauszus-glx:
The mousout event to clear the highlight is added to the highlight interaction and removed when the interaction finishes.


simon-leech (approved):
This is working well for me!

PR #1063, opened 2024-01-12
Failed to load workspace

dbauszus-glx:
A 500 'Failed to load workspace.' error should be returned instead of the zero/OSM only workspace if the workspace cache fails.

The instance should not crash.

The error should be logged to the node console.


RobAndrewHurst (approved):
Nice one!

PR #1070, opened 2024-01-23
Provided missing gazetteer to Coordinates Entry

RobAndrewHurst:


simon-leech (changes requested):
Thanks @RobAndrewHurst  - working as expected for me! 
✅ Tested with bogus lat and long and warning appears in the console.
✅ Tested with correct lat and long and i'm zoomed to that location. 
❓ When you search a bogus lat and long and the warning appears in the console. However, the user isn't given a message to say this. I wonder if we should have an alert too? 
❓ When you search for a correct lat and long that is outside of the locale extent (e.g. searching for a US city but in a UK restricted locale), the zoom to just goes to the further corner of the extent. I think this should also warn and alert the user they can't access that location? 

📃 Please could you check that gazetteer searching by lat and long is in the Config Docs too 🙂

AlexanderGeere (commented):
Sames test steps as simon, Looks like its zooming in on the real points and warning about ones outside the extent or not real. 
The incorrect/bogus co-ordinates get added to the panel.
Screenshot 2024-01-23 145015

RobAndrewHurst:
@dbauszus-glx @simon-leech @AlexanderGeere 

How do we feel about logging the warning of "Latitude and longitude values must be within valid ranges" if the user enters a bogus lat lon?

image

AlexanderGeere:
@RobAndrewHurst I like how that looks, much cleaner than the pop.

AlexanderGeere (approved):

RobAndrewHurst:
> @RobAndrewHurst I just removed the changes to the `fitView.mjs` as the changes in the `gazetteer.mjs` negated the need for them! All looking good!
> 
> Please can you test this out using these coordinates 63.032097003338635, -122.06125081191139 on a UK-restricted locale and you will see what I mean about the zoom to location going to the far corner of the restricted extent. I think in this case we should also show the same warning message?

I am aware of what you mean. I am trying to figure out the best way to solve this

RobAndrewHurst:
### Changes Made ✍️ 
- Added a check to ensure that the provided target extent is contained within the current extent before performing map operations.

> The code introduces a safeguard to verify the validity of the target extent in the optional options. If a target extent is provided, the pull request checks whether it is within the current extent. If not, an alert is triggered, and the function returns false, preventing invalid map operations.

This enhancement aims to improve the reliability of the code by validating geographic boundaries before executing critical map-related operations.

#### Testing
The changes have been tested locally and are functioning as expected with different locales performing different coordinate searches.


Please review the changes, and let me know if any further changes are needed :)

simon-leech (approved):
This is brilliant thank you @RobAndrewHurst ! 🎉🎉

dbauszus-glx (changes requested):
Can you check the gazetteer test workspace with this configuration?

```js
            "gazetteer": {
                "layer": "retailpoints",
                "qterm": "store_name",
                "leading_wildcard": true,
                "limit": 5,
                "no_result": "foo",
                "datasets": [
                    {
                        "title": "Postcode",
                        "qterm": "postcode",
                        "no_result": null
                    }
                ]
            },
```

this currently fails on the xhr check.

simon-leech:
@RobAndrewHurst @dbauszus-glx  I put together a fix for this but I think there's probably a more elegant way of doing this. I'm simply calling the search function with the same parameters that are passed if a `datasets` array is provided.

dbauszus-glx (approved):
This is working for me though I had to make a couple of changes.

```js
  this.Map.getView().fit(extent, {
    padding: [50, 50, 50, 50],
    duration: 1000,
    ...options
  })
```

The [ol.view.fit()](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit) method uses an options param as optional argument. This can be provided in custom script. Custom params not understood by the ol API should not be added to this options object. The mapview.fitView returns on error but should not return true or false if the extent isn't matched. This method is called from other script sources than the gazetteer as well.

Dictionary entries should be added when the Gazetteer is created but not on every search, nor every autocomplete keypress.

lng,lat locations should be checked with [extent.containsCoordinate()](https://openlayers.org/en/latest/apidoc/module-ol_extent.html#.containsCoordinate)

Please note that the mapview is usually 3857, but could be in a different projection.

```js
  const coord = ol.proj.transform(
    [location.lng, location.lat],
    `EPSG:4326`,
    `EPSG:${gazetteer.mapview.srid}`
  )

  if (!ol.extent.containsCoordinate(gazetteer.mapview.extent, coord)) {
    alert(mapp.dictionary.invalid_lat_lon);
    return;
  }
```

RobAndrewHurst:
Nice one @dbauszus-glx!

PR #1072, opened 2024-01-23
round shorthand for formatterParams.options.maximumFractionDigits

dbauszus-glx:
I have a feeling that formatting is often done in the query fieldfx rather than in the infoj entry config because of the formatterParams being rather verbose to match the [toLocaleString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) config.

This PR provides a shorthand `round:int` for the type:numeric entry config.

```js
{
  "title": "Aged <18",
  "type": "numeric",
  "round": 1,
  "field": "age0to18_perc",
  "suffix": "%",
  "inline": true
}
```

### DIFF
--- a/lib/ui/locations/entries/numeric.mjs
+++ b/lib/ui/locations/entries/numeric.mjs
@@ -22,6 +22,9 @@ export default entry => {
   if (entry.type === 'integer') {
 
     entry.formatterParams.options.maximumFractionDigits = 0
+  } else {
+
+    entry.formatterParams.options.maximumFractionDigits ??= entry.round || 2
   }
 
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):
Tested on american golf, works great. Much cleaner as well.
@dbauszus-glx Could you update the docs here https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#numeric-1, to mention this new way of rounding.

RobAndrewHurst (approved):
Looks all good to me :)

PR #1074, opened 2024-01-24
Typeof `function` check

RobAndrewHurst:
### Changes Made

This pull request introduces improvements to our codebase by incorporating the `typeof` check to ensure that certain properties are functions before attempting to execute them in the layer decorate and in the default view's javascript.

1. **Handling Layer and Plugin Methods**
   - In the code responsible for handling layer and plugin methods, we have added `typeof` checks to verify whether the properties are functions before attempting to execute them. This enhancement aims to prevent potential errors when invoking methods that might not be functions.

   ```javascript
   Object.keys(layer).forEach((key) => {
           // Check if mapp.layer[key] is a function before executing it
           typeof mapp.layer[key] === 'function' && mapp.layer[key](layer);

           // Or plugin method and provide the layer object as an argument.
           typeof mapp.plugins[key] === 'function' && mapp.plugins[key](layer);

           // It is possible to have a plugin method of the same name as a layer method.
       }
   });
   ```

### Motivation

The motivation behind these changes is to enhance the robustness of the code by explicitly checking the types of properties before executing them. By ensuring that methods are functions, we reduce the risk of potential runtime errors in the codebase and other custom views.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -247,10 +247,10 @@ export default async layer => {
   Object.keys(layer).forEach((key) => {
 
     // Check for layer method matching the layer key.
-    mapp.layer[key]?.(layer);
+    typeof mapp.layer[key] === 'function' && mapp.layer[key]?.(layer);
 
     // Or plugin method and provide the layer object as argument.
-    mapp.plugins[key]?.(layer);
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
Nice one @RobAndrewHurst - looking good to me! 🎉

dbauszus-glx (approved):
lgtm

PR #1077, opened 2024-01-24
mapp.host

dbauszus-glx:
The mapp library module should set the host if available on the head.

The mapp.host can be manually set but it shouldn't be required to set this in most cases.


simon-leech (approved):
This one is working well for me - tested in default view and also a custom view and helps to remove extra lines of code! Nice one! 🎉🎉

PR #1079, opened 2024-01-24
Load plugins in mapview creation

dbauszus-glx:
With the `loadPlugins: true` option the `mapp.Mapview()` method becomes async and must be awaited.

Plugins from the locale are loaded and executed.


RobAndrewHurst (approved):
Nice one! 

I really like IIAFE :)

Great for the use case of loading plugins from the mapview like that 👍

AlexanderGeere (approved):
> Nice one!
> 
> I really like IIAFE :)
> 
> Great for the use case of loading plugins from the mapview like that 👍

+1  🔥

simon-leech (approved):
Love it! 🎉

PR #1080, opened 2024-01-24
Update mapview.geoJSON method

dbauszus-glx:
The geojson format should be defined outside the function closure to not be re-assigned on every execution.

The geometry should be parsed from value if not provided.

The `Style` should be created from a `style` object if not provided.

`srid` should be fallback for the dataProjection.

### DIFF
--- a/lib/layer/featureFormats.mjs
+++ b/lib/layer/featureFormats.mjs
@@ -1,6 +1,8 @@
-export function geojson(layer, features) {
+const formatGeojson = new ol.format.GeoJSON
 
-  const formatGeojson = new ol.format.GeoJSON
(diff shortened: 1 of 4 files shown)


RobAndrewHurst:
image
 
I am getting this error that the entry on a mapview.geoJSON is not defined

dbauszus-glx:
@RobAndrewHurst can you retest this with the last commit?

RobAndrewHurst:
Still getting the error.

@dbauszus-glx You can replicate this with the vector.json workspace on bugs-testing

RobAndrewHurst (approved):

PR #1083, opened 2024-01-25
provide string array argument as filter for ui.location.infoj(loc, [])

dbauszus-glx:
If the infoj argument is provided as an array of strings a lookup for the entry with matching field will be attempted at the beginning of the iteration.

This does not backwards compatibility where the infoj argument is provided as an array of objects.

image

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -8,8 +8,21 @@ export default (location, infoj) => {
   // Create object to hold view groups.
   const groups = {}
 
+  if (Array.isArray(infoj)) infoj.forEach(entry => {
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
Nice one @dbauszus-glx  - this is a clean solution! 

📃 Please could you write this up in the Docs? then I'll be happy to approve!

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/wiki/MAPP.UI#infojlocation-infoj

simon-leech (approved):

RobAndrewHurst (approved):
Looks all good to me!

PR #1086, opened 2024-01-25
Tabulator rework + warning;

cityremade:
The aim of this PR is to provide the ability to replace the Tabulator import and dataview creation in a plugin.

It should also be possible to extend the `ui.Dataview()` method with custom dataviews.

The `dataview:string` property is used to lookup available dataview methods.

A warning will be issued if the dataview method is unknown.

```js
  if(!Object.hasOwn(mapp.ui.utils, _this.dataview)) {
    console.warn(`mapp.ui.utils.${_this.dataview} doesnt exist`)
    return;
  }
```

All dataviews must have a create() method.

The Dataview.mjs module will warn if the create method doesn't exist.

```js
  if(typeof mapp.ui.utils[_this.dataview].create !== 'function') {
    console.warn(`mapp.ui.utils.${_this.dataview}.create() method doesn't exist`)
    return;
  }
```

Direct calls to the legacy Tabulator method without the dataview string property will work but issue a warning.

```js
function Tabulator(_this) {
  if(!_this.dataview) console.warn('Please change mapp.ui.utils.Tabulator to mapp.ui.utils.tabulator.create');
  return mapp.ui.utils.tabulator.create(_this);
}
```

All Tabulator methods are in on file which can be replaced once a suitable plugin for the Tabulator dataviews is available.

The dataviews.mjs [panel] module must filter dataviews which cannot be created to prevent the panel render failing.

A custom dataview can be added in a plugin like so:

```js
mapp.ui.utils.handsometable = {
    create
}

function create(_this) {

    console.log(_this)
}
```

### DIFF
--- a/lib/ui/utils/_utils.mjs
+++ b/lib/ui/utils/_utils.mjs
@@ -1,10 +1,15 @@
-import Chart from './Chart.mjs'
+import chart from './Chart.mjs'
 
 import tabulator from './tabulator.mjs'
 
 function Tabulator(_this) {
   if(!_this.dataview) console.warn('Please change mapp.ui.utils.Tabulator to mapp.ui.utils.tabulator.create');
   return mapp.ui.utils.tabulator.create(_this);
+}
+
+function Chart(_this) {
+  if(!_this.dataview) console.warn('Please change mapp.ui.utils.Chart to mapp.ui.utils.chartjs.create');
+  return mapp.ui.utils.chartjs.create(_this);
 }
 
 import idleLogout from './idleLogout.mjs'
@@ -15,8 +20,9 @@ import resizeHandler from './resizeHandler.mjs'
 
 export default {
   chartjs: {
-    create: Chart
+    create: chart
   },
+  Chart,
   tabulator,
   Tabulator,
   idleLogout,


dbauszus-glx:
Tabulator must be imported as full from ESM. I have updated the sample plugin script. https://github.com/olifolkerd/tabulator/issues/3455

Events and setData method should be added and looked up from the tabulator object.

simon-leech (approved):
This is all looking good to me! 

I fixed a minor typo where the label for the headerfilters for Less Than and Greater Than were the wrong way round.

lib/ui/utils/Chart.mjs
```diff
@@ -165,4 +165,12 @@ export default async function (_this) {
 
   // Set _this.data if provided.
   _this.data && _this.setData(_this.data)
+}
+
+function say_hello(_this) {
```
RobAndrewHurst, line 170: This needs to be removed.

RobAndrewHurst:
image

The verticalFillMode seems to be failing in tabulator with a location's dataview.

You can replicate this with the scratches location dataview on the tabulator locale on the tabulator/workspace.json

RobAndrewHurst (approved):
Very nice!

Looks all good for me!

RobAndrewHurst:
@AlexanderGeere Do you want to have a look a this PR? :)

cityremade:
Yes works for me as well, I can now have either handson or tabulator within the same locale.

dbauszus-glx (approved):

PR #1088, opened 2024-01-25
Move default btn methods to plugins

dbauszus-glx:
locator, zoomToArea, and fullscreen behave exactly like plugins.

These 'default' plugins have been moved to the plugins module from the default view to clean up the default view.

Methods have been moved out of the <button> tag for better maintenance and linting.

### DIFF
--- a/lib/plugins.mjs
+++ b/lib/plugins.mjs
@@ -19,3 +19,93 @@ export async function svg_templates(plugin) {
 
   await Promise.all(promises)
 }
+
+export function locator(plugin, mapview) {
+
+  const btnColumn = document.getElementById('mapButton');
+
+  // the btnColumn element only exist in the default mapp view.
+  if (!btnColumn) return;
+
+  plugin.btn = mapp.utils.html.node`
+  <button
+    title=${mapp.dictionary.toolbar_current_location}
+    onclick=${(e) => {
+      e.target.classList.toggle('active');
+      mapview.locate();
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):
Tested in bugs and testing. Plugins still, code looks good

simon-leech (changes requested):
This is working for me when using `locator: {}` as per this PR. 
❓However, this fails if I use the currently applied configuration prior to this PR of `locator: true`. As such, should this be referred to as a breaking change as it fails against current configuration? 
❓I also noted that you cannot hide this default plugins, i'm unsure if we should allow that but should this be possible to set `locator:false` for instance to remove it? 
📃 Please can you update the Config Docs to include this? For instance https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#locator this amending.

dbauszus-glx:
@simon-leech The plugin must accept a true flag. I did a commit for this. I agree that the default should be off. We should discuss what's the preferred setting for the default view.

dbauszus-glx:
@simon-leech can you re-review. The requested changes have been made.

RobAndrewHurst:
image

@dbauszus-glx Is there a reason why there is a `plugin` param being passed to the function but isn't used?

dbauszus-glx:
@RobAndrewHurst No particular reason. It's just the default first param for these type of plugin methods. Could be null but doesn't hurt to know what the first param is.

PR #1093, opened 2024-01-26
Editing Geometry Fix

simon-leech:
**What**
This PR addresses a bug whereby editing a location `entry` of type `geometry` would save the changes to the db but not update the geometry shown on the map. 

**Solution**
By adding delete entry.geometry before the update call, this forces the geometry to be re-created and thus re-drawn to the map, fixing the described issue.


dbauszus-glx (approved):
lgtm

PR #1094, opened 2024-01-26
Re-drawing Geometry Fix

simon-leech:
#1093  This is an extension of this PR to also address the situation whereby you have an existing geometry and wish to re-draw it.


dbauszus-glx (approved):
lgtm

PR #1095, opened 2024-01-26
Get array of layer objects instead of layer keys

dbauszus-glx:
Setting `layers=true` request param for the `/api/workspace/locale` endpoint will return an array of layer objects instead of layer keys.

Fewer requests have to be made to the workspace endpoints and the default script is shorter.

The hidden osm layer should not be assigned at this stage but when trying to create a mapview with an empty layers array.


RobAndrewHurst (approved):
Nice one! 

Looks all good to me!

PR #1096, opened 2024-01-26
Check whether a plugin string endsWith some value from an array

dbauszus-glx:
An optional array argument endsWith which defaults to `['.js','.mjs']` can be provided to the loadPlugins method to limit which plugins should be loaded.

```js
// Load the ui_elements googleMaps plugins if defined in the locale.plugins
await mapp.utils.loadPlugins(locale.plugins, ['ui_elements.js','googleMaps.js']);
```

### DIFF
--- a/lib/utils/loadPlugins.mjs
+++ b/lib/utils/loadPlugins.mjs
@@ -1,26 +1,28 @@
-export default plugins => {
+export default (plugins, endsWith = ['.js','.mjs']) => {
 
   if (!Array.isArray(plugins)) return;
-  
-  return new Promise(resolveAll=>{
 
-    const promises = plugins.map(
-      plugin => 
-        new Promise((resolve, reject) => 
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):

AlexanderGeere:
@dbauszus-glx The zoom to location and fullscreen map plugins - do not load by default anymore, is that intended?

4.7.7-alpha:
Screenshot 2024-01-30 102358

This PR:
Screenshot 2024-01-30 102457

They can of course be added into the workspace, but then they are out of the usual order:
Screenshot 2024-01-30 103516

dbauszus-glx:
@AlexanderGeere Yes, that was in response to the change request from @simon-leech 

There should be a way to control the order of buttons but that should be handled in a different PR.

AlexanderGeere (approved):

PR #1097, opened 2024-01-29
Replace custom clone util with structuredClone

dbauszus-glx:
The layer style default assignment and graduated theme catStyle assignment need to be tested with the structuredClone webapi.


RobAndrewHurst (approved):
Nice! Works well!

AlexanderGeere (approved):
Looks good to me

PR #1098, opened 2024-01-29
Mapview attribution

dbauszus-glx:
Prevents the use of an id element e.g. `#Attribution`

Adds attribution to the mapview.map target without the need to specifiy a target element in the constructor.

Adds base css to mapp.css and removes css from default view head.

Removes markup from default view html.

Allows to add a logo to mapview.attribution constructor.

Uses structuredClone intead of assign to to empty object.

Moves attribution from mapview constructor into own module file.


RobAndrewHurst (commented):
Looks good to me!

RobAndrewHurst (approved):

PR #1100, opened 2024-01-30
Review ui.elements.modal

dbauszus-glx:
Modal must not have close button if not defined in constructor.

Modal must not have header if not defined in constructor.

Modal must not have inline styles if not defined in constructor.

Modal should take size from content.

Modal should be movable within the context of the parent.

A target in which the modal will be positioned absolute must be defined.

A header is optional. Can be provided as string or html element.

Close is optional. A close button will be displayed in the top right corner.

Default position is top: 0, left: 0.

Content must be provided to calculate the offset width and height.

```js
mapp.ui.elements.modal({
  target: retailplaceMapview.target,
  header: 'Foo',
  close: true,
  top: 20 // pixel, default 0
  right: 20 // pixel, default left: 0
  content: layerSwitch,
  css_style: 'padding: 10px;',
  contained: true,
  //containedCentre: true
})
```

With the `contained:true` flag the modal bounds can not be shifted outside the target bounds.

With the `containedCentre:true` flag the centre of the modal can not be shifted outside the target bounds.

### DIFF
--- a/lib/ui/elements/modal.mjs
+++ b/lib/ui/elements/modal.mjs
@@ -1,25 +1,26 @@
 export default (params) => {
+
   // Ensure the target is an HTMLElement before proceeding
   if (!(params.target instanceof HTMLElement)) return;
+
+  const closeBtn = params.close && mapp.utils.html.node`
+    <button
+      data-id=close
+      class="mask-icon close"
+      onclick=${closeModal}>`
+
+  const header = params.header?
+    mapp.utils.html.node`<h3>${params.header}${closeBtn}`
+    : closeBtn
 
   // Construct the modal element and append it to the target element
   params.modal = params.target.appendChild(mapp.utils.html.node`
     <div 
-      style="top: 10px; right: 10px; overflow: auto;"
+      style="top: 10px; right: 10px;"
       data-id=${params.data_id || 'modal'}
       class="modal">
-      <div class="header bold">
-        <span>${params.header}</span>
-        <button
-          data-id=close
-          class="mask-icon close"
-          onclick=${closeModal}>
-      </div>
+      ${header}
       ${params.content}`);
-
-  //set the width and height of the modal from where it gets defined.    
-  params.modal.style.width = params.width || '50%';
-  params.modal.style.height = params.height || '50%';
 
   // Function to handle the modal closure
   function closeModal(e) {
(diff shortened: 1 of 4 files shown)


RobAndrewHurst (changes requested):
## Request for change ✍️ 

After popping out a style into a modal and then closing the style, you have to select another style before being allowed to pop out or see the style again.

I think expected behaviour would be that the legend is returned to the layers panel.


https://github.com/GEOLYTIX/xyz/assets/43169158/6f6a7ace-95d8-484f-a25f-70440bb714e8

dbauszus-glx:
@RobAndrewHurst The issue with close method should be fixed now.

RobAndrewHurst:
> @RobAndrewHurst The issue with close method should be fixed now.

Lekker!

RobAndrewHurst (changes requested):
The fix has put back the modal after closing it. 

But if you have multiple modals open from different styles it seems to break.


https://github.com/GEOLYTIX/xyz/assets/43169158/026012b4-5a6f-4170-b172-b5bf641681c5

dbauszus-glx:
@RobAndrewHurst  Changing the theme will now remove the closest modal to the legend.

RobAndrewHurst:
> @RobAndrewHurst Changing the theme will now remove the closest modal to the legend.

Very nice! Thanks @dbauszus-glx working well!

simon-leech (changes requested):
@dbauszus-glx 
📃 Please can you write this all up in the Config Docs before I can review it in detail, as I'm not quite sure of all the options to test!

❓I applied just this to `layer.style` `    "allowModal": true,` . This added the popout modal button but the modal was added with a close button. Is this the expected behaviour and close is by default true?

dbauszus-glx:
@simon-leech The behaviour is still likely to change as we still need to review how legends should work.

simon-leech:
@dbauszus-glx  Noticed a little bug. 
image

If you add a modal to a style - hit pop out for the modal, then close the modal, then collapse the style - the modal button remains on top of the style element rather than being closed with the style.

dbauszus-glx:
The allowModal option has been removed.

simon-leech (approved):
@dbauszus-glx  Thanks for removing `allowModal` - much cleaner! 

Tested this out and working for me  :)

PR #1101, opened 2024-01-30
Signing API / Cloudinary signing

AlexanderGeere:
## What
Creates a new API for signing requests to third party services e.g cloudinary, s3, etc. 
The signing API Initially will only support cloudinary. 

## Why
Signing requests and then letting the client make the call to the service will take some load off the backend as well
further decoupling some of these services from the backend.

### How to test
Use this pull request along with the _sign_ workspace on the _bugs_testing_ branch.


lib/ui/locations/entries/cloudinary.mjs
RobAndrewHurst, line 1: Shouldn't this say 'This Module'? rather than plugin?

AlexanderGeere: Yes that makes more sense.

dbauszus-glx (changes requested):
It's mostly there.

I removed the warning in the cloudinary entry module.

The warning will be added to the _entries module once the cloudinary module can be removed after testing a cloudinary plugin to replace the existing module.

For now the current method should work and the plugin should override the current methods when loaded.

I get a warning in regards to third party cookies. This come from the image src request after an image has been uploaded with a signed URL.

image

These cookies should be removed if no longer needed.

AlexanderGeere:
@dbauszus-glx in to replicating the warning I see that res.cloudinary.com should not set a cookie (see https://github.com/cloudinary/cloudinary_npm/issues/308#issuecomment-590387316), which is the url for the the image request. Being logged in to cloudinary in a separate tab is what causes the warning.

dbauszus-glx (approved):
This is working now for me. I just had to ensure to clear the cache and remove old cloudinary cookies from browser.

PR #1103, opened 2024-01-30
Allow to define the attribution target

dbauszus-glx:
It must be possible to define the target of the attribution.

In the case of the default mapview the attribution must be assigned to the `#Map` parent element of the `#OL` mapview.target.

This is to allow for the attribution to float with the tabview and not over it.


RobAndrewHurst (approved):
Looks good to me!

simon-leech (approved):
Fixes the described issue for me!

PR #1105, opened 2024-02-01
Cut SQL array from location_get request params

dbauszus-glx:
This fixes an issue where the location_get query will err with a default layer filter.


AlexanderGeere (approved):
Tested on an instance, looks good to me

Issue #1113, opened 2024-02-02
Missing process.env.CLOUDINARY_URL


dbauszus-glx:
mod.sign.cloudinary fails without process.env.CLOUDINARY_URL defined.

PR #1115, opened 2024-02-05
Fix cloudinary missing URL error

AlexanderGeere:
### What
This PR addresses [#Issue 1113](https://github.com/GEOLYTIX/xyz/issues/1113).

### Why
If the cloudinary URL was missing the build would fail

### How 
Build the url when the endpoint is actually called


dbauszus-glx (approved):
nice one

PR #1116, opened 2024-02-06
Store user session token in process

dbauszus-glx:
The auth module should store user sessions in the process to reduce the number of requests for the [nano] session token to the ACL.

### DIFF
--- a/mod/user/auth.js
+++ b/mod/user/auth.js
@@ -4,6 +4,8 @@ const acl = require('./acl')
 
 const fromACL = require('./fromACL')
 
+const user_sessions = {}
+
 module.exports = async (req, res) => {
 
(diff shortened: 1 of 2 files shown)


RobAndrewHurst (changes requested):
Noticed some behaviour that when the ACL is missing the `session` column the query will fail as it should. 

But I get a failed login attempt, and then if you try enough times it will lock the account.



image

dbauszus-glx:
I have updated this PR. The user.session token will be deleted if the query to store the session fails due to a missing session column. The login will not fail. There will be an error message in regards to the failed query.

RobAndrewHurst (approved):
Nice one!

PR #1118, opened 2024-02-06
Set newValue to current date/time if value is null

dbauszus-glx:
Editing a date/time entry will set the newValue to the current date if the entry value is falsy.


RobAndrewHurst (approved):
Very nice!

PR #1121, opened 2024-02-08
Layer filter legend render

dbauszus-glx:
The `ni` filter which can be toggled in a categorized legend should be compatible with a different filter type, eg. a `like` filter on the same field.

A `ni` filter card in the filter panel should be linked to the categorized theme legend. Toggling categories in the legend should re-render the filter. Toggling the category in the filter card should re-render the legend.

Filter must cloned from infoj entry into the filter.list. Otherwise a structured clone of the infoj entry will fail after the remove method is assigned to the filter.

Several filter type methods did not make use of the shared applyFilter method.

The `in` filter must return checkboxes in a `.filter` class div.

The category legend will look for the filter matching the type and field. If a filter.card exists the `.filter` class element will be replacedWith the mapp.ui.layers.filters element.


simon-leech:
@dbauszus-glx  I just pushed two commits to fix issues with the basic and distributed thematic legends.

dbauszus-glx:
Last commit fixes two issues. The check for `filter.card` must be conditional whether the filter exists.

Loading images must be conditional whether an image is already loaded.

```js
        // Assign a load event for the image and load the image.
        const image = icon.legendStyle.getImage()

        const image1 = image.getImage(1)

        if (image1.src) {

          onImgLoad()
        } else {

          image1.addEventListener('load', onImgLoad)
          image.load();
        }
```

simon-leech (approved):
I've tested this one and worked through the issues I found with @dbauszus-glx  
What I've tested: 
- Distributed thematic, layer legend is now working after my commit.
- Basic thematic, layer legend is now working after my commit.
- Categorized thematic, you can toggle display of categories in the style legend, or in the filter panel. Toggling in the style panel and filter panel work together and clearing one uses the other (there was an issue where clearing one removed the filter that was resolved by @dbauszus-glx )

PR #1128, opened 2024-02-13
layer.zoomDisplay flag

dbauszus-glx:
The layer.zoomDisplay flag is introduced in this PR to control whether a layer should be displayed when in range.

The layer.show method will set the zoomDisplay true.

The layer.view changeEnd method will hide a layer without a layer.tableCurrent() but set the zoomDisplay true if the display is true. The layer.show() method will be called by the changeEnd method if layer.tableCurrent() and zoomDisplay is true.

layer.tableMax and layer.tableMin are removed since they are unused.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -12,8 +12,6 @@ export default async layer => {
     hideCallbacks: [],
     tableCurrent,
     geomCurrent,
-    tableMax,
-    tableMin,
     zoomToExtent,
   });
 
@@ -261,6 +259,19 @@ export default async layer => {
 
 function show() {
 
+  // Layer should be shown if possible.
+  this.zoomDisplay = true;
+
+  // Push the layer into the layers hook array.
+  this.mapview.hooks && mapp.hooks.push('layers', this.key);
+
+  if (this.tables && !this.tableCurrent()) {
+
+    // Layer is outside of zoomDisplay range.
+    delete this.display;
+    return;
+  }
+
   // Show the layer
   this.display = true;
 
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/assets/43169158/6c5a0476-e0d1-481d-9d1c-343753b31f6d

I am seeing some behaviour where some layer's switch won't toggle even when seeing the layer actually turn off.

simon-leech (changes requested):
✅ I like the visual change that layers remain disabled but showing the user they are set to display when zoom level changes. 
✅ Turning on a layer > Zooming out - sets layer disabled but toggled on > Zooming in - turns that layer back on. This is a much nicer user experience. 
✅ Zoom restricted layers with dataviews - dataviews are now removed when the layer is out of range. 

❓ I'm seeing a CSS issue for layers without zoom restrictions, the dropdown to expand the layer controls is not right-aligned. 
image
❓ Hitting the Zoom to Extent button should also turn on the layer, this currently doesn't happen. When you hit this button the button is then removed which also causes the issue flagged in 1. 
❓ When a layer is zoom restricted we should disabled expansion of the header>mask-icon-expander. This would prevent users from expanding any panels (style and filter but also plugin panels) that can amend the layer data. For instance currently you can expand the filter panel and try to apply a filter which throws an error. 
image

dbauszus-glx:
@RobAndrewHurst The layer toggle should now be fixed for layers without zoom restriction.

The button should not have an on and off class but just an on class to be toggled. Off is assumed to be the default.

```js
      <button
        data-id=display-toggle
        title=${mapp.dictionary.layer_visibility}
        class="${`mask-icon toggle ${(layer.zoomDisplay || layer.display)? 'on': ''}`}"
        onclick=${e => {
          const toggle = e.target.classList.toggle('on')
          toggle? layer.show(): layer.hide()
        }}>`
```

dbauszus-glx:
@simon-leech The zoomBtn will now change visibility instead of display style property to prevent the margin collapsing resulting in the buttons to move left to the title.

dbauszus-glx:
@simon-leech Always showing the zoomBtn is much easier. Otherwise the extent button maybe hanging around.

Extent and zoom to range now both show the layer.

dbauszus-glx:
The changeEnd event method of the layer [view] will now disable and remove the view content.

simon-leech (approved):
This is all working well for me now: 
1. Layer panels are disabled and enabled with zoom in/out
2. CSS issue is fixed 

Really like it! 🥳

RobAndrewHurst (approved):

PR #1130, opened 2024-02-13
Do not logout on failure to retrieve locales or locale

dbauszus-glx:
This causes issues with alternative login method, e.g. SAML.

An error to retrieve the locale e.g. from the server should not result in a log out.


dbauszus-glx:
Show a dialog instead of an alert.

simon-leech (approved):

RobAndrewHurst (approved):

RobAndrewHurst (commented):
image

Are we happy with this?

simon-leech:
@RobAndrewHurst  Looks good to me!

RobAndrewHurst (changes requested):
image

Are we happy with this?

PR #1131, opened 2024-02-13
headerFilter Map Issue

simon-leech:
When the headerFilter returns a single response value, the `.map` was failing.
This PR simply checks if the response is not an array, and converts to one if not - fixing the error.


dbauszus-glx (approved):
works for me.

RobAndrewHurst (approved):

PR #1135, opened 2024-02-14
update user in cookie

dbauszus-glx:
The `/api/user/cookie` endpoint returns the user object.

The endpoint will now query the user from the ACL and update language, roles, and admin flag.

image

The default view ands script will no longer store the user in the head but request the user from the cookie endpoint refreshing the cookie and updating the user object in the process.


AlexanderGeere (approved):
Nice!

RobAndrewHurst (approved):

PR #1139, opened 2024-02-15
Button Panel Ordering

simon-leech:
What?
Ability to re-order the buttons in the button panel (both core buttons and plugin buttons).

Why? 
With the move of zoom to area and fullscreen to plugins, and with the ability to add buttons to the button panel via plugins, the button panel can have a lot of buttons, and there is currently no way to control their order. 

How? 
Introduced a new `locale.button_order` array in which you pass the `id` of buttons.
The order of the buttons in the array will determine the order in which the buttons are added to the button panel. 
Any buttons not in the button_order array will currently not be added.

For example, the below configuration will yield what's shown in the screenshot. 
``` json 
   "button_order": [
        "zoomToArea",
        "btnZoomIn",
        "btnZoomOut",
        "locator",
        "fullscreen",
        "btnAdmin",
        "btnLogin"
      ],
```
image

Why? 
This is advantageous to allow us to control button ordering by on an individual instance level.

Issues
The implementation presently is surely not the best way, hiding the current button column, accessing the ids of button in the current button column, and then overwriting the button column. 
But we need to wait for the button column to be built as lot of plugins append buttons to this, so i'm unsure of a cleaner approach at the moment.

### DIFF
--- a/lib/plugins.mjs
+++ b/lib/plugins.mjs
@@ -29,6 +29,7 @@ export function locator(plugin, mapview) {
 
   const btn = mapp.utils.html.node`
   <button
+    id="locator"
     title=${mapp.dictionary.toolbar_current_location}
     onclick=${(e) => {
       e.target.classList.toggle('active');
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
Please check #1144 in regards to button plugins and the proposed execution order of any plugin.

PR #1140, opened 2024-02-15
vector / mvt layer with tables zoom restriction

dbauszus-glx:
The layer.view should not toggle the layer show / or hide method. This must be controlled from the layer itself to work in custom views where a layer may not have a view element, or a custom view element, or the mapp.ui is not even loaded.

The changeEnd event should only be added for layer with tables and not to every layer and then short circuit if the layer has no tables key.

The layer.view changeEnd event should only control the status / classList of layer.view elements.

The vector / mvt format changeEnd event should control whether the layer is displayed or not.

The layer show / hide event should only be fired when the zoom level changes in or out of range and not within or outside the range.

### mapview.view

It must be possible to override the locale.view when creating a mapview by providing the view object in the mapview constructor.

```js
mapp.Mapview({
      host: mapp.host,
      target: secondTop.querySelector('.map'),
      locale: locale,
      view: {
        lng: locationPin.lnglat[0],
        lat: locationPin.lnglat[1],
        z: 9
      },
      scrollWheelZoom: true,
    });
```

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -259,18 +259,8 @@ export default async layer => {
 
 function show() {
 
-  // Layer should be shown if possible.
-  this.zoomDisplay = true;
-
   // Push the layer into the layers hook array.
   this.mapview.hooks && mapp.hooks.push('layers', this.key);
-
-  if (this.tables && !this.tableCurrent()) {
-
-    // Layer is outside of zoomDisplay range.
-    delete this.display;
-    return;
-  }
 
   // Show the layer
   this.display = true;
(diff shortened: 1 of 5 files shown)


AlexanderGeere (approved):
Lgtm, I like that the layer stays on and comes back when you hit the right zoom level again. Nice!

RobAndrewHurst (approved):

dbauszus-glx:
The layer.zoomBtn display can be toggled if the button is right of the display toggle. This is a better indicator whether the layer is out of zoom range.

simon-leech (approved):
All working for me! 
Tested in default MAPP and also custom view and when zooming in/out the layers are re-drawn if they are within range.

PR #1142, opened 2024-02-15
clear highlight [candidates] on mouseleave and mousemove.

dbauszus-glx:
The mouseleave event must reset the highlights candidates set.

The mousemove event should only clear and reset the candidates if the sets are different.

### DIFF
--- a/lib/mapview/interactions/highlight.mjs
+++ b/lib/mapview/interactions/highlight.mjs
@@ -167,6 +167,9 @@ export default function (params) {
 
   function highlight(feature, e) {
 
+    // Clear curent highlight before assigning new.
+    clear()
+
     // Assign the layer and ID to the candidate object.
     mapview.interaction.current = Object.assign(feature, {
       layer: mapview.layers[feature.L.get('key')],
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (changes requested):
https://github.com/GEOLYTIX/xyz/assets/43169158/5e3ac43c-d1a6-4ee6-9d91-daed5e501866

I noticed that when the mouse enters one point, and doesn't exit but immediately goes into another point the previous point will still be highlighted

dbauszus-glx:
@RobAndrewHurst the current selection must be cleared at the beginning of the highlight method.

RobAndrewHurst (approved):
Nice One!

simon-leech (changes requested):
I'm seeing some issues with the long-click method here. 

1. If you long-click and then move off from that location, the popup remains on the mapview indefinitely. 
2. If you long-click, the geometry below the long click is highlighted and this highlight cannot be removed until the long-click popup is removed.

dbauszus-glx:
The shortcircuit has been removed as I don't really understand what it was doing.

The popup, e.g. from longclick should not stop the mousemove/highlight events.

The highlight should not be called twice for the same candidate.

simon-leech (approved):
This is looking good to me now - thanks Dennis!

PR #1143, opened 2024-02-15
If response from location-get is array, warn and return

simon-leech:
If the location-get returns two locations, it indicates the qID is not actually unique. This PR just warns and returns if this occurs, so you can see why a location cannot be selected.

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -107,7 +107,7 @@ async function syncFields(fields) {
   }
   // Check if the response is an array.
   else if (Array.isArray(response)) {
-    console.warn(`Location response returned more than one record for Layer: ${location.layer.key}.`)
+    console.warn(`Location response returned more than one record for Layer: ${this.layer.key}.`)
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
Can you make the checks as an else if to `!response` as these conditions would be mutually exclusive.

I'd also log the layer key and id to know which get query failed.

```js
if (!response) {

    console.warn('No data returned from location_get request.')
    return;

} else if (Array.isArray(response)) {

    console.warn('Location response returned more than one record. This is not allowed.')
    return;
}
```

simon-leech:
@dbauszus-glx Updated as per your comments - will mark this as ready for review 🙂

dbauszus-glx (changes requested):
Can you add the if else to the location/decorate check as well.

simon-leech:
@dbauszus-glx How's that? Updated to use the same if else in both location, I return a warning and then return the response from the query.

dbauszus-glx (approved):
lovely.

PR #1145, opened 2024-02-15
Load sync plugins in sync before async plugins

dbauszus-glx:
An array of plugin keys can be defined on the locale as `syncPlugins[]`

The default for this array if undefined is `['zoomBtn', 'admin', 'login']`

```js
locale.syncPlugins ??= ['zoomBtn', 'admin', 'login']
```

The plugins defined in this array will be executed in this order allowing us to configure the order in which buttons are added to the btnColumn element.

The zoomBtn, admin view link, and login/logout link have been converted into core plugins taking the elements out of the _default script and allowing us to control the order of all buttons.

### DIFF
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -134,14 +134,13 @@ export default (mapview) => {
   }).observe(mapview.Map.getTargetElement())
 
   // WARN!
-  mapview.locale.maskBounds && console
-    .warn('locale.maskBounds is set as mask:true in locale.extent')
+  mapview.locale.maskBounds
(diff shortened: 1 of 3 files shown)


simon-leech (changes requested):
This is working well for me - I like it! 

Only change - please can you add it to the Config Docs?

AlexanderGeere (approved):
Looks good to me!

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#syncplugins

simon-leech (approved):

PR #1147, opened 2024-02-19
Fix Numeric and Date Inputs

simon-leech:
The Tabulator Numeric and Date HeaderFilter Inputs were broken. 
This PR fixes the min and max inputs, using the same function for Numeric and Date by passing in a `min` or `max` flag - reducing the code maintenance required. 

The layerFilter Button was also broken, this now been fixed. 

This PR wants to be tested with: 

- Like, Set, Date, Numeric Filtering
- Like, Set, Date, Numeric Filtering with layerFiltering enabled
- layerFiltering and viewport Buttons also need to be tested.

### DIFF
--- a/lib/ui/utils/tabulator.mjs
+++ b/lib/ui/utils/tabulator.mjs
@@ -278,8 +278,8 @@ function like(_this) {
 
     function likeFilter(e) {
 
-      // Set layer filter and reload layer and table data.
-      if (headerFilterParams.layerFilter) {
+    // Set layer filter when layerFilter is true and headerFilterParams.layerFilter is true.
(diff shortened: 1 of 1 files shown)


RobAndrewHurst:
image

@simon-leech

simon-leech:
@RobAndrewHurst  Just pushed the fix!

dbauszus-glx (approved):
tested set and numeric filter

simon-leech:
@RobAndrewHurst @AlexanderGeere  
Please could someone test if this still works without a layerFilter button and with headerfilterparams.layerFilter set to true?

PR #1148, opened 2024-02-19
ol 8.2

dbauszus-glx:
Bump OL to 8.2

Move ol constructor into function closure to prevent a crash where mapp is loaded without openlayers imported prior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,6 @@ env
 package-lock.json
 node_modules/*
 
-*.pem
-*.crt
-
 public/css/.sass-cache/
 public/workspaces/*
 public/js/plugins/*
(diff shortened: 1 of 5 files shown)


RobAndrewHurst (changes requested):
Looks all good! Can we just include the vercel.json into the .gitignore as we don't want this to be committed by mistake to the codebase

dbauszus-glx:
> Looks all good! Can we just include the vercel.json into the .gitignore as we don't want this to be committed by mistake to the codebase

All *.json (including vercel.json) but the package.json should be excluded.

RobAndrewHurst:
> > Looks all good! Can we just include the vercel.json into the .gitignore as we don't want this to be committed by mistake to the codebase
> 
> All *.json (including vercel.json) but the package.json should be excluded.

Gotcha!

RobAndrewHurst (approved):

PR #1149, opened 2024-02-19
Allow hiding of Hovers and Labels

simon-leech:
We use checkboxes to control the visibility of hovers and labels via the Style Panel.

However, it is perfectly valid to want to hide these from view, being able to hide them has multiple uses: 
1. If a layer has just a default styling, then a style panel is rendered just for hover and label checkboxes. 
2. If a layer has lots of themes and uses setLabel and setHover, there is no need for 3 dropdowns that clutter the UI, and so hiding them would be much neater.

This PR introduces an optional flag `hidden` that can be applied to `layer.style.hover`, `layer.style.hovers`, `layer.style.label` and `layer.style.labels` to control whether the user sees this dropdown. 

This wants to be tested using: 
 - `layer.style.hover` 
 - `layer.style.hovers` 
 - `layer.style.label`
 - `layer.style.labels`
 - `setLabel`
 - `setHover`

### DIFF
--- a/lib/ui/layers/panels/style.mjs
+++ b/lib/ui/layers/panels/style.mjs
@@ -146,7 +146,7 @@ export default layer => {
     }))
   }
 
-  if (layer.style.hover) {
+  if (layer.style.hover && !layer.style.hover.hidden) {
 
     // Add checkbox to toggle label display.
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
works for me.

RobAndrewHurst (approved):

PR #1150, opened 2024-02-20
maplibre v4.0.2 bump and esm load

dbauszus-glx:
Update maplibre to 4.0.2 to be loaded from ESM instead of skypack.


RobAndrewHurst (approved):
Nice looks all good to me!

PR #1151, opened 2024-02-23
Add check for Style labelCheckbox

RobAndrewHurst:
## `labelCheckbox`

- Due to this PR [#1149](https://github.com/GEOLYTIX/xyz/pull/1149) it is possible for a layer style panel to not have a labelCheckbox.

- So I have added in a check on the `layer.style` object if its present before adding the `classList`

```js
 layer.style.labelCheckbox?.classList.add('disabled');
```


dbauszus-glx (approved):
lgtm

PR #1153, opened 2024-02-23
Use script module for default view

dbauszus-glx:
Use module script in order to facilitate top level await.

The module doesn't need to be loaded from host in the document head.

The module will only ever be required by the default view.

Log mapp version in module script.


RobAndrewHurst (approved):
Looks all good to me!

PR #1155, opened 2024-02-26
Remove not from if to remove

AlexanderGeere:
Clear all filters was just closing the filter panel, but had no effect on the layer.

This fix removes the filter and updates the layer accordingly.


dbauszus-glx (changes requested):
A filter set as default should not be cleared.

[Screencast from 26-02-24 11:30:02.webm](https://github.com/GEOLYTIX/xyz/assets/22201617/58737e4a-b5ea-4b39-93bf-b4bb89b6ddba)

The clear all filter text should not show when the filter drawer is collapsed.

AlexanderGeere:
@dbauszus-glx As of my most recent changes, the filter will now be reset to the filter that was present on load. This should correct the behaviour you observed while actually clearing the filter from the layer.

RobAndrewHurst:
There is still an issue flagged by the scan! Can you fix that @AlexanderGeere

dbauszus-glx:
This works for me now.

The `<li>` element is shown as selected in the filter dropdown for exsiting filter.

Filter without a field are not attempted to be deleted.

The zoom flag toggle has been removed as not required. Filter panel is unavailable outside zioom range.

Filter is deleted when key array length is falsy.

dbauszus-glx (approved):

PR #1156, opened 2024-02-26
OL9 support & googleMapTiles format

dbauszus-glx:
OL v9.0 bump and googleMapTiles support.

https://openlayers.org/en/latest/apidoc/module-ol_source_Google.html

https://github.com/openlayers/openlayers/pull/15465

https://github.com/openlayers/openlayers/releases/tag/v9.0.0

---
  - https://app.asana.com/0/0/1206708487854329


RobAndrewHurst (approved):
Nice one!

Are there any docs that we can reference for `maptype` and the `layerTypes`?

fnicollet:
@RobAndrewHurst The documentation for the parameters can be found here:
https://developers.google.com/maps/documentation/tile/session_tokens#required_fields

These options can be passed as options to the Google source constructor

Please note that as it was for 9.0, the Google source does not comply with Google's policies. A PR will soon be released to address those issues, see:
https://github.com/openlayers/openlayers/issues/15589
Note that you will need to manually add the Google logo to the map in your own code. The OpenLayers Google source example will be updated to show you how

PR #1158, opened 2024-02-27
The hover method must be assigned if not member of the hover object.

dbauszus-glx:
Fixes a bug where changing the hover through theme would assign a hover object without a method.


simon-leech (approved):
Fixed the described bug - thanks!

RobAndrewHurst (approved):

PR #1161, opened 2024-02-29
Clear filter gte lte fix

AlexanderGeere:
Clear filter fails on the type check on integer and numeric type filters.

Check if the type is undefined and then delete the field if that is the case.


simon-leech (changes requested):
@AlexanderGeere  The fix is working on this PR.

However, there's lot of other files that I think are part of your other formatter PR. Please can you resolve this?

PR #1163, opened 2024-02-29
fix clear for for gte and lte type

AlexanderGeere:
Numeric and integer type filters dont havea type, the type check was coming back as undefined.

Check the type is defined and then delete the type, or just delete the field.


dbauszus-glx (approved):
Looking good.

I removed nested layer.filter.current[filter.field] check.

PR #1164, opened 2024-03-01
Clear all filters fix - needs the second check

AlexanderGeere:
If the first check does the delete then Object.keys() cannot be performed on layer.filter.current[filter.field] as it will be undefined.

The second check ensures there is something run Object.keys on.


dbauszus-glx (approved):
👍 Good catch. I missed that condition previously.

simon-leech (approved):

PR #1166, opened 2024-03-01
Replace zoom draw style function with expression.

dbauszus-glx:
This was borked due to the assignment of a stroke to the style to the fill.


RobAndrewHurst (approved):
lgtm!

PR #1169, opened 2024-03-04
Null Geometries not being displayed when dependents used

simon-leech:
With this config below where - 
1. geometry_4326 is null 
2. geometry_not_null_4326 is populated by a trigger when a new point is added. 
``` json 
 {
      "display": true,
      "field": "geometry_4326",
      "fieldfx": "ST_AsGeoJSON(ST_TRANSFORM(geometry_4326,4326))",
      "type": "geometry"
}, 
{
      "display": true,
      "field": "geometry_not_null_4326",
      "type": "geometry",
      "dependents": ["geometry_4326"]
}
```

The geometry_4326 field will never display without the checkbox being toggled, due to an incorrect line in `geometry.mjs` that sets display to false without a value. 

This PR addresses that by removing the line of code.

---
  - https://app.asana.com/0/0/1206750193763473

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -46,9 +46,6 @@ export default entry => {
 
   entry.srid ??= entry.location?.layer?.srid
 
-  // Turn off display with no geometry to display.
-  entry.display = (entry.display && entry.value)
-
   // Drawing is only available within an edit context.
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
This works for me. I feel the assignment of the display flag from a combination of display and value is a legacy piece [of code].

PR #1170, opened 2024-03-05
Graduated zeroes error

AlexanderGeere:
When the value on graduated theme is zero the style is never set, even if there is a zero category. 

This PR will set the style if one has not been set. 

Using `graduated_theme` in bugs_testing, the polygon over birmingham will be blank on main, and will fill with green correctly on this PR.

---
  - https://app.asana.com/0/0/1206744686047675


dbauszus-glx (changes requested):
This doesn't seem right.

The three polygons have respective values of 0, 0.4, 0.6 and should display as the first three categories.

image

dbauszus-glx:
@AlexanderGeere Please check my last commit. I think this can be done with array methods.

```js
  if (!isNaN(catValue) && catValue !== null) {

    const isLassThan = cat => catValue <= cat.value;

    let index = theme.cat_arr.findIndex(isLassThan)

    let cat = theme.cat_arr.at(index)

    let catStyle = structuredClone(cat.style || cat)

    delete catStyle.label

    mapp.utils.merge(feature.style, catStyle)
  }
```

AlexanderGeere:
@dbauszus-glx  Works in the `graduated_theme` work space as well, looks a lot cleaner too. Nice!

simon-leech (approved):
Working for me too!

PR #1171, opened 2024-03-05
Fix issue snapping to mvt

dbauszus-glx:
This PR fixes an issue attempting to snap to an MVT layer with zoom level restrictions.

```js
"draw": {
  "polygon": {
    "snap": {
      "layer": "iuc"
    }
  }
}
```

The snap interaction should attempt to display the layer. Snapping is only possible to loaded features. Adding features to the snap source must be tried. MVT layer have duplicate features add duplicate features would crash this process.


simon-leech (changes requested):
@dbauszus-glx  
✅ This has resolved the issue when create a new geometry and trying to snap to a layer. 
❌ Editing an existing geometry throws a warning in the console. 
image


❓ In looking through old configuration - I see a lot of 
`layer.snap:true` - please can we add a warning to remove this as it doesn't do anything? 
`layer.snap.layer_key - please can we add a warning to update this to `layer.snap.layer.layer_key`?

dbauszus-glx:
snap.true should work. this should snap to the layer itself. might be a bug. i haven't tested editing but only drawing, will need to look into this.

RobAndrewHurst (approved):
Nice @dbauszus-glx 

All worked for me!

I have added in additional documentation for the mapview snap interaction and the layer mvt format modules.

I added in the params/properties at the top of the each function and nested them.

eg. 

```
/**
@param layer
@param [layer.srid] - description
@param [layer.tables] - description
*/

Please have a look and let me know what you think.

dbauszus-glx:
@RobAndrewHurst Looking good. I think just a reference for mapview, style, etc object is enough once these these objects are sufficiently described in place.

simon-leech:
Pushed a commit to 'allow' string assigment of snap: layer_key but warn about incorrect configuration.
image


This is still not working on editing a geometry - should this be a separate PR @dbauszus-glx  ?

PR #1174, opened 2024-03-06
Use password_reset for registration

dbauszus-glx:
User registration must set the password as password_reset not as password.

Otherwise a user trying to login with an unverified password will create failed logins.

The verification set's the password_reset as password if successful.

Added reqHost util which generates the host value for emails. Takes localhost into account. Logger key `reqhost` logs the host value.

### DIFF
--- a/mod/user/_user.js
+++ b/mod/user/_user.js
@@ -2,6 +2,8 @@
 @module /user
 */
 
+const reqHost = require('../utils/reqHost')
+
 const view = require('../view')
 
 const methods = {
@@ -64,9 +66,7 @@ module.exports = async (req, res) => {
     return res.send(`Failed to evaluate 'method' param.`)
   }
 
-  req.params.host = `${req.headers.origin 
-    || req.headers.host && 'https://' + (process.env.ALIAS || req.headers.host)}${process.env.DIR}`
-    || req.headers.referer && new URL(req.headers.referer).origin 
+  req.params.host = reqHost(req)
 
   if (!req.params.user && (method.login || method.admin)) {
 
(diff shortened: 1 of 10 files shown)


simon-leech (changes requested):
**Localhost**
* ❓ Blocked myself  > Trying to login gives me the 'Your account or password is incorrect'. If you have forgotten reset password' message.  > Then reset my password and it gives 'Your account has been blocked' message and a 500 error in the console. 
 Shouldn't this give the 'Your account has been blocked' straight away? As resetting your password has no effect here? 
* ✅ Unblocked myself > Able to login fine 
* ✅ Reset password > Received email > Attempted to login with new password that had not been verified > Unable to use new password
*  ✅ Reset password > Received email >Verified password reset > Able to login using new password & unable to login using old password
* ✅ 3 failed attempts > Received email and correct error message > unable to login until reset password > Old password no longer works 
*  ❓ Sign up when already in the ACL > I get a password reset email > Shouldn't we prevent signing up on the same email?
* ✅ Initial sign up > Received email > Verified > Email for approved > Able to login

dbauszus-glx:
@simon-leech Signing up resetting will always reset send a new verification mail. The first mail may be lost blocked in spam.

Blocked account should not be able to reset password or login.

Login should return `Authentication failed`, registration should return `User blocked`.

image

image

dbauszus-glx:
Registration should set both password and password_reset. This way an administrator can verify and approve an account without the account [email] being verified.

simon-leech:
Localhost testing for me looks good now! 😊

RobAndrewHurst (approved):

simon-leech (approved):
Tested locally and on a deployed instance and all appears to be good for me!

PR #1176, opened 2024-03-06
Fixes for infoj entry and filter issues

simon-leech:
**Issue**
* A `infoj` entry of `type:integer` will not be formatted correctly, neither formatting nor rounding is applied. 
* This was due to a check in line 30 of `numeric.mjs` that was conditional on `entry.formatterParams.locale`. However, there was no default applied. 
* I also noticed that providing a `maximumFractionDigit` of >2 when using certain locales (e.g. `en-GB`) was not applied.
* The `numericFormatter.mjs` was not checking if the type was integer so it was not formatting the value in the filter in this case. 

**Fixes** 
* The default locale is set to `en-GB`, but can be overriden in the configuration of an `infoj` entry. 
* If a `maximumFractionDigit` of >2 is applied, then the `minimumFractionDigit` (if not provided in the configuration), is set to the same value, so that this is applied. 
* `numericFormatter.mjs` now checks if the type if integer and sets the default locale to `en-GB`.  

**How to Test** 
Please test the various configuration options - examples below but this may not be exhaustive.
* `type: integer` + `formatterParams.locale` + `filter: integer`
* `type: numeric` + `round: 3` + `filter: numeric`
* `type: integer` + `filter: integer` 
* `type: numeric` + `filter: numeric`
``` json 
 {
      "field": "X",
      "type": "integer",
      "prefix": "£",
      "inline": true,
      "formatterParams": {
        "locale": "ar-EG"
      },
      "filter": {
        "type": "integer"
      }
    },
    {
      "field": "Y",
      "type": "numeric",
      "round": 3,
      "prefix": "£",
      "inline": true,
      "skipNullValue": true,
      "filter": {
        "type": "numeric"
      }
    },
    {
      "field": "Z",
      "type": "numeric",
      "prefix": "£",
      "inline": true,
      "skipNullValue": true,
      "filter": {
        "type": "numeric"
      }
    },
    {
      "field": "A",
      "type": "integer",
      "prefix": "£",
      "inline": true,
      "skipNullValue": true,
      "filter": {
        "type": "integer"
      }
    },

```

### DIFF
--- a/lib/ui/elements/numericFormatter.mjs
+++ b/lib/ui/elements/numericFormatter.mjs
@@ -80,10 +80,10 @@ function getFormatterParams(entry, inValue) {
   let separators = [entry.formatterParams?.thousand, entry.formatterParams?.decimal]
   let localeSeparators = Array.from(new Set(formattedValue.match(/\D/g)))
   localeSeparators[1] ??= '.'
-  
+
   //If not supplied look in the formatted string
   separators[0] = separators[0] ? separators[0] : localeSeparators[0]
-  separators[1] = separators[1] ? separators[1] : 
+  separators[1] = separators[1] ? separators[1] : localeSeparators[1]
 
   formattedValue = `${negative ? '-' : ''}${formattedValue}`
   return { value: formattedValue, separators: { thousands: separators[0], decimals: separators[1] } }
--- a/lib/ui/elements/numericFormatter.mjs
+++ b/lib/ui/elements/numericFormatter.mjs
@@ -19,8 +19,6 @@ export function getSeparators(entry) {
   //Do nothing if entry is null or no formatter is present
   if (!entry) return;
   if (!entry.formatter && entry.type !== 'integer') return { decimals: '.' };
-  entry.prefix = ''
-  entry.suffix = ''
   //Use a number to determin what the fields 
   //formatted input would look like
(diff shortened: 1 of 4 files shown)


AlexanderGeere:
I noticed the prefixes and suffixes werent not applying so i corrected that.

Prefixes and suffixes shouldnt apply in the filter input, nor will rounding, precision or min/maximumFractionDigits.

Looks good to me otherwise.

dbauszus-glx (changes requested):
This working well but the default should be no locale.

Then if the user object is present with a language the user language should be used as the locale for the number formatting.

dbauszus-glx (approved):
Works well with one small change.

If locale is set use this locale.

If locale is nullish use the user language as locale.

If locale is NULL, do not use a locale.

```js
  // Assign user language as locale if nullish but not null.
  if (entry.formatterParams.locale !== null) {
    entry.formatterParams.locale ??= mapp.user.language;
  }
```

simon-leech:
Updated the config docs to highlight this change. All ready to merge! 

https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#integer-1

RobAndrewHurst (dismissed):
Please add in the updated documentation into the modules.

PR #1177, opened 2024-03-08
ACL List changes

RobAndrewHurst:
## ACL List Changes

1. Added in documentation for the user/list module 📖 
2. Added in an optional param to search on a users email. 👤 

This will enhance the test automation of the login process as when verifying the user we don't need a full list of users. Just one in particular.

### DIFF
--- a/mod/user/list.js
+++ b/mod/user/list.js
@@ -1,11 +1,38 @@
 /**
+### List Users
+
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):
Looks good.

dbauszus-glx (approved):
lgtm

PR #1180, opened 2024-03-08
bump crypto libraries

dbauszus-glx:
The JWT library has been bumped to 9.0.2 which updates semver to 7.5.4 to resolve CVE-2022-25883

https://github.com/auth0/node-jsonwebtoken/blob/master/CHANGELOG.md

The bcryptjs lib has been replaced with the bcrypt native library. Bcryptjs hasn't been updated for 7 years and performs slower at run time. The native library will be compiled at init but performs better at run time.

https://www.npmjs.com/package/bcrypt


RobAndrewHurst (approved):
Nice one! looks all good to me!

AlexanderGeere (approved):
Looks good to me!

PR #1181, opened 2024-03-08
remove nanoid

dbauszus-glx:
NanoID was only used on the host to generate a unique[ish] id for the session and and process ID.

The random hex sequence generated by the builtin node crypto module can be used to the same effect.

One dependency less to worry about.


dbauszus-glx:
@cityremade I replaced NANO_SESSION in mod/auth with USER_SESSION

PR #1185, opened 2024-03-15
Role merge

dbauszus-glx:
Role objects should not be merged in the mapp library when a layer is decorated.

The method to merge role objects has been added to the Roles util module.

The role objects maybe on the local, layer, or any nested object (e.g. infoj entries).

The Roles.objMerge() method will be called before returning a layer or locale from their respective workspace modules.

This makes the Role.filter() method which is only called in the query mod after the layer is returned from getLayer mod redundant.


dbauszus-glx:
I removed the jest test.

RobAndrewHurst (changes requested):
Giving this a test on bugs-testing/roles/roles_filter_workspace.json

I can see a layer with the "Super" role defined even though my user doesn't have that role assigned.

image

dbauszus-glx:
The objMerge must happen in the workspace module before the obj is returned from the response.

AlexanderGeere:
@dbauszus-glx the roles are affecting the layers correctly, but when I tested on `bugs-testing/roles/roles_merge_workspace` and set my role to `Super` or `user` I can see both dataviews.

RobAndrewHurst (approved):
Nice one! All working for me again!

simon-leech (changes requested):
Role restriction on locales / layers is working for me. 

1. But merging layer information is not working with roles. 
A roles object on a layer of 
``` json 
"roles": { 
"super": { 
"toggleLocationViewEdits": true
},
}
```
When I don't have the super role I am still able to toggle editing, i shouldn't be able to do so. 

2. If I have a default locale containing layers, and try to remove one layer from a specific locale - I hit an error message. 
e.g. 
``` json 
"locale": {
"layers":{ 
"test": {}, 
"test_a": {}
} 
}, 
"locales": {
"UK": {
"layers": { 
"test": null
}
}
}
```

dbauszus-glx:
@simon-leech the check on negated roles was borked. this should now work.

The negatedRole flag should be assigned for every user without the 'foo' role.

```js
"!foo": {
  "negatedRole": true
}
```

dbauszus-glx:
@simon-leech 

The '*' role will now be spliced into the user_roles array prior to the objMerge.

dbauszus-glx:
@simon-leech 

A null layer should now remove the layer from layers list.

```js
  "locale": {
    "layers": {
      "OSM": {
        "display": true,
        "format": "tiles",
        "attribution": {
        }
      }
    }
  },
  "locales": {
    "test": {
      "layers": {
        "OSM": null,
        "retailpoints": {}
    }
  }
}
```

simon-leech (approved):
I've tested this and it looks good to me - 
Layer restrictions
Layer null in locale from default locale
Infoj role restrictions
Layer editing role restrictions 

All of these were working for me - i tested using a role, two roles and no role and I got the expected outcomes.

simon-leech (changes requested):
Noticed an issue with filtering and using a filter.current or filter.default. 
When you toggle a thematic option in a style you end up with an incorrect SQL WHERE Clause that ends with an AND incorrectly.

RobAndrewHurst:
@dbauszus-glx I have updated the tests for the roles utility module!

PR #1186, opened 2024-03-18
Assign params as URL paramstring to tile layer URI

dbauszus-glx:
The paramString util must return an empty string if provided with a falsy params argument.

The tile layer format should create a paramString to be appended to the layer.URI.

Params such as the api keys should be provided as params in the layer json. eg.

```js
      "here_hybrid": {
        "name": "HERE Hybrid",
        "format": "tiles",
        "URI": "https://{1-4}.aerial.maps.ls.hereapi.com/maptile/2.1/maptile/newest/hybrid.day/{z}/{x}/{y}/256/png8?",
        "params": {
          "apiKey": "here"
        },
        "attribution": {
          "© Here": "https://www.here.com/",
        }
      },
```

### DIFF
--- a/lib/layer/format/tiles.mjs
+++ b/lib/layer/format/tiles.mjs
@@ -10,10 +10,12 @@ export default layer => {
 
   layer.projection ??= 'EPSG:3857'
 
+  layer.paramString ??= mapp.utils.paramString(layer.params)
+
   layer.L = new ol.layer.Tile({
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
@dbauszus-glx 
This then isn't backwards compatible with previous here configuration?

RobAndrewHurst (approved):
Nice!

Looks all good to me!

PR #1188, opened 2024-03-20
Provide buffer param for cloudfront provider

dbauszus-glx:
The cloudfront endpoint returns either json or text from the fetch response.

Setting the `buffer=true` url parameter will return a buffer from the fetch response.

The buffer method is a member of the node-fetch module which must be required by the cloudfront module.

```
/api/provider/cloudfront?content_type=application/pdf&buffer=true&url={RESOURCES}/mapp_uk/userguide.pdf
```

The content_type must be set to `application/pdf` for the provider endpoint to set the response header with the buffer content.

---
  - https://app.asana.com/0/0/1206886274025768


simon-leech (approved):
This is working for me and deployed an instance in which its working too!

RobAndrewHurst (approved):

PR #1192, opened 2024-03-21
Remove empty cat filter from theme legend

dbauszus-glx:
Having a current filter, then toggling a category filter on and off will crash the wkt query because the check on the empty filter passes without checking for the key length.

---
  - https://app.asana.com/0/0/1206900457248763


simon-leech (approved):
Tested using filter.current and toggling the styling using both the cat filter itself and the switch all method and no errors and all working as expected.

PR #1194, opened 2024-03-22
Throw error for invalid query parameter substitution

dbauszus-glx:
The query module should throw an error when an invalid substitute param is checked.

image


freddie-wallace (approved):
Works as expected. Provides information which helps solve errors when configurating an instance.

RobAndrewHurst (approved):

PR #1198, opened 2024-03-27
Provide label and names for input elements

dbauszus-glx:
Labels associated with an input should contain the input.

Input elements should have a name.

This is to prevent issues in the google chrome dev tools.

image


simon-leech (commented):
Works for me - i'm not seeing any warnings anymore with checkboxes, gazetteer or sliders.

simon-leech (approved):

PR #1200, opened 2024-03-28
mapp.utils.polygonIntersectFeatures

dbauszus-glx:
This PR adds the  mapp.utils.polygonIntersectFeatures() method.

A type:polygon draw interaction will check for snap features coordinates to intersect the drawn polygon feature geometry.

Intersecting features are returned to the drawendCallback method.

The drawendCallback can be defined to store the feature id in a set and add the feature to a custom vector layer for visualisation.

```js
function drawendCallback(features) {

  features.forEach(feature => {

    const id = feature.get('id')

    if (layer.selectedLocations.ids.has(id)) return;
    layer.selectedLocations.ids.add(id)
    cloneSource.addFeature(feature)
  })
}
```

### DIFF
--- a/lib/utils/_utils.mjs
+++ b/lib/utils/_utils.mjs
@@ -66,6 +66,8 @@ import merge from './merge.mjs'
 
 import paramString from './paramString.mjs'
 
+import {polygonIntersectFeatures} from './polygonIntersectFeatures.mjs'
+
 import promiseAll from './promiseAll.mjs'
 
 import queryParams from './queryParams.mjs'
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/assets/43169158/a6bf06b4-3469-40fe-a3c1-2310308baee5

Noticed some off behaviour.

- If you select a bunch of features individually and click on them after selection the highlight doesn't disappear, but the counter does decrease.
- Then if you clear the selection and try select the features you just selected you wont be able to select them.

RobAndrewHurst (approved):

PR #1203, opened 2024-04-05
Codi 🐶 Web Test Runner 🏃

RobAndrewHurst:
### Codi 🐶 Web Test Runner 🏃 

I have implemented a Web Test Runner into Codi and Implemented it into the test_view.

- This allows for us to give multiple files into a loaded in a script on a view.
- Counts all the passed/failed tests ✅⛔

In this instance I load one module, that calls another which means we can chain these tests, and still keep the count.

Please have a look and let me know if you have any suggestions.
 

---
  - https://app.asana.com/0/0/1206998202384504

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "clean-jsdoc-theme": "^4.2.17",
-    "codi-test-framework": "^0.0.15",
+    "codi-test-framework": "^0.0.23",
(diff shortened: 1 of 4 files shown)


AlexanderGeere (approved):
Looks good.

simon-leech (approved):
Brilliant - thank you Rob! So good having the summary at the end!

PR #1206, opened 2024-04-12
Modal Target Adjustment 🎯

RobAndrewHurst:
## Modal Target Adjustment 🎯

This PR addresses the issue where the modal goes out of view when the map is resized. The changes ensure that the modal's position is adjusted dynamically to stay within the target/map boundaries. Note the modal adjust's to a target, so in plugins and other uses of the modal it needs to be the map target if it needs to work with tabs at the bottom of the map.

## Changes ✍️ 

- Introduced a [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to observe changes in the map's dimensions.
- Created an `adjustModalPosition()` function to calculate the maximum allowed positions for the modal based on the current map dimensions and adjust the modal's position accordingly.
- Modified the `closeModal()` function to disconnect the `ResizeObserver` when the modal is closed.

## Implementation Details

1. A new [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instance named `resizeObserver` is created, and the `adjustModalPosition()` function is passed as the callback. This function will be called whenever the observed element's dimensions change.

2. The `resizeObserver.observe(modal.target)` method is used to start observing changes in the map's dimensions.

3. Inside the `closeModal()` function, the `resizeObserver.disconnect()` method is called to stop observing changes when the modal is closed.

4. The `adjustModalPosition()` function calculates the maximum allowed positions for the modal based on the current dimensions of the target. It adjusts the modal's `left` and `top` positions using `Math.min()` and `Math.max()` to ensure that the modal stays within the target(map) boundaries. If the modal's position exceeds the map boundaries, it will be adjusted to the nearest valid position.

## Testing 🧪 

- Tested the changes by resizing the map and verifying that the modal stays within the view.
- Ensured that the modal can still be dragged and positioned within the map boundaries.
- Confirmed that the modal closes properly and the `ResizeObserver` is disconnected when the modal is closed.

---
  - https://app.asana.com/0/0/1207053242940136

### DIFF
--- a/lib/ui/elements/modal.mjs
+++ b/lib/ui/elements/modal.mjs
@@ -26,17 +26,18 @@ export default (modal) => {
     // Calc left from right and offSetWidth
     modal.left = modal.target.offsetWidth - modal.node.offsetWidth - parseInt(modal.right)
   }
-  
+
   modal.node.style.top = `${modal.top || 0}px`
   modal.node.style.left = `${modal.left || 0}px`
 
   modal.node.addEventListener('mousedown', handleStartEvent);
-  modal.node.addEventListener('touchstart', handleStartEvent);      
+  modal.node.addEventListener('touchstart', handleStartEvent);
 
   // Function to handle the modal closure
   function closeModal(e) {
     e.target.closest('.modal').remove();
     modal?.close?.(e);
+    resizeObserver.disconnect();
   }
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Lovely - working well for me thank you @RobAndrewHurst ! 👏

AlexanderGeere (approved):
Working Well! nice one.

PR #1209, opened 2024-04-15
Update layer test

RobAndrewHurst:
### Layer Test Changes ✍️ 

I updated the layer test to: 
- check if we get a location rather than asserting if we get an id back from the `getLastLocation` template.

> Please check if this reaches other edges cases when it comes to the last location of a layer 🌐


simon-leech (approved):
Nice one @RobAndrewHurst  - prevents an issue where the tests failed if no data was in a table!

PR #1210, opened 2024-04-17
Update Openlayers version and add warning for script loader.

dbauszus-glx:
The  OL version in the default script has been bumped to 9.1

The OL script loader will be added in the mapp script if OL is not already loaded. It doesn't make sense to load the script if already loaded.

mapp.utils.olScript() will call mapp.ol.load() for legacy support.

mapp.ol.current is the current supported version. e.g. 9.1

The mapp module will warn if the loaded version of OL is below the current version.

Screenshot from 2024-04-17 11-48-17

### DIFF
--- a/lib/mapp.mjs
+++ b/lib/mapp.mjs
@@ -58,7 +58,7 @@ if (window.ol === undefined) {
 
   if (olVersion < _ol.current) {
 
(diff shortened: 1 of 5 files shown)


simon-leech (approved):
This is working well for me - and like the warning in views!

RobAndrewHurst:
image

@dbauszus-glx Fixed the warning for if the ol version is bellow what is defined.

PR #1211, opened 2024-04-17
Add time param to tile request

dbauszus-glx:
The browser may try to access a tile with the same request params from cache.

The cache may be invalidated by Antivirus software, and/or browser/OS settings.

The data in the tile may have changed.

A ERR_CACHE_READ_FAILURE type failure may cause the layer to flicker.

This PR adds a timestamp to the request params in the tileUrlFunction to make the request for the same tile unique.


simon-leech:
@dbauszus-glx  Do we still want to implement this one too? 
Even though it looks like PR #1210  fixes the issue ?

PR #1216, opened 2024-04-18
Fix for Numeric Filter on field with no data

simon-leech:
A numeric filter on a field that contains no data will currently fail. 
This is as the response from the `template_max` and `template_min` are undefined.
image

This PR: 
1.  checks for the response, removing the error, 
2.  if no response is returned - returns text instead of a slider to tell the user of this. 
unnamed
3. Fixes `parseInt` to `Math.round` as values of say 1001.99 would be rounded to 1001 not 1002 previously.

---
  - https://app.asana.com/0/0/1207110786328348

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -101,7 +101,7 @@ async function generateMinMax(layer, filter, minmax) {
 
   // If the response contains response.min or response.max - we can assume the query was successful.
   if (response[minmax]) {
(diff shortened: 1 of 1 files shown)


simon-leech:
@RobAndrewHurst  Yes please - just fixing the SonarCloud now - then over to you! :)

AlexanderGeere (approved):
Looks good to me

PR #1218, opened 2024-04-22
Add st_distance_ab and st_intersects_ab query templates

dbauszus-glx:
Default queries for location field lookup based on the locations geometry.

The entry.nullValue can be used to assign a value if the query does not return a response.

This should make the query_fields plugin redundant.

---
  - https://app.asana.com/0/0/1207158269914913

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -141,8 +141,15 @@ export default (location, infoj) => {
           .xhr(`${entry.host || entry.location?.layer?.mapview?.host || mapp.host}/api/query?${paramString}`)
           .then(response => {
 
-            // Assign query response as entry value.
-            entry.value = entry.field? response[entry.field] :response;
+            if (response) {
+
+              // Assign query response as entry value.
+              entry.value = entry.field ? response[entry.field] : response;
+
+            } else {
+
+              entry.value = entry.nullValue
+            }
 
             // Check whether entry should be skipped.
             if (skipEntry(entry)) {
--- a/lib/utils/queryParams.mjs
+++ b/lib/utils/queryParams.mjs
@@ -49,9 +49,6 @@ export default _this => {
 
     // qID will be taken if a location object is provided with the params.
     qID: _this.queryparams.qID ? _this.location?.layer?.qID : undefined,
-
-    // geom will be taken if a location object is provided with the params.
-    geom: _this.queryparams.geom ? _this.location?.layer?.geom : undefined,
     
     // Email will be taken if a location object is provided with the params.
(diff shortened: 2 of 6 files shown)


simon-leech:
I made a few changes here- 
1. Added qID and geom to the queryparams so you can pass them as true and they are accessed from the layer values. 
2. Added a new query for multiple rows returned (closest distance) 
3. Updated the st_intersect and st_distance queries based on feedback from @danielsawyers 

Example config here 
``` json 
"infoj": [
  {
    "title": "Test",
    "inline": true,
    "field": "field",
    "query": "st_intersects_ab",
    "queryparams": {
      "qID": true,
      "id": true,
      "table": true,
      "geom": true,
      "field_b": "field",
      "as": "field",
      "table_b": "schema.table",
      "geom_b": "geom_4326"
    },
    "run": true
  },
  {
    "title": "Distance to X",
    "inline": true,
    "field": "dist",
    "query": "st_distance_ab",
    "queryparams": {
      "qID": true,
      "id": true,
      "table": true,
      "geom": true,
      "table_b": "schema.table",
      "geom_b": "geom_p_4326"
    },
    "run": true
  },
  {
    "display": true,
    "label": "Location Table",
    "type": "dataview",
    "target": "location",
    "query": "st_distance_ab_multiple",
    "queryparams": {
      "qID": true,
      "id": true,
      "table": true,
      "geom": true,
      "as": "field",
      "field_b": "field_name",
      "limit": 10,
      "table_b": "schema.table",
      "geom_b": "geom_p_4326"
    },
    "table": {
      "layout": "fitColumns",
      "columns": [
        {
          "field": "store_name",
          "title": "Name"
        },
        {
          "field": "dist",
          "title": "Distance"
        }
      ]
    }
  }
]
```

simon-leech (approved):

dbauszus-glx:
@simon-leech The changes do not work. I highly recommend against using the geom lookup. Not all tables have a geom field. Some layer may use a geoms object. In this case it will be impossible to geom: true, and geom:"whatever" would also be recognised as geom: true. In both cases this will mean crunch for the query. Additionally the cognitive load of the queryParams method is pushed over the limit by adding the geom true flag. Using geom: string does have none of these problems.

simon-leech:
> @simon-leech The changes do not work. I highly recommend against using the geom lookup. Not all tables have a geom field. Some layer may use a geoms object. In this case it will be impossible to geom: true, and geom:"whatever" would also be recognised as geom: true. In both cases this will mean crunch for the query. Additionally the cognitive load of the queryParams method is pushed over the limit by adding the geom true flag. Using geom: string does have none of these problems.

Good point @dbauszus-glx - I have removed "geom": true and just using "geom": string is all working :)

PR #1220, opened 2024-04-23
Check and remove filter params from request for query templates without a filter placeholder.

dbauszus-glx:
The location_new layer query template does not have a ${filter} substitute param and will crash if a query is sent for a layer with a default filter.

---
  - https://app.asana.com/0/0/1207135159328593


AlexanderGeere (approved):
Looks good to me. Nice!

simon-leech (approved):
This one is working all good for me - tested using a filter.default on a layer, and adding a new location to this layer

PR #1231, opened 2024-04-26
add ID to location_get warning

KieronFerrey:
When running the test_view it is useful to get the ID of a failed location request for diagnosing the issue. I updated the console warnings to return the ID.


RobAndrewHurst:
What instance can I test this with? Please DM this information 👍

PR #1233, opened 2024-04-29
Circle changes

simon-leech:
Opening this up as a draft as its not ready yet.

Changes Included:

If you provide a circle.radius and circle.units then the panel is not provided, and instead you just have the button to begin drawing.
  "circle": {
          "label": "1km Catchment",
          "radius": 1000,
          "units": "meter"
        }

---
  - https://app.asana.com/0/0/1207171659105005

### DIFF
--- a/lib/ui/elements/drawing.mjs
+++ b/lib/ui/elements/drawing.mjs
@@ -28,8 +28,8 @@ mapp.utils.merge(mapp.dictionaries, {
     draw_polygon: 'Polygon',
     draw_rectangle: 'Rechteck',
     circle_config: 'Zirkel Einstellung',
-    draw_circle: '2 Punkt Zirkel',
-    draw_circle_2pt: 'Zirkel',
+    draw_circle: 'Zirkel',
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
Can you break this down into two separate  PR? I am ok with the circle change, but believe the geojson import wants to be broken down. Can you create an issue with more details in regards to the specification for the geojson import?

simon-leech:
@dbauszus-glx  I have removed the geojson import from this PR - just got the circle changes now :)

dbauszus-glx (approved):

dbauszus-glx:
I added the hidePanel flag in order to maintain the current default behaviour. A radius and unit will always be assigned as default. The config object will just override the defaults. The dictionary label can be assigned in the defaults since the config will override. A spread operator should be used for this rather than object assignment.

simon-leech (commented):
Added to the Config Docs here - 
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#circle

But all working well for me! :)

PR #1234, opened 2024-04-29
Encode Roles in Admin Panel

simon-leech:
The roles need to be `encoded` when passed to the database. 
This is to prevent a role of e.g. `Test&Test_more` being passed to the database as just `Test`.


RobAndrewHurst (approved):

AlexanderGeere (approved):
Looks good.

dbauszus-glx (approved):
Working as expected. Well done.

PR #1236, opened 2024-04-29
Filter Fix

simon-leech:
In the `filter.mjs` - we were calling the `generateMinMax` function to get the data for the minimum or maximum values of the filter sliders. 
However, we were not awaiting these which meant that the values were not being returned correctly, and the no data message was being shown when there was actually data.
---
  - https://app.asana.com/0/0/1207195219504067

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -145,7 +145,8 @@ async function generateMinMax(layer, filter, minmax) {
   })}`);
 
   // If the response contains response.min or response.max - we can assume the query was successful.
-  if (response[minmax]) {
+  // Have to check that response is not null as well
+  if (response && response[minmax]) {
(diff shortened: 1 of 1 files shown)


simon-leech:
Just pushed another commit to check that the `filter.min` and `filter.max` are numeric, if not the no data message is returned

KieronFerrey (approved):
Looks good to me

AlexanderGeere:
Added a check for whether response is `null` before checking `response[minmax]`.

Added a simple test workspace as well: [Filter workspace](https://github.com/GEOLYTIX/xyz_resources/blob/bugs-testing/bugs_testing/filter/filter.json)

AlexanderGeere (approved):

dbauszus-glx (approved):
lgtm

PR #1237, opened 2024-04-29
Tabulator 'Set' Headerfilter - no data from distinct query

AlexanderGeere:
### What
Using the `set` `headerfilter` and `distinct` in the `headerFilterParams` would create an error, as the dropdown would attempt to iterate an empty list .

This PR should correct that by making a dropdown that has a no data message.

### Test Space
[`bugs_testing/dataviews_filter/workspace.json`](https://github.com/GEOLYTIX/xyz_resources/tree/bugs-testing/bugs_testing/dataviews_filter)

---
  - https://app.asana.com/0/0/1207190169901915

### DIFF
--- a/lib/ui/utils/tabulator.mjs
+++ b/lib/ui/utils/tabulator.mjs
@@ -515,7 +515,7 @@ function set(_this) {
     // Create dropdown for render.
     let dropdown = mapp.utils.html.node`<div class="ul-parent">`
 
-    if (headerFilterParams.distinct) {
+      if (!headerFilterParams.options || headerFilterParams.options.length === 0) {
 
       // Query distinct field values.
       mapp.utils.xhr(`${_this.layer.mapview.host}/api/query?` +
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
@AlexanderGeere can you add this to the tabulator plugin? ui.utils.tabulator will be removed from the mapp library as soon as the plugin is reviewed.

AlexanderGeere:
@dbauszus-glx Will do.

Changed it slightly so the dropdown works but just says no options.

simon-leech (approved):
All working well for me 👍

dbauszus-glx (changes requested):
The method should not be dependent on the distinct flag but whether an options array is provided.

See filters for reference. https://github.com/GEOLYTIX/xyz/blob/main/lib/ui/layers/filters.mjs

If an array of options is defined than this should be used. If not, a query should be attempted to populate the options for the filter.

dbauszus-glx (approved):

PR #1238, opened 2024-04-29
ui.elements.searchbox;

cityremade:
Adds a new ui element searchbox.
Takes the following parameters:
```
mapp.ui.elements.searchbox({
    target: container,
    name: 'your-search-input', 
    placeholder: 'Search something'
    searchFunction: my_search_function 
  })
```

* 'search-input' name by default
* my_search_function(e), no search function defined by default, should there be one?

### DIFF
--- a/lib/ui/elements/_elements.mjs
+++ b/lib/ui/elements/_elements.mjs
@@ -15,6 +15,7 @@ import dropdown_multi from './dropdown_multi.mjs';
 import btnPanel from './btnPanel.mjs';
 import legendIcon from './legendIcon.mjs';
 import modal from './modal.mjs';
+import searchbox from './searchbox.mjs';
 import slider from './slider.mjs';
 import slider_ab from './slider_ab.mjs';
 import { numericFormatter, getSeparators } from './numericFormatter.mjs';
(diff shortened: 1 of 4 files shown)


dbauszus-glx:
I've added a check on the searchbox argument. Otherwise this would crash the mapp process.

```js
mapp.ui.elements.searchbox()
```

image

dbauszus-glx (approved):
working well for me with this small test plugin.

```js
mapp.plugins.searchbox = () => {

    const layersEl = document.getElementById("layers");
  
    if (!layersEl) return;

    const searchbox = mapp.ui.elements.searchbox()

    layersEl.append(searchbox.target)
}
```

simon-leech:
I also tested using this a similar plugin - 
``` js
mapp.plugins.searchbox = () => {

    const layersEl = document.getElementById("layers");

    if (!layersEl) return;

    // Name the searchbox
    const searchboxparams = {}
    searchboxparams.name = "Search Simon Test";
    searchboxparams.placeholder = "Search for layers";
    searchboxparams.searchFunction = oninput;

    const searchbox = mapp.ui.elements.searchbox(searchboxparams)

    function oninput(event) {
        const searchStr = event.target.value;

        // Clear the searchbox list
        searchbox.list.innerHTML = '';

        // Mapp.dictionary is an object that contains all the terms that are in the dictionary
        // Loop through the dictionary and check if the searchStr is in each term
        Object.values(mapp.dictionaries.en).forEach(term => {
            if (term.toLowerCase().includes(searchStr.toLowerCase())) {
                // Add matching term to the searchbox list
                searchbox.list.append(mapp.utils.html.node`<li><span>${term}`);
            }
        });
    }
    layersEl.append(searchbox.target)
}
```

image


This is working well for me and nice and easy to set up :)

simon-leech (approved):

cityremade:
thanks for reviews, good to go 🙌🏻

PR #1240, opened 2024-05-01
Add check on cluster distance being less than 1

simon-leech:
This PR adds a check to the `cluster.distance` value to ensure it is more than 1. 
I noticed a lot of configuration using values less than 1, but this currently silently fails. 
This PR simply warns about the issue, forcing the developer to resolve it to remove the warning.

---
  - https://app.asana.com/0/0/1207208979036434


dbauszus-glx (approved):
Looks good. I just shortened the warning. And inidicated that the cluster will be removed after the warning.
image

PR #1241, opened 2024-05-02
Fix key value dictionary for the custom views

KieronFerrey:
Custom views don't have a mapp.user obj. This pull request checks for this and uses mapp.language which is in the custom view.

---
  - https://app.asana.com/0/0/1207220592858068


dbauszus-glx (approved):
lgtm

PR #1242, opened 2024-05-02
Add SRC env subtitution to the getLayer Module

simon-leech:
The replacement of keys `${HOST}` for example was previously only happening in the `getLocale` module. 
This PR adds it to the `getLayer` module too - which means that file paths within for example `layer.meta` can substitute ENV variables. 

The below would now resolve to whatever `SRC_HOST` is defined to in the ENV settings
``` json 
  "meta": "<a target='_blank' href='${HOST}/...`>Click here for User Guide</a></div>",
```

---
  - https://app.asana.com/0/0/1207220364404463

### DIFF
--- a/mod/workspace/getLayer.js
+++ b/mod/workspace/getLayer.js
@@ -68,10 +68,16 @@ module.exports = async (params) => {
     // Merge the workspace template into the layer.
     layer = merge(layer, template)
   }
+
   // Subtitutes ${*} with process.env.SRC_* key values.
   layer = JSON.parse(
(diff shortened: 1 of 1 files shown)


dbauszus-glx (commented):
Unmatched references should not be replaced with undefined.

```json
"meta": "<a target='_blank' href='${FOO}'>Click here for ${BAR} User Guide</a>",
```
${BAR} should remain in place if not defined as ENV.

image

dbauszus-glx (approved):

PR #1243, opened 2024-05-02
Update keyvalue_dictionary.mjs

KieronFerrey:
Fix for the custom view key value dictionary, but having an issue with the replaceKeyValue function as it doesn't assign the dictionary doesn't filter correctly 
const entry = dictionary.filter(entry => entry.key === key).find(entry => entry.value === value);


dbauszus-glx (changes requested):
mapp.language is never undefined.

in mapp.mjs line 74
```
language: hooks.current.language || 'en',
```

You can remove this check
```js
if (!mapp?.user?.language || !mapp.language) return;
```

dbauszus-glx:
This should now be working with the check removed in the keyvalue_dictionary method.

image

dbauszus-glx (approved):

PR #1246, opened 2024-05-07
Update clean-docs version

RobAndrewHurst:


dbauszus-glx (approved):
tested and works. i also bumped jsdocs itself.

RobAndrewHurst:
We actually don't need JSDoc as a dependency as it's part of the theme.

PR #1247, opened 2024-05-07
Gazetteer using searchbox element;

cityremade:
Gazetteer definition now includes searchbox element. 
This fixes a conflict between behaviour of search result select and select from a dropdown element.

### DIFF
--- a/lib/ui/Gazetteer.mjs
+++ b/lib/ui/Gazetteer.mjs
@@ -11,22 +11,12 @@ export default gazetteer => {
     }
   })
 
-  gazetteer.input = mapp.utils.html.node`<input
-    name="gazetteer-search-input" 
-    type="search" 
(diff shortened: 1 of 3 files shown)


RobAndrewHurst:
Also documented the ui/gazetteer module

simon-leech (commented):
I don't know if this should be in the same PR or not - but there appears to be no hover CSS for list elements - which I think it quite unclear for the user. 
Should we change this in this PR or open a different one for it?

RobAndrewHurst:
Oh I see what you mean

dbauszus-glx (approved):
Working well.

I added the CSS for webkit-search-cancel-button

simon-leech (approved):
All good with me and glad the CSS is resolved too :)

PR #1249, opened 2024-05-08
CSS Inputs - hover li background color replaced;

cityremade:
from:
background-color: var(--color-light);
to:
background-color: var(--color-primary-light);

<li> now highlighted on hover;


dbauszus-glx (approved):
lgtm

Issue #10, opened 2017-12-06
Enable additional styles for vector layers


dbauszus-glx:
We would like to add category feature styles to vector layers to style for example retail places type and pitch score areas.

cityremade:
That is categorised classification.

dbauszus-glx:
categorized and graduated themes are now enabled for MVT layers.

Issue #1228, opened 2024-04-25
dropdown element input filter


dbauszus-glx:
It should be possible to provide a filter flag to the dropdown element.

In this case the dropdown element should be a text filter input which will actively update the dropdown ul/li elements shown.

```js
mapp.ui.elements.dropdown({inputfilter: true})
```

With the input filter flag set the header > span should be an input element.

```html
<button class="dropdown active" data-id="dropdown">
      <div class="head">
      <span data-id="header-span">Retail - Logos<!--isµ2-->
      </span>
      <div class="icon"></div>
      </div>
      <ul style="max-height: 1066px; width: 310px;"><li>Retail - Logos<!--isµ1--></li><li>Icons<!--isµ1--></li><li>Dots<!--isµ1--></li><!--isµ3--></ul></button>
```

On input the render of the ul list items will be filtered.

dbauszus-glx:
For example the filter.dropdown may have a huge list of items.
![image](https://github.com/GEOLYTIX/xyz/assets/22201617/5471aa31-aada-4c15-9219-cfe4bc3f5dcc)

With the inputfilter it should be possible to reduce the list shown.
![image](https://github.com/GEOLYTIX/xyz/assets/22201617/42c7eee3-79d2-4285-abe8-011aaad9fe95)

dbauszus-glx:
Additionally the dropdown wants a parameter to control the max length shown.

PR #1250, opened 2024-05-08
ui.elements.dropdown: search input list filter

cityremade:
Related: https://github.com/GEOLYTIX/xyz/issues/1228

---
  - https://app.asana.com/0/0/1206524747200757

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -304,6 +304,20 @@ async function filter_in(layer, filter) {
     style="display: flex; gap: 10px; flex-flow: row wrap; padding: 10px;">`
 
     const selected_tags = new Set();
+
+    // creates an interactive search tag element
+    function createTag(val) {
+      return mapp.utils.html.node`<div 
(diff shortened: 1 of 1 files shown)


cityremade:
to apply on field:
```json
"filter": {
     "type": "in",
     "searchbox": true
 }```

cityremade:
Closes #10

dbauszus-glx (changes requested):
The lookup seems to check whether the letter is contained not whether the result begins with the search term.

Screenshot from 2024-05-13 10-15-25

image

Especially for the postcodes I would expect hundreds of results to be shown in the dropdown list.

Filter are just one example but the filter functionality should be added to the mapp.ui.elements.dropdown method to be available for any ui method.

layers filter should only contain an absolute minimum of ui logic.

cityremade:
@dbauszus-glx this is the logic of the search function which is a simple js-esque .includes(pattern). As we discussed before, we could make this available to a wider circle and ask what/how people would like to find matching results.

Issue #1212, opened 2024-04-18
Drawing interaction context menu


dbauszus-glx:
Add undo last vertex to drawing interaction context menu.

Indicate in the context menu that 'esc' is cancel and 'del' key are associated with the undo action.

PR #1251, opened 2024-05-10
Drawing Instructions Modal

simon-leech:
This PR addresses a common feature request I have been asked for and links to issue https://github.com/GEOLYTIX/xyz/issues/1212. 
I chose to address 1212 in this way as the user needs to know what options are available when they begin drawing, changing the context menu would only appear at the end of drawing interaction. 

Often users aren't clear exactly how to use the drawing interactions. 
This PR adds a modal element when drawing interactions are started that simply gives guidance on the drawing interaction options available. 
Hopefully this will just aid user experience a bit as it makes it clearer. 

All the text in the modal is using dictionary terms so can be translated. 
The modal is closable and will close when the drawing interaction finishes or the button state is inactive. 

image

---
  - https://app.asana.com/0/0/1207284671598597

### DIFF
--- a/lib/mapview/interactions/draw.mjs
+++ b/lib/mapview/interactions/draw.mjs
@@ -38,6 +38,8 @@ export default function(params){
     // Whether the draw interaction event should be handled.
     condition: e => {
 
+      if (mapview.interaction.wait) return false;
+
       // Right click
       if (e.originalEvent.buttons === 2) {
 
--- a/lib/ui/elements/_elements.mjs
+++ b/lib/ui/elements/_elements.mjs
@@ -15,6 +15,7 @@ import dropdown_multi from './dropdown_multi.mjs';
 import btnPanel from './btnPanel.mjs';
 import legendIcon from './legendIcon.mjs';
 import modal from './modal.mjs';
+import helpModal from './helpModal.mjs';
 import searchbox from './searchbox.mjs';
 import slider from './slider.mjs';
 import slider_ab from './slider_ab.mjs';
@@ -55,6 +56,7 @@ export default {
   dropdown_multi,
   legendIcon,
   modal,
+  helpModal,
   searchbox,
   slider,
   slider_ab,
--- a/lib/ui/elements/drawing.mjs
+++ b/lib/ui/elements/drawing.mjs
@@ -161,11 +161,11 @@ mapp.utils.merge(mapp.dictionaries, {
 })
 
 function deleteModal() {
-  if (document.querySelector("#Map > div.modal")) {
-    document.querySelector("#Map > div.modal").remove();
-  }
-
+  const modal = document.querySelector('[data-id=modal_drawing]')
+  
+  modal && modal.remove()
 }
+
 function GenerateModal(type) {
 
   // If a modal is already present, remove it
@@ -192,6 +192,7 @@ function GenerateModal(type) {
     <ul>${mapp.dictionary.draw_modal_save_single}</ul>
     </li>`
   }
+
   // Add the header
   const header = mapp.utils.html`<h3>${mapp.dictionary.draw_modal_title}</h3>`;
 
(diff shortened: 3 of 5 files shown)


simon-leech:
**I start poly, close modal. Click on point and then on poly again.**
_This should be resolved now._

**The modal method should be flagged to only show the modal when the flag is set.
Moving towards a future where it is possible to store config it can then be possible to remove the flag and not see the modal anymore if no longer desired.**

_I have added the `infoModal:true` flag that allows the developer and subsequently in the future the user to control the display of this information modal._

**The modal should be removed from within the modal method using the data_id.
Before a new data_id: 'modal_drawing' is created the existing one will be removed.**

**Does the modal method require an empty method for the close to work? This is hard to configure in json and should be looked at.**
_There was a bug in the `modal.mjs`_
_Previously the code was this, which meant passing just `true` to the close would result in an error._
``` js
  function closeModal(e) {
    e.target.closest('.modal').remove();
    modal?.close?.(e);
}
```
_I have updated the function to check on the close value being a function and only run it if so - removing the error._ 
``` js 
  function closeModal(e) {
    e.target.closest('.modal').remove();
 // Run the modal close function only if it exists
    if (modal?.close && typeof modal?.close === 'function') {
      modal.close(e);
    }
}
```
**Is it possible to provide the header as html template without the .node calling the web api?**
_Resolved this one._


I have also added conditional text so that the polygon type has a more verbose set of content than the rest

dbauszus-glx:
I did a few changes for the drawing interactions module but want to revise this more.

I renamed the method to help modal.

Whether the modal should be shown or not should be flagged in the locale. Config. This will allow us to create a plugin which allows the user to turn the modal off for all layers in the locale. And in turn allows the user to store this config in the user locale.

Whether the modal should be shown or not should be controlled within the modal method itself and not whether the modal method is called. Otherwise each call will require a condition check instead of a single condition check.
 
The deleteModal should use the data-id queryselector. No condition is required this can be always called as nothing will be deleted if there is nothing to delete.

```js
function deleteModal() {
  const modal = document.querySelector('[data-id=modal_drawing]')
  
  modal && modal.remove()
}
```

The classlist.toggle returns true on addition and can be used inside the if statement which returns after the interaction finish. No else, add, remove required.

The finish event calls the callback and removes the active class as well as removes the interaction.

Toggling the same interaction on, off, and [on again](https://www.youtube.com/watch?v=8OIkw9kJ0u4) will finish the current interaction and remove the class before starting the interaction. Hence it is required to add the 'active' class and modal after the interaction is triggered. 

This is more concise but can further be simplified as the same commands are run for each drawing interaction element.

```js
        const btn = e.target

        if (!btn.classList.toggle('active')) {

          layer.mapview.interaction.finish()
          return;
        }

        !layer.display && layer.show()

        layer.draw.point.callback = feature => {

          layer.draw.callback(feature, layer.draw.point)

          btn.classList.remove('active')

          delete layer.mapview.interaction

          deleteModal();

          // Set highlight interaction if no other interaction is current after 400ms.
          setTimeout(() => {
            !layer.mapview.interaction && layer.mapview.interactions.highlight()
          }, 400)
        }

        layer.mapview.interactions.draw(layer.draw.point)

        helpModal();

        btn.classList.add('active')
```

dbauszus-glx:
The drawOnclick method now has all the duplicate code from the individual drawing methods.

```js
function drawOnclick(e, layer, interaction) {

  const btn = e.target

  if (!btn.classList.toggle('active')) {

    layer.mapview.interaction.finish()
    return;
  }

  !layer.display && layer.show()

  interaction.callback = feature => {

    layer.draw.callback(feature, interaction)

    btn.classList.remove('active')

    delete layer.mapview.interaction

    helpModal();

    // Set highlight interaction if no other interaction is current after 400ms.
    setTimeout(() => {
      !layer.mapview.interaction && layer.mapview.interactions.highlight()
    }, 400)
  }

  layer.mapview.interactions.draw(interaction)

  helpModal(interaction.helpModal);

  btn.classList.add('active')
}
```

dbauszus-glx:
The helpModal is now a ui.element

```js
let helpModal

export default (params) => {

  // Only one helpModal.node can be shown at any one time.
  helpModal?.node && helpModal.node.remove()

  if (!params) return;

  helpModal = {
    target: document.getElementById('Map'),
    height: 'auto',
    width: '200px',
    css_style: 'padding: 0.5em;',
    top: 30,
    left: 60,
    contained: true,
    close: true,
    ...params
  }

  // Create the modal
  mapp.ui.elements.modal(helpModal);
}
```

dbauszus-glx (approved):

dbauszus-glx:
The drawOnclick method needs to be exported for ease of use in plugins.

dbauszus-glx:
The drawing interaction condition requires a wait check to not process inputs while waiting for the geometry function to resolve.

The wait condition will be removed in the API response callback.

AlexanderGeere (commented):
Looks good! I didn't know you could right click to clear a vertex 👍

RobAndrewHurst (approved):

PR #1252, opened 2024-05-10
infoj_order mode

dbauszus-glx:
The infoj method has a second infoj_order argument. This argument can be an array of keys and entry objects.

layer.infoj entries will be filtered in the order of keys. It is checked whether the key matches the entry's key, field, query, or group key value.

The `key` key value is specifically used as an identifier for entries and serves no other purpose.

The infoj_order argument can be a mix of entry objects and keys. Entry object are directly spliced into the infoj array and don't require to be in the layer.infoj array.

The infoj_order can be defined as an array in the layer json. The infoj_order plugin will be retired in favour of this.

---
  - https://app.asana.com/0/0/1207275671216623

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -74,63 +74,7 @@ export default (location, infoj_order) => {
       entry.value = entry.default || entry.value
     }
 
-    if (entry.query) {
-
-      // Assign queryparams from layer, and locale.
-      entry.queryparams = Object.assign(
-        entry.queryparams || {},
-        entry.location.layer.queryparams || {},
-        entry.location.layer.mapview.locale.queryparams || {})
-
-      // Check whether query returns data.
-      if (entry.queryCheck || entry.run === true) {
-
-        // Stringify paramString from object.
-        const paramString = mapp.utils.paramString(mapp.utils.queryParams(entry))
-
-        // Delete run 
-        delete entry.run;
-
-        // Add flag to outline it has already been ran
-        entry.hasRan = true;
-
-        // Run the entry query.
-        mapp.utils
-          .xhr(`${entry.host || entry.location?.layer?.mapview?.host || mapp.host}/api/query?${paramString}`)
-          .then(response => {
-
-            if (response) {
-
-              // Assign query response as entry value.
-              entry.value = entry.field ? response[entry.field] : response;
-
-            } else {
-
-              entry.value = entry.nullValue || null;
-            }
-
-            // Check whether entry should be skipped.
-            if (entrySkip(entry)) {
-
-              // Remove the entry.node from location view.
-              entry.node.remove();
-              return;
-            }
-
-            // Create element to be appended into empty entry.node
-            const el = mapp.ui.locations.entries[entry.type]?.(entry)
-
-            el && entry.node.appendChild(el)
-          })
-
-        continue;
-
-      } else if (entry.field && !entry.hasRan) {
-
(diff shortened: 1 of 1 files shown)


lib/ui/locations/infoj.mjs
```diff
@@ -146,6 +146,12 @@ export default (location, infoj_order) => {
         // Stringify paramString from object.
         const paramString = mapp.utils.paramString(mapp.utils.queryParams(entry))
 
+        // Delete run 
+        delete entry.run;
+
+        // Add flag to outline it has already been ran
+        entry.hasRan = true;
```
RobAndrewHurst: ++

cityremade:
when including a dataview object in the order I see the following error: 
infoj.mjs:130 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'layer')
at line 130 in infoj.mjs: `entry.location.layer.queryparams || {}`
Entry in the order array doesn't inherit properties when they are referenced.

lib/ui/locations/infoj.mjs
```diff
@@ -114,7 +114,7 @@ function entryDefault(entry) {
 
 function entryTitle(entry) {
 
-  if (entry.title) return;
+  if (!entry.title) return;
```
RobAndrewHurst, line 117: Noticed an issue here where if the entry had a title it would return. 

Was just missing the operator.

KieronFerrey (changes requested):
Tested with @simon-leech - so throwing feedback here. 

1. Can we add `entry.type` to the options - as this enables plugins that add elements to the infoj to be ordered.
2. `entry.group` is not working correctly - this is possibly due to the fact that we aren't looping over the infoj and looking for every option that has that group - so it only returns one entry in each group. 
3. dataviews in the infoj don't seem to be ordering

dbauszus-glx:
@KieronFerrey group won't work since the values are not unique. tyoe won't work neither for the same reason. we need to remove the group key since sorting requires a unique identifier.

simon-leech:
I just pushed a commit to this PR after chatting to @RobAndrewHurst. 
This PR simply console warns the developer if a value provided in `infoj_order` array does not exist in the `location.infoj`. 
This is required to ensure that the developer added a unique identifier to plugin entries or other entries that may not be unique.

AlexanderGeere (approved):
Looks good to me, Can order dataviews as well!

dbauszus-glx:
I have removed group from the infoj_order lookup since this is by definition not unique.

simon-leech:
I just pushed a check on the mapview and locale - as is custom views called entries may not have a locale
``` js 
// Assign queryparams from layer, and locale.
  entry.queryparams = Object.assign(
    entry.queryparams || {},
    entry.location.layer.queryparams || {},
    entry.location.layer?.mapview?.locale?.queryparams || {})
```

dbauszus-glx:
@simon-leech Can you test whether this works using the spread operator?

image

Aren't the locale queryparams already assigned to the layer somewhere else?

simon-leech:
@dbauszus-glx  Yeah all seems to be working for me :)

simon-leech (approved):

KieronFerrey (approved):
Looks good to me

RobAndrewHurst (approved):
Lets go!

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#infoj_order

PR #1253, opened 2024-05-13
Updating data within a json type field.

AlexanderGeere:
### What
Query for updating the values within a `json` field.

### How
Uses concatenation within sql to add the existing json field. 

### Why
We willl need this for the entering data directly from the report on a table.

### Test space
`bugs_testing/ui_elements` - scratch layer. 
- Alter the data in the edit fields and then click off and you should see the change reflected in the neondb in `public.scratch`, the `json_field` value.


simon-leech (dismissed):
This looks all good to me - the query looks fine :)

dbauszus-glx (changes requested):
I love the idea to have editable entries for JSON key values.

However having a query template to update the JSON feels awkward.

We already have a type:json entry which is editable allows for an update of the json object in SQL .

```json
{
  "type":"json",
  "field":"json_field",
  "edit": true
}
```

This entry can be hidden but allows access to the json object value from other entries.

This will get around the problem to request the same object multiple times for each location.

image

Another problem is the type edit. We potentially want to use any kind of entry type edit input to store the value in a json object.

This will become even more important in the future when we may decide to support other object stores for locations than SQL.

I have added a json_field type to this PR which looks up the json_field value from a json entry.

Alternatively you can provide the value directly using a fieldfx like so:

```json
          {
            "title": "Name",
            "type":"json_field",
            "jsonField":"json_field",
            "key":"name",
            "edit": true,
            "inline": true
          },
          {
            "title": "Surname",
            "type": "json_field",
            "jsonField":"json_field",
            "key": "surname",
            "edit": true,
            "inline":true
          },
          {
            "title": "Age",
            "type": "json_field",
            "field": "age",
            "fieldfx": "json_field -> 'age'",
            "jsonField":"json_field",
            "key": "age",
            "edit": true,
            "inline":true
          }
```

Can you add the edit logic to this entry to allow for the test of the update query without the need of a plugin?

dbauszus-glx:
The entryObject method can be simplified by assigning a const field and then checking and shortcircuiting on field, fieldEntry, fieldEntry.value not being an object.

@simon-leech can you give me an example where the objectMergeFromEntry requires a lookup on type? This should be removed since the type is by definition not unique like the field.

```js
function entryObject(entry) {

  const field = entry.objectAssignFromField || entry.objectMergeFromField || entry.json_field

  // entry must have a lookup field
  if (!field) return;

  const fieldEntry = entry.location.infoj.find(_entry => _entry.field === field)

  // info must contain a matching fieldEntry
  if (!fieldEntry) return;

  // fieldEntry must have an object type value.
  if (fieldEntry.value !== 'object') return;

  // Lookup for json value field entry
  if(entry.json_field) {

    if(!entry.key){
      console.warn('json_field requires entry.key to be specified')
      return;
    }

    entry.value = fieldEntry.value[entry.key]
  }

  if (entry.objectAssignFromField) {

    Object.assign(entry, fieldEntry.value)
  }

  if (entry.objectMergeFromField) {

    mapp.utils.merge(entry, fieldEntry.value)

    // // Find the entry to merge.
    // let fieldEntry = entry.location.infoj.find(_entry => _entry.type === entry.objectMergeFromEntry)

    // // Only merge the entry.merge object.
    // fieldEntry
    // && fieldEntry.merge instanceof Object
    // && mapp.utils.merge(entry, fieldEntry.merge)
  }
}
```

dbauszus-glx:
I changed `key` to `json_key` as this is better descriptive.

The update json query and multiple promises are not required.

At the beginning of the location update you can iterate through the json_field, json_key, newValues and assign the value to the json fieldEntry.newValue object.

```js
  this.infoj
    .filter(entry => entry.json_field)
    .filter(entry => entry.json_key)
    .filter(entry => typeof entry.newValue !== 'undefined')
    .forEach(entry => {

      const fieldEntry = this.infoj.find(_entry => _entry.field === entry.json_field)

      fieldEntry.newValue ??= fieldEntry.value || {}

      fieldEntry.newValue[entry.json_key] = entry.newValue

      delete entry.newValue
    })
```

[Screencast from 16-05-24 19:48:45.webm](https://github.com/GEOLYTIX/xyz/assets/22201617/41f3d6c8-2243-4063-9ebf-bf94161050ac)

dbauszus-glx (approved):

PR #1255, opened 2024-05-15
Infoj Order Issue

simon-leech:
If `layer.infoj_order` is undefined, previously custom views calling `mapp.ui.locations.infoj` would crash with this error. 
image

This PR simply adds optional chaining operators to prevent this error. 

---
  - https://app.asana.com/0/0/1207324780035008

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -28,7 +28,7 @@ export default (location, infoj_order) => {
   groups = {}
 
   // The infoj_order may be assigned to the layer.
-  infoj_order ??= location.layer.infoj_order
+  infoj_order ??= location?.layer?.infoj_order
 
   // infoj argument is provided as an array of strings to filter the location infoj entries.
   const infoj = Array.isArray(infoj_order) ?


dbauszus-glx (approved):
well spot. fix looks alright.

lib/ui/locations/infoj.mjs
```diff
@@ -28,7 +28,7 @@ export default (location, infoj_order) => {
   groups = {}
 
   // The infoj_order may be assigned to the layer.
-  infoj_order ??= location.layer.infoj_order
+  infoj_order ??= location?.layer?.infoj_order
```
RobAndrewHurst, line 31: Shouldn't a location always have a layer??

dbauszus-glx: Yes, it should. But better safe than sorry.

RobAndrewHurst: Alrighty!

simon-leech: I wasn't 100% sure on this - I thought this was the case and should always have a layer and thought better to be safe!

dbauszus-glx: Gazetteer locations may not have a layer I think.

RobAndrewHurst (approved):
Looks all good to me!

PR #1261, opened 2024-05-16
add input.invalid class

dbauszus-glx:
Instead of setting the min and max value when out of range the input event will add the invalid class and return.

The invalid class is opacity 0.5

[Screencast from 16-05-24 12:33:20.webm](https://github.com/GEOLYTIX/xyz/assets/22201617/4c5f262a-61c5-4cf9-8620-cf112385fe02)

### DIFF
--- a/lib/ui/elements/slider.mjs
+++ b/lib/ui/elements/slider.mjs
@@ -33,15 +33,20 @@ export default params => {
 
     let val = parseFloat(e.target.value)
 
-    if (isNaN(val) || val < params.min) val = params.min
-
-    const group = e.target.closest('.input-range')
+    if (isNaN(val) || val < params.min) {
+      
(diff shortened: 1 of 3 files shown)


cityremade:
Greying out while typing is not the most graceful behaviour. I would prefer something more comprehensive like adding a warning:
Screenshot 2024-05-20 at 13 17 36

simon-leech:
I agree with @cityremade  - I think we need to tell the user why its been disabled too with a message.

dbauszus-glx:
@cityremade I create a ticket for an element utility to show whether an input is invalid. I don't want to add this in this PR which is a bug that makes input with min values impossible.

simon-leech (approved):
Functionality is working for me (it was the other day but happy to review UI in a separate PR and just get this fix in!)

PR #1262, opened 2024-05-16
mvt_clone add layer

dbauszus-glx:
Adding `entry.L` layer to the mapview.map needs to be try...catched to prevent error on adding the same layer multiple times.

### DIFF
--- a/lib/ui/locations/entries/mvt_clone.mjs
+++ b/lib/ui/locations/entries/mvt_clone.mjs
@@ -93,6 +93,14 @@ export default entry => {
       entry.view = entry.panel
     }
     
+    // Add the clone layer to the mapview.Map.
+    try {
+      entry.mapview.Map.addLayer(entry.L)
+
+    } catch {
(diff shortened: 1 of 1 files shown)


lib/ui/locations/entries/mvt_clone.mjs
```diff
@@ -94,7 +94,12 @@ export default entry => {
     }
     
     // Add the clone layer to the mapview.Map.
-    entry.mapview.Map.addLayer(entry.L)
+    try {
+      entry.mapview.Map.addLayer(entry.L)
+
+    } catch {
+      // Will catch assertation error when layer is already added.
```
RobAndrewHurst, line 101: Should we not log the error still?

dbauszus-glx: Not required since it is not an error on which you can act. We could do a hasLayer check instead before addLayer but that would be more code and no performance gains.

RobAndrewHurst (approved):

Issue #1259, opened 2024-05-16
array input element [mapp.ui.elements.arrayInput()]


dbauszus-glx:
We like to create an input element for arrays. It must be possible to parse the input.

On input (eg. return key, or confirm button inside the input) the input value will be parsed and added to an array.

The array values should be displayed as 'pill' elements below the input. It must be possible to delete individual pill elements.

![image](https://github.com/GEOLYTIX/xyz/assets/22201617/eaaa2234-e5d6-4c79-a66b-d4ac5f4b29ae)

mapp.ui.elements.arrayInput

dbauszus-glx:
I've created this drag order for a pill control before. https://codepen.io/dbauszus-glx/pen/yLKVZLY

PR #1263, opened 2024-05-20
Pill component

cityremade:
Pill component adds a container of selected pills such as filtered values. 
It adds and allows removal of selected pills.
Values passed can create an initial set of pills.
The object has 2 functions: addCallback and removeCallback with both taking value and array of selected values as arguments.

Related to https://github.com/GEOLYTIX/xyz/issues/1259 but addresses the issue only in part.

### DIFF
--- a/lib/ui/elements/_elements.mjs
+++ b/lib/ui/elements/_elements.mjs
@@ -1,9 +1,9 @@
 /**
- ### mapp.ui.elements{}
+### mapp.ui.elements{}
 
- Module to export all the ui element functions used in mapp.
- * @module /ui/elements
- */
+Module to export all the ui element functions used in mapp.
+@module /ui/elements
+*/
 
 import card from './card.mjs';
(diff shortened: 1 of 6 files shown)


dbauszus-glx (approved):
The component.create method was only called from the component.add method and should be merged into one method.

A class `.pill` should be used for the pill divs in the container.

The empty target div is not required. If the target is instanceof htmlelement then the container can be added to the target. Otherwise the container can be manually added from the decorated object.

The close button should only be appended if `removeCallback: true`. Otherwise it would not be possible to create a pill component just showing the elements.

I added `type:pills` entry in order to test the pillComponent element.

```json
{
  "type": "pills",
  "default": [1,2,3,4,5,6,7,"A","B","C"],
  "css_pill": "background-color: red; border-radius: 25%; padding: 0 0.5em;"
},
```

It should be possible to configure the css_pill assigned as inline.

image

PR #1264, opened 2024-05-20
Review _user API mod

dbauszus-glx:
The _user API methods should each individually check whether the ACL is provided and user admin rights are required.

The admin view method was pushed into it's own mod file `user/admin.js`

The `user/delete` API will now allow for a logged in user to remove themselves.


AlexanderGeere (approved):
Looks good to me

RobAndrewHurst (approved):

AlexanderGeere (approved):

RobAndrewHurst:
I am happy for this to be reviewed. @simon-leech Do you still want to write some tests for this? Or should I try give it a crack?

simon-leech:
@RobAndrewHurst  Happy for you to pick up the tests then I can review the PR and tests all in one go?

RobAndrewHurst:
Sure!

RobAndrewHurst:
> @RobAndrewHurst Happy for you to pick up the tests then I can review the PR and tests all in one go?

The main problem with testing these modules is that we can't test them via the CLI. We would want to write a login_test similar to the test_view to the run on deployed instances. This will then fully test the ACL. 

I don't see a point in mocking the ACL in the CLI. 

Let me know what you think?  

@AlexanderGeere @dbauszus-glx @simon-leech

simon-leech:
> > @RobAndrewHurst Happy for you to pick up the tests then I can review the PR and tests all in one go?
> 
> The main problem with testing these modules is that we can't test them via the CLI. We would want to write a login_test similar to the test_view to the run on deployed instances. This will then fully test the ACL.
> 
> I don't see a point in mocking the ACL in the CLI.
> 
> Let me know what you think?
> 
> @AlexanderGeere @dbauszus-glx @simon-leech



> > @RobAndrewHurst Happy for you to pick up the tests then I can review the PR and tests all in one go?
> 
> The main problem with testing these modules is that we can't test them via the CLI. We would want to write a login_test similar to the test_view to the run on deployed instances. This will then fully test the ACL.
> 
> I don't see a point in mocking the ACL in the CLI.
> 
> Let me know what you think?
> 
> @AlexanderGeere @dbauszus-glx @simon-leech

@RobAndrewHurst Agreed - sounds quite tricky to mock, maybe we can add a login_test view as a separate PR? Once this one is in?

dbauszus-glx:
Added 

/docs/module-_user.html

/docs/module-_utils_reqHost.html

RobAndrewHurst:
> @RobAndrewHurst Agreed - sounds quite tricky to mock, maybe we can add a login_test view as a separate PR? Once this one is in?

Sounds good!

dbauszus-glx:
Added /docs/module-_user_acl.html and /docs/module-_user_add.html

dbauszus-glx:
The auth module should use async verification of the token.

dbauszus-glx:
The login module should use the cookie module to assign a cookie.

The cookie module should use the fromACL module to get the user.

At the moment it is possible to recreate the cookie unless the user is blocked which is duplicate code.

dbauszus-glx:
The update method as called from the admin panel will now do the verification bits.

```js
  let verification_by_admin = ''
  if (req.params.field === 'verified' && req.params.value === true) {

    verification_by_admin = `
      , password = password_reset
      , password_reset = NULL
      , failedattempts = 0
      , verificationtoken = NULL
      , approved = true
      , approved_by = '${req.params.user.email}|${ISODate}'
    `
  }

  // Get user to update from ACL.
  const rows = await acl(`
    UPDATE acl_schema.acl_table
    SET
      ${req.params.field} = $2
      ${verification_by_admin}
      ${approved_by}
    WHERE lower(email) = lower($1);`,
    [email, req.params.value])

  if (rows instanceof Error) {
    return res.status(500).send('Failed to access ACL.')
  }
```

The verification can be called directly by an admin.

```
http://localhost:3000/api/user/update?email=dennis@mail.com&field=verified&value=true
```

RobAndrewHurst (changes requested):
Can we force the users verification flag be set to false when they reset their passwords? So that when an admin user goes to reset their password via the verification flag they don't have to toggle it from true to false to true.

dbauszus-glx:
@RobAndrewHurst No, we cannot. Otherwise I could set your verification to false knowing your email address.

RobAndrewHurst:
We then need a way for Admin's to easily see who needs to be verified again

dbauszus-glx:
I added a column re_verification which is false if there user [record] has a verification token. If clicked on this the the verified true update query will be sent and the verified and approved columns will be toggled after successful update query.

RobAndrewHurst:
image

Resetting a users email (roberthurstsa@gmail.com) I don't see a green tick in the outstanding password reset column.

But after re-verifying the user through the admin panel, I then see a green tick for the user. Seems as if the logic needs to be reversed?

image
^^ This is after re-verifying the user.

dbauszus-glx:
Green tick means that the user is fully verified. Happy for you to reverse the logic.

RobAndrewHurst:
> Green tick means that the user is fully verified. Happy for you to reverse the logic.

Ah ok! If you put it like that I am happy with it!

RobAndrewHurst:
image

Should we change the message returned from this to maybe "Your session has been terminated. Please login again."

RobAndrewHurst:
image
 Maybe something like this.

RobAndrewHurst (changes requested):
image

On vanilla email templates we don't see the host of the email that is being sent. In this case it should be `localhost:300`. I set the `ALIAS` in the .env and I don't see it coming through.

dbauszus-glx:
@RobAndrewHurst The issue with missing host should be resolved in last commit.

The register request get's short circuited in the api module since it doesn't have user auth by definition.

Hence the reqHost util is required in the register module itself.

RobAndrewHurst (approved):
I have run through the ACL process. Looks all good!

simon-leech (commented):
I'm happy with this too, seems to be working from my testing!

simon-leech:
It doesn't seem to want me to mark it as approved though idk why 😂

simon-leech (approved):

dbauszus-glx:
The register new user method will no longer create a SQL Select statement but provide a values array to the ACL mod.

```js
  const USER = {
    email: req.body.email,
    password: req.body.password,
    password_reset: req.body.password,
    language: req.body.language,
    verificationtoken: req.body.verificationtoken,
    access_log: [`${date}@${req.ips && req.ips.pop() || req.ip}`]
  }

  if (process.env.APPROVAL_EXPIRY) {
    INSERTS.push('expires_on')
    VALUES.push(expiry_date)
    USER['expires_on'] = expiry_date
  }

  // Create new user account
  const rows = await acl(`
    INSERT INTO acl_schema.acl_table (${Object.keys(USER).join(',')})
    VALUES (${Object.keys(USER).map((NULL,i) => `\$${i+1}`)})`,
    Object.values(USER))
```

The user admin script will use tabulator 6.2 imports from unpkg.

The user admin cell toggle script will apply the verified logic for verified column clicks as well as re-verification.

RobAndrewHurst (approved):
**Nice one** all works for me

PR #1265, opened 2024-05-20
mvt_clone - reduce instances where the layer has to be removed or added.

dbauszus-glx:
Only the show() / hide() methods should require the layer L to be added or removed from the mapview.


simon-leech (approved):
This seems to be working as expected for me and resolves the few bugs found

RobAndrewHurst (approved):

PR #1266, opened 2024-05-21
`mapp.ui.elements.pills` Documentation and Rename 👀

RobAndrewHurst:
## `mapp.ui.elements.pill` Documentation and Rename 👀 

- ✍️ Renamed `pillComponent.mjs` -> `pill.mjs` to fit in with the current naming conventions.

- ✍️  Corrected the `pill.mjs` documentation so that the module page will be produced from the generate process.
     - The `@module` tag and the `@function` tag need to be in separate comment blocks for JSDoc 📖


dbauszus-glx (approved):
Agree on naming but it should be pills as a pill [singular] is a class that makes up the pills element/entry.

PR #1271, opened 2024-05-21
Add query to count number of points within a polygon

simon-leech:
Adds a new template query for counting the number of locations that intersect a geometry. 

Example set up here - 
``` json 
"infoj": [
  {
      "title": "Points in the Polygon",
      "inline": true,
      "field": "count",
      "query": "st_count_intersects",
      "queryparams": {
        "id": true,
        "qID": true,
        "table": true,
        "geom": "geom_4326",
        "table_b": "schema.table_name",
        "geom_b": "geom_p_4326"
      },
      "run": true
    }
]

```

### DIFF
--- a/mod/workspace/templates/_queries.js
+++ b/mod/workspace/templates/_queries.js
@@ -53,10 +53,13 @@ module.exports = {
   },
   st_distance_ab: {
     template: require('./st_distance_ab'),
-  },  
+  },
   st_distance_ab_multiple: {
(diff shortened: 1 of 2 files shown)


dbauszus-glx (approved):
I am ok with this as long as this is renamed to st_intersects_count. 

It feels awful specific though.

Given that this can be achieved with a query defined in the workspace.

We certainly don't want to maintain hundreds of specific queries which will be distributed in every instance. Used or not.

PR #1277, opened 2024-05-22
utils.xhr onerror resolve, post body default.

dbauszus-glx:
The XHR module default method must resolve if the web API errs.

The onload event method will never be called if the XHR fails on a CORS exception.

'POST' should be assumed as XHR method if a post.body is provided.

Added module and function documentation.

```js
/**
## mapp.utils.xhr(params)

The default method of the xhr module returns a promise which resolves with an response from a XMLHttpRequest.

@module /utils/xhr

*/ /**

The params object/string for the xhr utility method is required.

The params are assumed to the request URL if provided as a string argument.

The request params and response are stored in a Map() if the cache flag is set in the params object argument.

The method is assumed to be 'POST' if a params.body is provided.

@function default
@param {Object} params The object containing the parameters.
@param {string} params.url The request URL.
@param {string} [params.method=GET] The request method.
@param {string} [params.responseType=json] The XHR responseType.
@param {Object} [params.requestHeader={'Content-Type': 'application/json'}] The XHR requestHeader.
@param {string} [params.body] A stringified request body for a 'POST' request.
@param {boolean} [params.resolveTarget] Whether the target instead of target.response should be resolved.
@param {boolean} [params.cache] Whether the response should be cached in a Map().
@returns {Promise} A promise that resolves with the XHR.
*/
```

image

### DIFF
--- a/lib/utils/xhr.mjs
+++ b/lib/utils/xhr.mjs
@@ -1,7 +1,30 @@
 /**
-## mapp.utils.xhr()
+## mapp.utils.xhr(params)
+
+The default method of the xhr module returns a promise which resolves with an response from a XMLHttpRequest.
 
 @module /utils/xhr
+
+*/ /**
+
+The params object/string for the xhr utility method is required.
+
+The params are assumed to the request URL if provided as a string argument.
+
+The request params and response are stored in a Map() if the cache flag is set in the params object argument.
+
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):
Nice one!

Good docs! Looks all good to me!

simon-leech (approved):

PR #1278, opened 2024-05-22
Remove nanoid dependency

dbauszus-glx:
The nanoid dependency isn't used anymore in the backend in favour of the `crypto.randomBytes()` method.

```js
user.session = crypto.randomBytes(10).toString('hex')
```


AlexanderGeere (approved):
lgtm

PR #1279, opened 2024-05-22
css: .pill class now prevents user select;

cityremade:
As in description.
Prevented user select on the .pill element.


simon-leech (changes requested):
Working well and can see text isn't selectable. 

However, should we have it so the cursor is a pointer on the pills, and changes to a hand when you hover over the remove button?

RobAndrewHurst (approved):

simon-leech (approved):

PR #1280, opened 2024-05-23
Pills - added remove method.

cityremade:
Separating remove method makes it possible to be accessed programatically - like deselect value in the list and have the pill removed;


dbauszus-glx (approved):
Much cleaner. Like it. 👍

PR #1281, opened 2024-05-23
Dropdown : Reverse based upon page position

KieronFerrey:
This PR addresses an issue where when a dropdown element is at the bottom of the available viewport space - you cannot read the ul elements in it and/or they overspill off the page. 

This PR calculates the available space for the dropdown, and if its less than 150px, the dropdown has the `.reverse` class added to it so it opens upwards, to remove the issue. 

---
  - https://app.asana.com/0/0/1207389827168975

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -71,28 +71,30 @@ export default (params) => {
       const viewport = document.body.getBoundingClientRect()
 
       // Calculate the difference between the bottom of the dropdown and the viewport height.
-
       const viewDiff = viewport.height - bounds.bottom;
 
-      // If the difference is less than 150px, reverse the dropdown and set the maxHeight to the difference.
-      if (viewDiff < 150) {
-        console.log('reverse');
-        e.target.parentElement.classList.add('dropdown-reverse')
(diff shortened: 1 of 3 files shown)


simon-leech:
@KieronFerrey  I pushed a minor change just now. 
I rebuilt the CSS using the commands found here - https://github.com/GEOLYTIX/xyz/blob/main/DEVELOPING.md#esbuild

Also, when the space is less than 150px, I set the max height of the element to 150px (as its opening upwards). Previously it was being set as the available space below, which meant you could have a very small dropdown opening upwards. 


I think this is all good from my point of view now :)

dbauszus-glx (changes requested):
This needs to wait until after the dropdown pills is merged since they are both making changes to the same module script.

dbauszus-glx:
#1284 has now be merged and needs to be merged into this PR for testing.

simon-leech:
I merged main and resolved merge issues.

simon-leech (approved):
Working as expected for me - when the space available is small the dropdown is reversed to open upwards

RobAndrewHurst (approved):
Nice one! looks all good to me @KieronFerrey

Issue #1268, opened 2024-05-21
layer type entry method to replace mvt_clone


dbauszus-glx:
The vector_layer entry and mvt_clone entry type should essentially work the same but are quite different at the moment.

The vector_layer doesn't support themes.

There is a lot of duplicate code which is hard to maintain.

There should be a core layer entry into which the mvt_clone type uses to add to the clone specific items.

The clone logic isn't really required since an mvt layer can be created from the format method in a similar fashion to the vector layer.

It should be possible to create (draw) and modify geometries on vector and mvt [clone] layer.

simon-leech:
Using an `mvt_clone` entry - the style panel is always expanded by default even when display is false, it should be hidden in this scenario. If you toggle it on and then off again the style panel disappears os its just an issue on initial load. 

Swapping the `mvt_clone` thematic / display labels does not work presently as it fails with the error - `Uncaught TypeError: e.reload is not a function`

dbauszus-glx:
The styleParser is required for mvt_clone and vector [feature] layers. These entry methods can not be async. The show method has to be async in the revised feature_layer entry method. This draft PR cannot proceeed into review until this issue has been resolved. https://github.com/GEOLYTIX/xyz/pull/1344

simon-leech:
You cannot toggle label elements on and off for mvt_clone or vector at the minute. 

If you don't provide a default style but do provide a theme in the entry, you hit a no default style warning - even though the underlying layer has one.

dbauszus-glx:
The mvt_clone logic is convoluted and must be removed. Adding the type:layer entry method and keeping the mvt_clone type would increase complexity and be a maintenance nightmare.

The mode of filtering features for an mvt_clone layer is a featureLookup object. Populated by a query the lookup feature properties are merged with the mvt geometry feature prior to styling the feature. Layer template themes are not usable for the featureLookup since the properties maybe constructed and not available in the layer data at rest. If a featureLookup is used the style should be defined in the entry. The mvt format will only query MVT geometries with ids if the layer has a featureLookup object.

The recommended mode for filtering mvt layer is a featureSet of IDs. A query with the value_only flag can return an array of IDs for the featureSet by nesting a select statement for the id field inside a `SELECT ARRAY ()` sql statement. Switching themes and selection can only work with a featureSet.

The default layer view is not ideally suited for display inside the location view. Only a customised style panel should be shown by default. https://github.com/GEOLYTIX/xyz/issues/1154

PR #1282, opened 2024-05-23
Request geometry from API

dbauszus-glx:
The aim of this PR is to allow the geometry entry to be extended by API calls.

Instead of copy and pasting most of the geometry entry script to create custom entry types for geometries which are requested from an API it should be possible to provide an entry.api method which will be called within the geometry entry show method if no value is available in the entry.show() method.

The entry.show() method must be async to allow for blocking api calls. The default is non-blocking.

It should be possible to provide edit controls to request geometries but also to use default and/or custom draw methods.

```json
{
  "label": "GLX Catchment",
  "field": "isoline_h3",
  "fieldfx": "ST_asGeoJson(isoline_h3)",
  "type": "isoline_glx",
  "edit": {
    "draw": {
      "polygon": true
    }
  },
  "params": {
    "url": "XXXX",
    "api_key": "XXXXX"
  },
  "zIndex": 99,
  "style": {
    "strokeColor": "#0064B2",
    "strokeWidth": 2
  }
}
```

It should also be possible to modify geometries received from an API.

Any styling must only be applied by the geometry entry method.

For now geometries will always be stored as 'geojson' which will be turned to the native postgis geometry format in the location_update template.

However it should be possible to request and display other Openlayers supported formats. The new `mapview.geometry()` method supports any OL format. The `mapview.geoJSON()` method will set the geojson format before calling the geometry method.

The geometry created by the mapview.geometry() method should not be stored on the params object to prevent having to delete this object if the value updates.

The location_get method would not return a response if all fields are null. This has been resolved by concatenating the _.layer.qID with the fields array.

The geojson.js query template was bugged. Where `IS NOT NULL` is provided with the geom if now viewport is set.

The vector_layer entry method has been extended to apply theme styles and support API methods as well. However this work is incomplete and would too far extend of this PR. I have created a ticket to eventually merge the vector_layer and mvt_clone layer to support API calls, allow for the modification of geometries, and allow to draw new geometries. https://github.com/GEOLYTIX/xyz/issues/1268

A render event was added to the location view element.

This allows for rendering the location view again without update.

```js
location.view.addEventListener('render', ()=>renderLocationView(location))

location.view.addEventListener('updateInfo', () => {

    // Hides the upload icon.
    location.view.querySelector('.btn-save').style.display = 'none'

    // Location has toggle editing.
    if (location.editToggle) {

      location.editToggle.classList.remove('on')

      // Remove edits from infoj entries.
      location.removeEdits()
    }

    // Refresh dataviews
    if (location.layer?.dataviews) {
      Object.values(location.layer.dataviews).forEach(dv => {
        if (dv.display === true) {
          dv.update();
        }
      });
    }

    location.renderLocationView()
  })

function renderLocationView(location = this) {

  // Remove location.viewEntries.
  location.viewEntries.remove()

  // Enables the location view node and child elements.
  location.view.classList.remove('disabled')
  
  // Recreate location.viewEntries.
  location.viewEntries = location.view.appendChild(mapp.ui.locations.infoj(location))
}
```


RobAndrewHurst (changes requested):
Menu is shown for point geometry before the geometry is returned from the response. 

Menu needs to appear only when the geometry from the request is returned!

dbauszus-glx:
The geometry edit method will no longer append the delete button if the field is the current layer geometry.

https://github.com/GEOLYTIX/xyz/pull/1282/commits/d2fc198c3ac3f96ccf20227835eb7ff80c684b85

dbauszus-glx:
> Menu is shown for point geometry before the geometry is returned from the response.
> 
> Menu needs to appear only when the geometry from the request is returned!

This issue is caused by starting the draw interaction without setting drawend: null.

RobAndrewHurst:
Nice thanks! @dbauszus-glx. Is there anything else that is outstanding on this?

dbauszus-glx:
> Nice thanks! @dbauszus-glx. Is there anything else that is outstanding on this?

I am not aware of issues.

RobAndrewHurst (approved):

PR #1283, opened 2024-05-24
Key is ID = Throw Error

simon-leech:
In the `location_update` and `location_new` queries, it is possible to update the `id` field on the table on the db. 
However, doing so will actually overwrite the `%{id}` reserved parameter.
Example using location_update
``` sql 
UPDATE ${_.table} SET ${fields.join()} WHERE ${_.layer.qID} = %{id};
```

You are on location unique_id = 777. 
You want to update id field to be 123. 
The query you would expect to look like
``` sql
UPDATE schema.table SET id = 123 WHERE unique_id = 777`
```
It actually does this 
``` sql
UPDATE  schema.table SET id = 123 WHERE unique_id = 123`
```

**Proposed Fix**
- Simply don't allow the id field to be updated. 

**Long-Term Fix**
- We should replace %{id} with a better unique descriptor, such as %{infoj_id} or similar that wouldn't be used a field name on the db

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -93,7 +93,7 @@ async function update() {
 
   if (!Object.keys(newValues).length) return;
 
-  await mapp.utils.xhr({
+  const location_update = await mapp.utils.xhr({
     method: 'POST',
     url:
       `${this.layer.mapview.host}/api/query?` +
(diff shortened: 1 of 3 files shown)


dbauszus-glx (approved):
The check for the key should be against the _.layer.qID which may not be 'ID'.

This should be prior to the body iteration and throw an error.

```js
  // The location ID must not be altered.
  if (Object.keys(_.body).some(key => key === _.layer.qID)) {

    throw new Error('You cannot update the id field as it is a reserved parameter.')
  }
```

The location.update function should alert, enable the view without changes and return.

```js
  if (location_update instanceof Error) {
    alert(`Location update has failed.`)
    this.view?.classList.remove('disabled')
    return;
  }
```

RobAndrewHurst (approved):

simon-leech:
I just pushed a commit to make sure the Error message throws the Error using the layer.qID not 'id' so it is explicit what the issue is.

PR #1284, opened 2024-05-24
dropdown_pills added to elements;

cityremade:
Now filter[in] supports multi select with values as pills; Config inside in-filter: "dropdown_pills": true


dbauszus-glx:
The duplicate code issue has been resolved in https://github.com/GEOLYTIX/xyz/pull/1284/commits/20d3b48473aa9ba3e620bc9adb7826bebac38f09

Inside the filter.mjs I add pills to be dropdown_pills, ie. true || false/undefined

```js
  if (filter.dropdown || filter.dropdown_pills) {

    const dropdown = mapp.ui.elements.dropdown({
      pills: filter.dropdown_pills,
      multi: true,
      placeholder: 'Select Multiple',
      entries: filter[filter.type].map(val => ({
        title: val,
        option: val,
        selected: chkSet.has(val)
      })),
      callback: async (e, options) => {

        if (!options.length) {

          // Remove empty array filter.
          delete layer.filter.current[filter.field]
        } else {

          // Set filter values array from options.
          Object.assign(layer.filter.current, {
            [filter.field]: {
              [filter.type]: options
            }
          })
        }

        applyFilter(layer)
      }
    })

    return mapp.utils.html.node`<div class="filter">${dropdown}`
  }
```

In the dropdown.mjs I add the pills element to params.pills if true.
```js
  params.pills &&= mapp.ui.elements.pills({
    pills: [...params.selectedTitles],
    addCallback: (val, _pills) => {
      params.callback?.(null, [..._pills]);
    },
    removeCallback: (val, _pills) => {

      // deselect value in the list
      const item = ul.find(li => li.title === val);
      item.classList.remove('selected');
      params.selectedTitles.delete(item.title);
      params.selectedOptions.delete(item.dataset.option);
      params.callback?.(null, [..._pills]);
    }
  });
```

A dropdown pills by definition will always be a dropdown multi and I made the mistake of introducing the dropdown_multi element before thinking this through which leaves us now with this technical debt module.
```js
export default (params) => {

  console.warn('mapp.ui.elements.dropdown should be used with the multi flag')

  params.multi = true

  return mapp.ui.elements.dropdown(params)
}
```

dbauszus-glx:
I've added a maxHeight param for dropdowns.

dbauszus-glx (approved):
this works as expected for me.

AlexanderGeere:
Should these work when an `in` filter is applied to an integer field? 

I put an example in [`bugs_testing/filter/filter.json`](https://github.com/GEOLYTIX/xyz_resources/blob/bugs-testing/bugs_testing/filter/filter.json) 

The pills dont get removed when you click the `x`.

dbauszus-glx:
@AlexanderGeere this should now be fixed by adding and checking on the data-value attribute instead of the title.

RobAndrewHurst (approved):

simon-leech (approved):

PR #1286, opened 2024-05-28
Unsaved Changes CSS

simon-leech:
The `val-changed` class was being applied to edited infoj entries, but no CSS was being applied. 
This PR just adds a background-colour CSS to the `val-changed input` so the user is clear which information has been edited and not saved.
Added color-changed variable 
image

Asana
---
  - https://app.asana.com/0/0/1207402520812811

### DIFF
--- a/public/css/_colours.css
+++ b/public/css/_colours.css
@@ -17,5 +17,7 @@
     --color-on: #E18335;
 
     --color-no: #A21309;
+    
+    --color-changed: #ffffa7;
 
 }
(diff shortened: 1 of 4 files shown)


dbauszus-glx (dismissed):
Only named colours should be used in mapp css. Not hardcoded colour values which will make it impossible to theme colours.

simon-leech:
@dbauszus-glx Good point - just pushed and Added color-changed variable

RobAndrewHurst (approved):
Looks all good to me! Thanks @simon-leech

PR #1291, opened 2024-05-31
Deep merge [clone] locale for storage in userIDB

dbauszus-glx:
The locale must be merged [deep cloned] into an empty locale object in order to scrub htmlelements which might be in the plugin objects.

The merge util must warn and not assign the htmlelement to the clone target.

The layer filter should use spread instead of object assign in decorator.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -117,9 +117,10 @@ export default async function decorate(layer) {
   };
 
   // Set layer filter.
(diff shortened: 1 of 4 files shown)


RobAndrewHurst (approved):
Nice one! Thanks @dbauszus-glx.

Good works with the docs as well!

simon-leech (approved):
All working for me :)

PR #1292, opened 2024-06-03
Ability to Test for the creation/ deletion of locations

simon-leech:
**What?**

- This PR extends the functionality of the existing test_view to test for the creation / deletion of locations. 
- This is a very useful feature as this allows us to test that any database triggers / functions are working correctly, and permissions on tables that are editable are set up correctly.

**How?**

- After we get the last location on a layer, we can then access its `infoj`. 
- Using the infoj, we have access to the pin entry and geometry entry types.
- If the layer uses `layer.draw.point` we pass the last location pin geometry to the `location_new` template to create a new location.
- If the layer uses a polygon draw method (`layer.draw.circle`) for example, we pass the last location geometry entry of the same field as the `layer.geom` to the `location_new` template to create a new location.
- We then `get` the new location to check we can open it in the location panel.
- If the layer has `layer.deleteLocation`, we then test deleting this location and removing it from the location panel. 
- Note - we only test deleting locations if the layer has a `layer.draw` method, as otherwise we could be deleting locations that didn't create in the testview - impacting client data. 
- Note - the infoj must therefore contain the `pin` and/or the `geometry` required for this testing. This can be `class: display-none` if the client does not want to see it. 

image

---
  - https://app.asana.com/0/0/1207373884014202

### DIFF
--- a/public/tests/layer.test.mjs
+++ b/public/tests/layer.test.mjs
@@ -76,8 +76,12 @@ export async function layerTest(mapview) {
 
                                     if (layer?.draw?.point) {
                                         geometry = pin.geometry;
-                                    } else {
+                                    } else if (layer?.draw?.polygon || layer?.draw?.line || layer?.draw?.rectangle || layer?.draw?.circle) {
                                         geometry = polygon.geometry;
+                                    } else {
+                                        // We don't want to test this layer as it doesn't have a core draw object method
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):
Looks all good to me!

AlexanderGeere (approved):
Looks good to me!

PR #1293, opened 2024-06-04
Text Entry - distinct query fix field fix

AlexanderGeere:
The `entry.edit.query` is currently being ignore as:

```js
template: entry.edit.query || entry.json_field ? 'distinct_values_json' : 'distinct_values'
```
(ui/locations/text.mjs/line87)

Will cause `distinct_values_json` to be used if either `entry.edit.query` is supplied or if `json_field` is supplied.

This PR will fix that behaviour.


simon-leech (approved):
Well spotted - lgtm !

dbauszus-glx:
Did you test this with a json_field entry as well?

PR #1295, opened 2024-06-04
Test Updates

simon-leech:
Updated the `sqlFilter` and `validateRequestParams` tests from `.js` files that weren't being ran to `.mjs` so they are correctly used now. 
Added a new `reqHost` test. 

---
  - https://app.asana.com/0/0/1207466965068723


RobAndrewHurst (approved):
Nice one Simo!

PR #1298, opened 2024-06-04
Update Administration Panel

simon-leech:
This PR adds the Test View to the Admin Panel, and adds some CSS to the buttons in the btn-row in the Admin Panel

image


RobAndrewHurst (commented):
I added in the download param by default! But it all looks good!

RobAndrewHurst (approved):

dbauszus-glx (changes requested):
The default view is empty without a /dir which is valid.

image

The links won't work neither.

dbauszus-glx (approved):

Issue #1296, opened 2024-06-04
Distinct filter on json_field


dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/1293#event-13035165361

This fix now makes the distinct filter fall over for a in filter on a json_field with json_key.

```js
{
  "type": "json",
  "field": "json_field",
  "edit": true
},
{
  "title": "Name",
  "json_field": "json_field",
  "json_key": "name",
  "edit": true,
  "filter": {
    "type": "in"
  },
  "inline": true
},
```

The distinct field query must work for jsonb and json_field.

PR #1299, opened 2024-06-05
JSON Entries - Editing distinct query fix

AlexanderGeere:
The incorrect query was being run for `json_field` entries.
Editing should work on `jsonb_field` and `json_field` entries using this PR. 

Can test with `bugs_testing\ui_elements\workspace.json`


AlexanderGeere:
@dbauszus-glx Using a json field as a layer filter won't be possible. Is that something we should try implement in this PR?

RobAndrewHurst:
**This PR relates to #1296**

dbauszus-glx (approved):
Works well with my test case for json_field and jsonb_field

```json
          {
            "type": "json",
            "field": "json_field",
            "edit": true
          },
          {
            "title": "Name",
            "json_field": "json_field",
            "json_key": "name",
            "edit": {
              "options": []
            },
            "filter": {
              "type": "in"
            },
            "inline": true
          },
          {
            "title": "Surname",
            "json_field": "json_field",
            "json_key": "surname",
            "edit": {
              "placeholder": "Select from list",
              "options": []
            },
            "inline": true
          },
```

I added a check on the filters list to prevent an entry without a field to crash attempting a distinct_values query

Issue #1301, opened 2024-06-05
Filter Response Min Failing when 0


simon-leech:
**tldr; 
`if(0) `
resolves to false.**

In `lib/ui/layers/filters.mjs` on line 151 we have this check 
``` js 
if (response[minmax]) {
    filter[minmax] = filter.type === 'integer' ? Math.round(response[minmax]) : parseFloat(response[minmax]);
}
``` 
What it was assumed this is doing was checking that the response from the query contained either "min" or "max". This was working fine for numbers more than 0. 
However this if statement resolved to `if(0) { }` when the minimum value is 0, which was subsequently not setting the filter.min property, which subsequently printed out the `no data for this field` message to the user incorrectly.

PR #1302, opened 2024-06-05
Filter Min = 0 Fix

simon-leech:
# Filter Min Fix

## Description
I have added a new filter_minmax query that returns an array of both min and max values. This removed some of the complexity in the generateMinMax function as you no longer need to pass in a string of "min" or "max".

I have added a new filter.minmax array that gets set to [filter.min, filter.max] initially. The benefit of this is that by extracting the query response from the filter.min and filter.max values, we separate providing a specific min or max in the configuration vs getting them from the database.

I have resolved the issue with clear and reload of the filter, it will now update the filter values using the query.
I have added a new message that displays when the min and max are identical.
image

## GitHub Issue

 - [#1301](https://github.com/GEOLYTIX/xyz/issues/1301)

## Type of Change

## How have you tested this? 
Ran this using a field that contained 0 as the minimum field value. 

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -148,7 +148,7 @@ async function generateMinMax(layer, filter, minmax) {
   // If the response contains response.min or response.max - we can assume the query was successful.
   // Have to check that response is not null as well
   response ??= {}
-  if (response[minmax]) {
+  if (!isNaN(response[minmax])) {
     filter[minmax] = filter.type === 'integer' ? Math.round(response[minmax]) : parseFloat(response[minmax]);
   }
 
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -217,8 +217,8 @@ async function filter_numeric(layer, filter) {
   // Create affix for rangeslider input label.
   const affix = (entry.prefix || entry.suffix)? `(${(entry.prefix || entry.suffix).trim()})`: ''
 
+  // Set formatter and filterInput for tabulator inputs.
   entry.formatter ??= entry.formatterParams
-
   entry.filterInput = true
 
   // Create the range slider element.
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -130,13 +130,19 @@ function filter_null(layer, filter) {
 }
 
 /**
- * @function generateMinMax
- * @description Generates the minimum and maximum values for numeric filtering.
- * @param {Object} layer - The layer object to apply the filter to.
- * @param {Object} filter - The filter configuration object.
- * @returns {Promise<void>}
- */
+@function generateMinMax
+@description
+Queries the min and max bounds for a filter field. The min and max bounds can be implicit in the filter config.
+
+@param {Object} layer MAPP layer object.
+@param {Object} filter Filter object.
+@param {string} filter.field Field to filter.
+@param {numeric} [filter.min] Min bounds.
+@param {numeric} [filter.max] Max bounds.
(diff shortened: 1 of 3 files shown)


dbauszus-glx (changes requested):
I really struggle to wrap my head around this. How can I test this.

But I also have issues with the integer filter.

If the layer has only one value and min max are the same then the filter should be disabled.

No input is possible.

image

The min max should be run again if I close and open the filter. I added a value to the example and had to reload the locale to see the filter changing.

```js
"field": "minutes",
"filter": {
  "type": "integer"
},
```

simon-leech:
I have added a new `filter_minmax` query that returns an array of both min and max values. This removed some of the complexity in the `generateMinMax` function as you no longer need to pass in a string of "min" or "max". 

I have added a new `filter.minmax` array that gets set to `[filter.min, filter.max] initially. The benefit of this is that by extracting the query response from the `filter.min` and `filter.max` values, we separate providing a specific min or max in the configuration vs getting them from the database. 

I have resolved the issue with clear and reload of the filter, it will now update the filter values using the query. 
I have added a new message that displays when the min and max are identical. 
image

lib/ui/layers/filters.mjs
```diff
@@ -133,25 +134,20 @@ function filter_null(layer, filter) {
...
     locale: layer.mapview.locale.key,
     layer: layer.key,
     table: layer.tableCurrent(),
     field: filter.field,
   })}`);
 
-  // If the response contains response.min or response.max - we can assume the query was successful.
-  // Have to check that response is not null as well
-  response ??= {}
-  if (response[minmax]) {
```
RobAndrewHurst, line 151: Nice one removing this! 

We must still take note that `if(0)` is falsy.

dbauszus-glx (approved):
It should be possible to define the step in config.

Added to docs.

```js
  // if (!filter.step) {

  //   filter.step = filter.type === 'integer' ? 1 : 0.01;
  // }

  filter.step ??= filter.type === 'integer' ? 1 : 0.01;
```

I've updated the docs for the generateMinMax to include field, min, max [optional]

Implicit check can be done by checking min, max not a number, ie undefined.

Promise<void>, ie no return doesn't require a return in description.

```js
/**
@function generateMinMax
@description
Queries the min and max bounds for a filter field. The min and max bounds can be implicit in the filter config.

@param {Object} layer MAPP layer object.
@param {Object} filter Filter object.
@param {string} filter.field Field to filter.
@param {numeric} [filter.min] Min bounds.
@param {numeric} [filter.max] Max bounds.
*/

async function generateMinMax(layer, filter) {

  const response = await mapp.utils.xhr(`${layer.mapview.host}/api/query?${mapp.utils.paramString({
    template: `field_minmax`,
    layer: layer.key,
    table: layer.tableCurrent(),
    field: filter.field,
  })}`);

  // Assign min, max from response if not a number.
  const min = isNaN(filter.min) ? response.minmax[0]: filter.min
  const max = isNaN(filter.max) ? response.minmax[1]: filter.max

  // Parse min, max as interger or float depending on type.
  // Round integer to be correct up or down.
  filter.minmax[0] = filter.type === 'integer'? Math.round(min) : parseFloat(min)
  filter.minmax[1] = filter.type === 'integer'? Math.round(max) : parseFloat(max)
}
```

lib/ui/layers/filters.mjs
dbauszus-glx: This should be resolved.

PR #1307, opened 2024-06-06
Update docs and allow for numeric filter value.

dbauszus-glx:
The `filter_in` function should work on an array of numeric values.

eg. 
```json
{
  "field": "minutes",
  "type": "integer",
  "filter": {
    "type": "in",
    "searchbox": true
}
```

The searchbox function doesn't require a defaultSearch method which was used as a method to append the no search results.

The searchbox function should warn if no searchFunction is provided.

I split the add() and remove() methods from the pills(component) method for clarity.

Updated the docs for 
@module /ui/layers/filters, @function filter_in

@module /ui/elements/searchbox, @function searchbox

@module /ui/elements/pills, @function pills, @function add, @function remove

I found this hard to understand.
```js
        filtered
        .filter(i => !pillComponent.pills.has(i)) // only those not selected already
        .map((i, j) => {
          if(j > 9) return; // show only 10 top matches
          searchbox.list.appendChild(mapp.utils.html.node`<li onclick=${e => {
            if(pillComponent.pills.has(i)) return;
            pillComponent.add(i);
          }}>${i}`);
```

I prefer to use named variables, where can be the index. Using i as value, and j as index was very hard to understand.
I prefer to use filter separately eg (i<9) opposed to a condition in an array map/forEach method. I prefer to use forEach if nothing is returned from the method which is the point of the map method.

```js
        filtered
          .filter(val => !pillComponent.pills.has(val)) // only those not selected already
          .filter((val, i) => i < 9)
          .forEach(val => {

            // Append li element to searchbox.list
            searchbox.list.append(mapp.utils.html.node`
              <li onclick=${() => {
                // Add pill not yet in pillComponent
                !pillComponent.pills.has(val) && pillComponent.add(val);
              }}>${val}`)
          });
```


AlexanderGeere (approved):
Looks Good to me!

PR #1310, opened 2024-06-07
ui.elements.input

cityremade:
Added an element for input of type 'number'.
Used in numeric entry as a test area.
```
{
                    "title": "numeric field",
                    "field": "numeric_field",
                    "type": "numeric",
                    "edit": {
                      "min": 0,
                      "max": 100,
                      "placeholder": "Number between 0 and 100"
                    }
                },
                {
                    "title": "integer field",
                    "field": "integer_field",
                    "type": "integer",
                    "edit": true
                }
```

Includes validation if min/max defined in parameters.


cityremade:
To clarify further:

- this element is of type number only. I am concerned with complexity if we put text and number inputs together. Should they be separate elements?

- Another case is numeric entry with formatting like separators and currency symbols - this is a mix of numeric argument (in an input of type text) and string display. Where does it go from here? Wasn't this special input meant for editing fields in tables? It takes a lot of space in `numeric.mjs`

- It is used in numeric entry. 

- Another use of numeric input is the slider element. Slight difference in use as slider/range would never bounce to null/undefined.

Let me know your thoughts.

dbauszus-glx (changes requested):
Review in progress

dbauszus-glx:
mapp.language is assigned as mapp.user.language and will always have a value.

hence `formatterParams.locale` will always have a value and doesn't need to be checked before creating the localeValue as const.

`undefined` is not a number hence only null and NaN need to be checked in the condition for the value.

dbauszus-glx (changes requested):
I've changed the name of the input module to numericInput to avoid confusion.

I added missing params to the docs block.

entry.edit will always be an object. set to be an object if true. hence the checks for the object members are not required.

entry.edit.min max is set to the postgres range to avoid a crash when trying to update integer outside the range.

I assigned the numericChecks function to oninput event to be called on key or input by copy paste value.

I was not able to work out the reason for the onblur bit. Is this needed? If so, can you add a comment why this is needed so I can test this.

```js
    onblur=${e => {
      if (e.target.value === '') e.target.classList.remove('invalid');
    }}
```

cityremade:
> 
> ```js
>     onblur=${e => {
>       if (e.target.value === '') e.target.classList.remove('invalid');
>     }}
> ```

When .invalid class is on it is toggled off when input is **emptied and on blur**. Otherwise the .invalid style would be left there on cleared and valid (null) input. One of edge cases.

dbauszus-glx (approved):
happy to merge this in and look at other issues separately.

PR #1313, opened 2024-06-11
Query Button PR

simon-leech:
This PR moves the query_button plugin into the core. 
The method returns a button element which will execute a parameterised query.
"reload" is used to reload the location.layer after the query has been executed (e.g. when the location geometry has been updated).
"dependents" is used to reload the dependent fields after the query has been executed.
"alert" is used to display a message after the query has been executed.

```json
{
  "label": "Snap to Postal Sector",
  "type": "query_button",
  "query": "catchment_statistics_snap_to_postal_sector",
  "queryparams": {
    "id": true
  },
  "alert": "Query has executed!",
  "reload": true,
  "dependents": [
    "geom_3857",
    "perimeter",
    "area"
  ]
}

### DIFF
--- a/lib/ui/locations/entries/_entries.mjs
+++ b/lib/ui/locations/entries/_entries.mjs
@@ -28,6 +28,7 @@ Exports a collection of entry methods for location views.
 - time
 - title
 - vector_layer
+- query_button
 
 @module /ui/locations/entries
 */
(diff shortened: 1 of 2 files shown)


dbauszus-glx (approved):
Happy for this to be merged. Just done quick review on the docs.

Optional params should be in square brackets not the other way around.

Also added to wiki. https://github.com/GEOLYTIX/xyz/wiki/Workspace-Configuration#query_button

PR #1323, opened 2024-06-14
Merge entry style into location style for geometry entry.

dbauszus-glx:
Geometry entry style maybe defined as a opacity only to be merged with the location style.


RobAndrewHurst:
Is there a workspace easily accessible for this PR? Or will I need to create one?

dbauszus-glx:
> Is there a workspace easily accessible for this PR? Or will I need to create one?

Any vector layer without a style config.

RobAndrewHurst:
Nice one! Works for me!

Tested it with `plugins/isolines.json` 🧪

RobAndrewHurst (approved):

PR #1325, opened 2024-06-14
Ensure that each entry has a key.

dbauszus-glx:
Each entry should be assigned a key in the infoj iteration if not already assigned.

The key can be an entry.field which is unique. Otherwise a running index can be assigned within the iteration.

The key should be used primarily as a data_id to ensure that similar elements such as checkboxes have a unique data-id within their location.view context.


cityremade:
could entry index within infoj[] be used as fallback key?

simon-leech (approved):

dbauszus-glx:
The problem with the index is that it may not be known. The key is known and may be used in a plugin to access a data_id element.

PR #1326, opened 2024-06-14
Ensure that the style.zIndex is only used in the olStyle module

dbauszus-glx:
The style.zIndex must not be confused with the layer.zIndex.

The style.zIndex defines the render order within a layer.canvas

The layer.index defines the order of layer canvases within the Openlayers mapview.

### DIFF
--- a/lib/layer/format/googleMapTiles.mjs
+++ b/lib/layer/format/googleMapTiles.mjs
@@ -4,6 +4,8 @@ export default async (layer) => {
 
   layer.style.mapType ??= 'roadmap'
 
+  layer.zIndex ??= 0
+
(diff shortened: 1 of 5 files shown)


RobAndrewHurst:
@simon-leech We will need to add in a condition to the auto-update to move the zIndex from the `layer.style.zIndex` to the `layer.zIndex`

simon-leech:
@RobAndrewHurst  Good point! 
PR for that fix is here https://github.com/GEOLYTIX/xyz-ignite/pull/56

simon-leech (approved):
This looks ok to me :)

RobAndrewHurst (approved):

PR #1332, opened 2024-06-18
Don't spread style when style object is empty

simon-leech:
# Pull Request Template

## Description

Assigning `entry.style` using spread assignment means that it is not possible to have an empty style object. This is valid as we often use an empty style object to generate third-party isolines for example, but not draw them on the map. 
For example, an entry configuration of this which used to not draw, will currently draw to the map which is incorrect.
``` json
{
      "field": "x",
      "fieldfx": "ST_AsGeoJSON(x)",
      "type": "third_party_plugin",
      "display": true,
      "style": {},
      "class": "display-none"
}
```

By checking on the length of `entry.style` this means that if the object is empty, we won't merge in the `entry.location.style` thus keeping the entry with no styling, so it won't be drawn to the map. 
``` js 
// Assign entry.style to location.style if entry.style is not an empty object.
  if (Object.keys(entry.style).length) {
    entry.style = { ...entry.location?.style, ...entry.style }
  }
```

## Type of Change

## How have you tested this? 
Tested locally on my machine.

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -100,8 +100,10 @@ export default function geometry(entry) {
   // Return if entry has no geometry value and cannot be drawn in to.
   if (!entry.value && !entry.draw && !entry.api) return;
 
-  // Assign entry.style to location.style
-  entry.style = { ...entry.location?.style, ...entry.style }
+  // Assign entry.style to location.style if entry.style is not an empty object.
+  if (Object.keys(entry.style).length) {
+    entry.style = { ...entry.location?.style, ...entry.style }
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
image

This shouldn't break with null or undefined style.

style: null should be possible and is indeed clearer to indicate that the geometry should not be styled.

should be resolved in my commit

RobAndrewHurst (approved):
Looks all good to me

PR #1333, opened 2024-06-18
Remove default dataviews imports; Allow for dynamic dataviews.

dbauszus-glx:
The dataview methods have been revised and are now fully documented.

The dataview chart and tabulator modules require plugins and have been removed from the ui.utils.

The dataview plugin specific configuration wiki should be moved to a wiki with the relevant plugins.

The dataview.create method is not required to be async with dataview plugins loading the required third party libraries.

The location and layer dataviews are essentially the same. A location must have a layer which is assigned as the dataview layer by the dataview location entry method.

The should not be duplicate code between the location and layer dataview methods.

The display and hide dataview methods are assigned in the dataview creation itself.

The dataview create method defines whether a dataview has already been created. The dataview create method can be used to re-create a dataview but this requires a change to the dataview plugin. The target element children must be replaced rather than appended.


AlexanderGeere (changes requested):
When opening a location on `bugs_testing/dataviews/tabulator_workspace` the tabview tables work but the location doesn't open with this error:

Screenshot 2024-06-27 134302

Works on `v4.9.1` and `main`.

dbauszus-glx:
@AlexanderGeere 
This should be resolved in https://github.com/GEOLYTIX/xyz/pull/1333/commits/5dd0cf46c119ef621b69b18d807192c6c7ca4d1b

The issue was that chatjs library wasn't loaded. The checkDataview method must assign and return the error.

The location entry and layer [panel] dataview methods must return if the dataview creation errs.

image

dbauszus-glx:
The dataview decorator method creates a typedef:dataview which is added to the global in the ui/Dataview module.

Within the mapp core code the mapp.ui.Dataview() method is only called from the dataview locations entry or the dataviews layer panel.

A dataview from a layer panel is assumed to be in tab[view].

A dataview show and hide method for tabview display are assigned before the Dataview is decorated.

The show method will create the tab when first called. The addTab event method will sortcircuit if a tab has already been created. The dataview.hide() method will call the tab.remove() method.

```js
    dataview.show ??= () => {

      // Create tab after dataview creation is complete.
      dataview.tabview.dispatchEvent(new CustomEvent('addTab', {
        detail: dataview
      }))

      // Show the dataview tab.
      dataview.show()
    }

    dataview.hide ??= () => {

      dataview.display = false

      dataview.remove()
    }

    if (mapp.ui.Dataview(dataview) instanceof Error) return;

    // Display dataview if layer and dv have display flag.
    layer.display
      && dataview.display
      && dataview.show()

    layer.showCallbacks.push(()=>{
      dataview.display && dataview.show()
    })
```
In the location entry method the tabview is created if defined before the dataview is created. This will assign the tab.show() and tab.hide() methods to the dataview object. The dataview is a tabview as well as a dataview. The dataview will be created/updated in the activate event for the panel. This is debounced to only create a dataview in an active panel and not n dataviews all at once if multiple tabs are added on the initial load.

```js
  // Find tabview element from data-id attribute.
  entry.tabview ??= typeof entry.target === 'string'
    && document.querySelector(`[data-id=${entry.target}]`)

  // Dataview will be rendered into a tabview panel.
  if (entry.tabview) {

    // Assign border style based on the location view record (from list)
    entry.tab_style ??= `border-bottom: 3px solid ${entry.location.style.strokeColor}`

    // Assign target html element for dataview.
    entry.target = mapp.utils.html.node`
      <div class="dataview-target">`

    // Create tab after dataview creation is complete.
    entry.tabview.dispatchEvent(new CustomEvent('addTab', {
      detail: entry
    }))

  } else {

    // Dataview will be rendered into location view.
    entry.locationViewTarget = mapp.utils.html.node`
      <div class="${`location ${entry.class}`}">`

    entry.target = entry.locationViewTarget
  }

  if (mapp.ui.Dataview(entry) instanceof Error) return;

  // Dataview should be displayed.
  entry.display && entry.show()
```

dbauszus-glx:
@AlexanderGeere The activate method is now assigned in the addTab as a function to be debounced rather than using the dispatch event.

AlexanderGeere (approved):
Looks good to me! I check the report on t`ui_elements/workspace.json` tables work there as well.

simon-leech (changes requested):
Layer.dataviews now always throw a warning even though they didn't previously.

image

Testing on a custom view throws these errors that are not there off this PR.
image
If the tabulator plugin is not included - you see loads of warnings and errors. Removing the tabulator plugin is a breaking change so this must be a point release if we want to do this.

dbauszus-glx:
@simon-leech 

I remove the warning if the dataview type isn't implicit but can be inferred from table or chart object entry.

I assign the key from query/title/label if not implicit to help identifying the dataview in error message.

I assign an update method with a warning to prevent the update query crashing the script execution.

simon-leech:
Toggling the display of a dataview from an infoj entry in the tabview works to add to the tabview. 
But then toggling the checkbox to false, the dataview is removed from display within the tabview tab, but the tab itself is not removed.

In custom views if the tabulator/ chartjs plugins are not loaded I see errors / warnings. This is a divergence from the existing set up so needs to be well explained / perhaps a point release as it is technically a breaking change. 
image

dbauszus-glx:
@dbauszus-glx the issue with tabs not hiding has been resolved in last commit. The hide [dataview] method must be overwritten with the remove tabview method.

simon-leech:
`queryCheck` does not work as it previously did. 
Previously if no data was returned the checkbox was disabled. 
Now it returns a No Data message - is this expected / what we want to do?

AlexanderGeere (dismissed):

dbauszus-glx:
We can alter the behaviour but I need to understand what the queryCheck flag should do?

The current behaviour for an entry with queryCheck is that the query will be executed and the data  stored as value.

Then a tabview will be created if the checkbox is clicked, and the dataview will be created. The query will be run again because the update query is run after the dataview is created.

dbauszus-glx:
@simon-leech I added the check to disable the dataview entry with `querycheck:true` and null response.

simon-leech (approved):
I've tested this across a range of use cases: 

- custom plugins
- tableviews in the location 
- tableviews in the tabview
- MapChange and Viewport tableviews
- Zoom restricted layers (tables turn off correctly)
- Checkbox tableview views
- queryCheck

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #1338, opened 2024-06-19
maplibre and style.zIndex

simon-leech:
# Pull Request Template

## Description

1. Added a warning on `layer.style.zIndex` deprecation and to use `layer.zIndex` instead in `styleParser`.
2. In PR https://github.com/GEOLYTIX/xyz/pull/1326 the `maplibre` `layer.zIndex` was uncommented. This was a mistake as it doesn't work correctly. I re-commented but added information about why it is commented out. 

## Type of Change

## How have you tested this? 
Tested locally.

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/layer/format/googleMapTiles.mjs
+++ b/lib/layer/format/googleMapTiles.mjs
@@ -4,18 +4,17 @@ export default async (layer) => {
 
   layer.style.mapType ??= 'roadmap'
 
-  layer.zIndex ??= 0
-
   layer.L = new ol.layer.WebGLTile({
+    key: layer.key,
+    className: `mapp-layer-${layer.key}`,
+    zIndex: layer.zIndex,
     source: new ol.source.Google({
       key: layer.apiKey,
       scale: 'scaleFactor2x',
       highDpi: true,
       mapType: layer.style.mapType,
       layerTypes: layer.style.layerTypes
-    }),
-    layer: layer,
-    zIndex: layer.zIndex
+    })
   })
 
 }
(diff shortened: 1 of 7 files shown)


dbauszus-glx (changes requested):
The maplibre layer must be able to work with z-index.

It is a valid configuration would be to have a maplibre mask layer (eg. transparent with buildings and roads) which we want to display on top of a vector layer.

Another example would be labels. You may have a maplibre layer with just labels which should be drawn on top.

Please test this sample layer https://studio.mapbox.com/styles/dbauszus/cjackgsl84dmh2ro26eaz8qf5

This has just building outlines.
image

simon-leech:
Moving back into the draft - the issue isn't with the zIndex assignment its when you turn on the layers that they appear in the wrong order.

dbauszus-glx (approved):
This works now for me but had to do a couple changes as this was a mess.

The zIndex is assigned in the addLayer if not implicit in config, the index in the layer array iteration to be added will be used.

Every layer should have key, zIndex, and className.

The classname must be assigned to the container if the container is returned from the layer render method.

The preserve drawing buffer option should be taken from the layer config.

The `layer: layer` must not be assigned to the layer creation.

simon-leech:
I can't review as it was my PR but all working for me correctly now :)

PR #1341, opened 2024-06-20
Queryselector fails with data-id beginning with zero

dbauszus-glx:
A valid construction is to create a geometry type entry to request geometries from an API.

Without a key and a field 0 may be assigned as key.

A data-id queryselect fails beginning with 0.

```
{
  "type": "isoline_mapbox",
  "params": {
    "access_token": "***"
  }
},
```


simon-leech (approved):
All good with me :)

PR #1343, opened 2024-06-21
Boolean entry  - unnecessary label after symbol

simon-leech:
## Description

The boolean entry has an extra label defined that is not required. As the checkbox is created with a label or title. 
This adds an ugly field or label to the end of the checkbox that is only removed by passing a  `entry.label = " "`.
image

This PR simply removes this from the end of the boolean entry. So that `entry.title` is used instead of appending this to the end. 

image

## Type of Change

## How have you tested this? 
Tested locally with an entry of this configuration 

``` json 
  {
      "title": "Flag",
      "field": "flag",
      "type": "boolean",
      "inline": true
    }
```
## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/entries/boolean.mjs
+++ b/lib/ui/locations/entries/boolean.mjs
@@ -33,22 +33,22 @@ export default function boolean(entry) {
 
     return mapp.ui.elements.chkbox({
       label: entry.label || entry.title,
-      checked: entry.newValue !== undefined ? entry.newValue: entry.value,
+      checked: entry.newValue !== undefined ? entry.newValue : entry.value,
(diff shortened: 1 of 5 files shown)


RobAndrewHurst (approved):
Nice one! Looks all good to me! 

I have added in a test in `public/tests/lib/ui/locations/entries/boolean.test.mjs` which will take an entry that is `inline` and a value of `false` which will result in a mask-icon of close for the div, that will be checked in the test

PR #1344, opened 2024-06-21
Mapview svg templates

dbauszus-glx:
The svg_templates should not be plugin but a util.

The utility method must be async to allow for the fetching of svg templates from external sources.

The mapview creation must be async to await svg_templates in a locale.

It should be possible to define svg_templates in a layer.style.

This requires the styleParser utility to be async. The svg_templates utility will be called from the styleParser method.

The styleParser method is required for feature layers (eg. vector / mvt). Feature layer format methods must be async.


dbauszus-glx:
The styleParser is required for mvt_clone and vector [feature] layers. These entry methods can not be async. The show method has to be async in the revised feature_layer entry method. This draft PR cannot proceeed into review until this issue has been resolved. https://github.com/GEOLYTIX/xyz/issues/1268

PR #1346, opened 2024-06-24
Geometry - Configurable Labels

simon-leech:
# Pull Request Template

## Description

1. When populating a geometry field using an API, we need to then run `update()` so that any `dependents` are also updated. 
2. Added `entry.edit.modify_label` and `entry.edit.delete_label` that can be configured. 

## Type of Change

## How have you tested this? 
Tested locally.

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -187,6 +187,9 @@ async function show() {
   if (!this.value && this.api) {
 
     await this.api(this)
+
+    // Update dependents and location view.
+    update(this)
   }
 
   if (this.L) {
@@ -234,13 +237,13 @@ function edit(entry) {
         entry.value = null
         update(entry)
     }}>
-    ${mapp.dictionary.delete_geometry}`)
+    ${entry.edit.delete_label || mapp.dictionary.delete_geometry}`)
 
   entry.elements.push(mapp.utils.html.node`
     <button
       class="flat bold wide primary-colour modify-btn"
       onclick=${()=>entry.modify()}>
-      ${mapp.dictionary.modify_geometry}`)
+      ${entry.edit.modify_label || mapp.dictionary.modify_geometry}`)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
The update command was taken out on purpose.

An API geometry can be stored within a plugin.

An API geometry may not be stored in the location at rest.

Turning on a catchment geometry without a field for display only will crash the process with the change in this PR.

```js
          {
            "label": "HERE Catchment",
            "type": "isoline_here",
            "params": {
              "apiKey": "***"
            }
          },
```

What should the expected behaviour be for ongoing edits?

I have a change in a text area. Displaying the geometry with a field will now update all val change entries and finish the edit mode.

image

simon-leech:
@dbauszus-glx  Yeah i didn't think this was the best way to handle it. 

My use case is a third-party API call to a drivetime that then updates a separate field. The drivetime is not shown on the map (it has no style). When the drivetime call it finished the fields should be updated as it is a dependent of the geometry drivetime field.

dbauszus-glx:
@simon-leech This will have to be done with the blocking capability. The geometry field entry will be ahead of the dependent field entry. 

The geometry entry has the blocking and display flag. If the entry has no value it will be generated. The location view is disabled with the blocking flag. After the geometry has been created and stored the location view will re-render. Dependent fields should be synced.

dbauszus-glx:
@simon-leech What you are trying to do must be controlled in the plugin. 

The syncFields should be added to the finally condition of the call which requests the geometry. Whether a geometry is stored or not should not matter.

[Screencast from 25-06-24 09:54:25.webm](https://github.com/GEOLYTIX/xyz/assets/22201617/6b955ae5-a12b-4113-b5dd-854e8b67c330)


```js
    .finally(async () => {
      if (entry.dependents) {
        await entry.location.syncFields(entry.dependents)
      }
      entry.blocking && entry.location.view.dispatchEvent(new Event('render'))
    })
```

dbauszus-glx (approved):
Labels are working as expected.

image

```js
            "edit": {
              "delete_label": "Del",
              "modify_label": "Mod"
            },
```

PR #1353, opened 2024-06-27
Reverse theme style assignment; Check for useridb flag in locale.

dbauszus-glx:
The theme style assignment in the styleParser was the wrong way. The theme style should be spread into a structured clone of the default style before being assigned as theme style.

The default view script should check for a useridb flag in the locale in addition to the url parameter [hook] check.

If the theme style fillColor is blue, and the default fillColor is red, then blue must be assigned as the fillColor if the theme is current.

Storing styles will not work otherwise because no matter which property you change the default property would always be assigned on top of the theme style, not the other way around.

```js
{
  style: {
    default: {
      fillColor: 'red'
    },
    theme: 
      style: {
        fillColor: 'blue'
      }
    }
  }
}
```

### DIFF
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -239,7 +239,10 @@ export default layer => {
     if (typeof theme.style === 'object') {
 
       // Assign the default style to the theme.style
-      Object.assign(theme.style, structuredClone(layer.style.default))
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
All looks good to me - can't see any issues :)

PR #1359, opened 2024-07-02
Admin view

cityremade:
Admin view redesign including mobile usability;

---
  - https://app.asana.com/0/0/1207866004703647


dbauszus-glx:
The optional method param must end with a question mark. There must only be one get and one post application definition for the same endpoint.

```js
app.get(`${process.env.DIR||''}/api/user/:method?`, api)

app.post(`${process.env.DIR||''}/api/user/:method?`, express.json({limit: '5mb'}), api)
```

The update method will now receive a req.body. I changed the put to update in the user view script.

image

The user add method should be extended to also check for a req.body.

The checks for the params.user must be first. The params user is the admin user.

image

simon-leech:
Various list of changes from me / Admins here :)

They were big fans of the changes so well done @cityremade ! 🥳

**Main View**
1. Capitalise Email in the table header
2. Below Show blocked accounts checkbox add text saying 'Click on an email in the table to make changes to the users permissions and roles'
3. Make New Orange instead of red
4. Make Awaiting Verification red - as it requires fixing

**Form** 
1. Capitalise Email 
2. Update tooltips on checkboxes - should all start 
'Check the box to verify a user' 
'Check the box to re-verify after password reset' 
'Check the box to approve a user' 
'Check the box to make user an administrator'
'Check the box to allow the user API privileges'
'Check the box to block the user' 
3. Option to pass in text about what User Roles are (perhaps an eye that expands).
4. 'Account Expiry' instead of 'Account expires on'

jfitzpatrick17:
Looks really good @cityremade! 👏

One thought I had...should we make the Status column sortable (by clicking on the column header) so it's easier for the user to find all of the accounts which require action?

AlexanderGeere (changes requested):
When i click cancel the changes get saved. I think `save` and `cancel` are overlapping maybe.

cityremade:
In order to use role description add roles object in the workspace:
`"roles": {
    "this role": {
      "label": "This is a label that explains what this role does"
    }
  }`

mod/user/add.js
simon-leech, line 1: Why does this have a if statement for `req.body` that just `console.logs` ?

mod/user/update.js
simon-leech, line 1: Why does this have a if statement for `req.body` that just `console.logs` ?

simon-leech (changes requested):
Few things from me 

❌ With a short user table you end up with a scrollbar that's unrequired. Can we set it to only appear if required? 

image

❓ On mobile the checkbox tooltips are not accessible. Wonder if we should put them one a row one by one with the checkbox and explanation side-by-side? 

❓ Do we need the dialog to say changes have been saved ? As when you click Save the form is closed anyway? 

🌍 Can we convert all the hard-coded english strings to a dictionary object? So we can support multiple languages? (It should use the logged in users language for the language of the administration panel) 

🎨 Could we maybe style the role options slightly different as they are quite hard to distinguish from the underlying form due to similar colour?
image

❓Can we add a little padding before the i 
image

❓ Can we put each Role text and label as a separate bullet point so they are laid out well? For example below my roles are APAC East and Global, so they should be on separate bullet points and separate lines?
image

mod/user/add.js
dbauszus-glx: The user put method should be merged with the user add method. There should not be two methods doing the same but slightly different. I added the console.log() to show the request body and that a separate module is not required.

dbauszus-glx (changes requested):
The `user/put` and `user/add` modules should be merged as they do the same.

cityremade:
> The `user/put` and `user/add` modules should be merged as they do the same.

create and update should be separate. I'd rather merge put + update;

RobAndrewHurst (changes requested):
Can we add the ability to sort by the columns in the table? 

Clicking on them does nothing

cityremade:
> Can we add the ability to sort by the columns in the table?
> 
> Clicking on them does nothing

nope. Sorting was deliberately removed - now search filters are in place.

RobAndrewHurst:
> > Can we add the ability to sort by the columns in the table?
> > Clicking on them does nothing
> 
> nope. Sorting was deliberately removed - now search filters are in place.

Why were they removed?

jfitzpatrick17:
> > > Can we add the ability to sort by the columns in the table?
> > > Clicking on them does nothing
> > 
> > 
> > nope. Sorting was deliberately removed - now search filters are in place.
> 
> Why were they removed?

I also asked for this. I use this all the time in the current console to sort by when people have last logged in...

simon-leech:
I'm getting error on save 
error: column "re_verification" of relation "users" does not exist as i think it should be looking at the verified column when you re_verification?

Also i noticed on mobile the dialog text doesn't wrap so it gets lost off the page. 
image

RobAndrewHurst:
Me thinks this is using an ACL table that is missing a column? It would have been part of #1264

simon-leech:
> Me thinks this is using an ACL table that is missing a column? It would have been part of #1264

Not the case - in #1264  it had a conditional on the field. We need to put this back in as its uses the verified field on the db.

image

RobAndrewHurst:
We don't want to do these tests on private ACL's but on public ones. @cityremade Can you redeploy using the ACL on our neon instance?

RobAndrewHurst (changes requested):
Issues outstanding, it also seems we need a spec of what has been done in the PR, and test out each individual UX.

RobAndrewHurst:
> > Me thinks this is using an ACL table that is missing a column? It would have been part of #1264
> 
> Not the case - in #1264 it had a conditional on the field before storing to the db. We need to put this back in as its uses the verified field on the db -there is no reverfication_field.
> 
> image

Nice pick up! :)

dbauszus-glx (changes requested):
The current admin view as well as any custom admin view must still work.

The updated admin view should be taken out of this PR and provided as a public view for testing the updated user API modules.

We can then test the user API modules on the current view and against the custom view.

The views can be swapped in the future in a separate PR.

cityremade:
Closing as the task has been split into 2 separate changes. This branch is no longer needed.

PR #1360, opened 2024-07-02
Disable location view for blocking api request from geometry entry me…

dbauszus-glx:
The api call for a geometry entry maybe defined as blocking.

In this case the location.view must disabled before the api method is called from the show method.

Iteration of the infoj entries array will `break` while the view is disabled.

The location view must be re-rendered once the API call has been resolved. This will enable the view and re-start the infoj entries iteration.

The display condition call to show the geometry should be prior to the creation of the checkbox.

---
  - https://app.asana.com/0/0/1207704292668498


simon-leech (approved):
Fixes the described issue

PR #1361, opened 2024-07-02
Fix expiry insert

dbauszus-glx:
The INSERTS and VALUES array were left over from legacy logic and must be removed. The insert statement for the user registry is build from the user object instead.

Updated the module documentation to include typedefs.


RobAndrewHurst (approved):
Nice! All works for me!

PR #1362, opened 2024-07-03
Replace legend children elements instead of legend element.

dbauszus-glx:
Replacing the legend in the feartureFields.process method breaks the link between the legend element and the document.

Rendering the legend element thereafter will not update the DOM representation of the element. The element has no parents.

The style.legend children must replaced with the spread children of the theme.legend.node created by the theme method.

The theme methods must still be documented.

The code to replace the children should not be replicated in each theme method. There should be a layer ui utility to redraw the legend if required.

The distributedLegend theme module has been removed as this was a placeholder for an empty theme. The stylePanel method must not crash with a legend method unavailable but simply provide a style panel drawer without a legend.

The global style typedef has been added to the docs.


simon-leech (changes requested):
image

Not sure if this is just me but I'm unable to test this as i'm hitting an error from the style module about drawers when using a  workspace with groups of layers.

dbauszus-glx:
@simon-leech this shouldn't have anything to do with grouped layer but was caused by an undefined entry in contents array.

This should be resolved by making sure that the legend is only pushed into style drawer contents if not undefined.

AlexanderGeere (approved):
Looks good to me

simon-leech:
If you have the distribution count on and then zoom into an area with no point (e.g. the sea). 
Instead of remvoing all the categories from the legend, it leaves them all. 
image

Also if you zoom to an area with no points after having a legend containing 1 category. it does not update to 0. Neither the [value] but it should also remove that final legend element.
image

simon-leech:
Distribution Jenks on an mvt layer doesn't seem to work. It has no legend and all the values are the same.
image

simon-leech (approved):
✅ RESOLVED - If you have the distribution count on and then zoom into an area with no point (e.g. the sea). Instead of removing all the categories from the legend, it leaves them all.

PR #1363, opened 2024-07-04
Lookup theme from string

dbauszus-glx:
A lookup of the theme from the themes object should be attempted if the theme is provided as a key string.


simon-leech (approved):
Working for me :)

PR #1365, opened 2024-07-05
zero config bug and secret

dbauszus-glx:
The auth module must return null if no environment secret is defined in the environment.

The workspace cache variable should not crash if the WORKSPACE env is not defined.

An empty default view should open without crashing the process.

image


dbauszus-glx:
A dialog is shown if no layers are available.

RobAndrewHurst (approved):
image

Nice. Only thing is that on wider screens the modal appears quite high up. But doesn't bother me. Just something I am noting.

dbauszus-glx:
@RobAndrewHurst this should only be temporary as these are to be replaced with modal dialogs in the future.

PR #1366, opened 2024-07-08
infoj_skip in layer decorator

dbauszus-glx:
The infoj_skip array should only processed once when the layer is decorated.

Creating a new Set on each iteration of multiple entry object properties on each iteration of every entry in the infoj is incredibly resource wasteful.

The uncommented negated array check in the filters panel lookup is hard to understand test or maintain.

It is much easier to set a flag in the infoj_array based on inclusion in the infoj_array which must a be an array and therefore supports the [array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -134,6 +134,24 @@ export default async function decorate(layer) {
   // Set layer opacity from style.
   layer.L.setOpacity(layer.style?.opacity || 1);
 
+  // Check which infoj entries should be skipped.
+  if (Array.isArray(layer.infoj_skip)) {
+
+    layer.infoj
+      .filter(entry => [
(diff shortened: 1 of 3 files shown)


RobAndrewHurst:
image

When skipping an entry I am getting an error.

You can replicate this by using the workspace in infoj/infoj_skip.json

RobAndrewHurst (changes requested):
See comment above.

dbauszus-glx:
I must have accidentally deleted the tip of the arrow function. @RobAndrewHurst can you test again.

RobAndrewHurst (approved):

simon-leech (approved):

PR #1368, opened 2024-07-10
location.create() method as replacement for draw interaction callback with defaults/default_fields location updates.

dbauszus-glx:
The `layer.draw.callback()` method has been removed from the layer.decorator method.

The replacement method is `mapp.location.create()` which is defined in the mapp/location create module.

The location.create method will be called from the drawing elements drawOnclick methods interaction callback.

The location.create method will assign defaults properties from interaction properties with interaction.default_fields as lookup.

This is required to assign defaults from api geometry methods, eg. drawing isolines.

The location/decorate module documentation has been updated. Modules do not have params.

The location.decorate method should be named decorate.

The mapview module documentation has been updated. Modules do not have params.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -65,66 +65,6 @@ export default async function decorate(layer) {
     console.warn(`Layer: ${layer.key}, please move draw.delete to use layer.deleteLocation:true.`);
   }
 
-  // Callback which creates and stores a location from a feature returned by the drawing interaction.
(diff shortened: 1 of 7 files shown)


simon-leech (approved):
I've tested drawing circles. polygons, points, and also using custom plugins and can see the defaults being stored to the db - all looks good to me 😅

AlexanderGeere (approved):
Looks good to me

PR #1369, opened 2024-07-11
Review and document entry.default, entry.nullValue, and defaults.

dbauszus-glx:
The `defaults` entry method has been deprecated as this does only a subset of the functionality provided the userDatetime plugin.

This is to prevent duplicate code. The plugin code is more advanced and can be maintained outside the core easing testing of the core code.

The entryDefault and entryNullValue methods have been reviewed and documented.

The core difference being that defaults should only apply to an editable entry providing the default as new value if the entry value is undefined.

The nullValue will only be assigned to non editable entries if the entry.value is undefined.


dbauszus-glx:
@AlexanderGeere I cannot reproduce this.

When I create a new location which is editable the site_name is editable and empty.

```js
          {
            "title": "Site Name",
            "field": "site_name",
            "inline": true,
            "edit": true,
            "nullValue": "N/A"
          },
```

image

AlexanderGeere (approved):

RobAndrewHurst (approved):
Nice one!

simon-leech (approved):
Good with me too

PR #1373, opened 2024-07-12
Check for ACL module being a function.

dbauszus-glx:
The ACL module will export the require object as an empty object and not resolve as the ACL function default export from the ACL module.

The cookie request must check whether an user object or an error object is returned from query,


simon-leech (approved):
Working for me - I tested by removed the PRIVATE key and prior to this before the debugger crashes out. 

On this PR I get an 'ACL unavailable' message.

PR #1375, opened 2024-07-16
Dataview entry exception for show method.

dbauszus-glx:
The dataview decorator is async for legacy reasons.

However this means that the dataviiew decorator returns a promise which will not be caught as instance of error by the dataview entry method.

The dataview entry method may attempt to show the dataview and crash out if the dataview check has failed.

The dataview entry method should check whether the entry.show method exists on execution.

An error should be logged in the console.

It must be possible to select locations with a caught error exception.


simon-leech (approved):
Working for me - I removed the `tabulator` plugin from the workspace. 

I get errors saying `mapp.ui.utils.tabulator doesnt exist` as expected, but I am still able to select a location even with this error.

AlexanderGeere (approved):
lgtm

PR #1376, opened 2024-07-16
feature_info highlight plugin; highlight popup; mapview typedef

dbauszus-glx:
The `feature_info` core plugin provides a custom highlight interaction to display a feature info in a mapview popup.

The highlight interaction was modified to not short circuit with a popup currently displayed. This wants a good test as I am not sure why this was required in the first place. Unfortunately the short circuit was never commented on in the first place.

A typedef for the mapview has been added to the globals.

The mapview decorator method has been named.

[Screencast from 16-07-24 13:38:03.webm](https://github.com/user-attachments/assets/b8ac5d93-0e2d-467b-8066-7d85822aca16)


AlexanderGeere (changes requested):
We should probably strip some cluster feature as the popup gets very big when clickiing on some points:

Screenshot 2024-07-17 142208

dbauszus-glx:
By default the features, geometry, and Styles array will now be stripped from the popup object.

image

Setting the log flag will log the F object in the console with all the stuff.

image

Setting the features flag in the plugin config will add the layer.cluster.label as features array to the feature properties.

image

simon-leech (approved):
Happy with this plugin - tested on a range of layer types and is a useful debugging tool

AlexanderGeere (approved):
Nice! Looks good

PR #1377, opened 2024-07-16
Only check highlight style if typeof object

dbauszus-glx:
Screenshot from 2024-07-16 15-37-48

The styleParser must not assign an empty highlight style.

Otherwise the highlight interaction will re-style an mvt layer, even if there is nothing to re-style.

### DIFF
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -85,11 +85,13 @@ export default function styleParser(layer) {
 
   layer.style ??= {}
 
-  // Assign a default highlight style and ensure that zIndex is infinity if not implicit.
-  layer.style.highlight ??= {}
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
This can be tested by setting a breakpoint in the styleFeature method. This shouldn't break if there is nothing to style.
[Screencast from 17-07-24 11:59:30.webm](https://github.com/user-attachments/assets/ee358fa9-3d10-4234-b2e1-0d8de288a31b)

simon-leech (approved):
Working for me - breakpoint was previously hit repeatedly when hovering over locations on the map. 
On this PR the breakpoint is not hit.

RobAndrewHurst (approved):

PR #1380, opened 2024-07-17
Log error instead of warning for missing entry.type in infoj check ; log error instead of warning for invalid format.

dbauszus-glx:
Instead of a warning the the infoj method should console an error when the `entry.type` is not an own property of the `mapp.ui.location.entries{}` method.

image

Add requires for entries module.

image

### DIFF
--- a/lib/mapview/addLayer.mjs
+++ b/lib/mapview/addLayer.mjs
@@ -24,6 +24,9 @@ export default async function (layers) {
       layer.display = layersSet.has(layer.key);
     }
 
+    // Layer without format should not be decorated.
+    if (!layer.format) continue;
+
     // Check the layer format.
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
All good for me. 
I pushed a minor commit as I noticed that we were warning about missing layer file path - when we should error for that too .

dbauszus-glx:
@simon-leech Can you check with undefined layer.format. It is valid to define a layer without format in order to use the template but not add the layer. This must not err.

dbauszus-glx:
@simon-leech  I added the check on layer.format but we shouldn't add too much more into this PR.

Will update the PR title.

AlexanderGeere (approved):
Looks good to me.

RobAndrewHurst (approved):

Issue #1370, opened 2024-07-11
Empty editable numeric field


dbauszus-glx:
Removing focus from an editable empty numeric field collapses the input element like so.

![image](https://github.com/GEOLYTIX/xyz/assets/22201617/2f6de1cb-ebba-42f2-ae66-c923eb092262)

PR #1381, opened 2024-07-17
Numeric input css fix

AlexanderGeere:
## Description

Adding `display-none` to the input was shrinking the box when there was nothing in the box, removed this behaviour.

Additionally: 
 - formatting should now occur as you type in the field.
 - Setting locale to null should remove all formatting.

## GitHub Issue
[#1370](https://github.com/GEOLYTIX/xyz/issues/1370)

## Type of Change

## How have you tested this? 
Tested Locally on `bugs_testing/numeric_formatting/workspace.json`

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

---
  - https://app.asana.com/0/0/1207876619030553


simon-leech (changes requested):
A few little issues I noticed
1. You are unable to put decimals in, it just seems to remove them
2. If you click on an input, type into it, click onto it again, it gets wiped empty but then doesn't re-populate when you click off again.

RobAndrewHurst:
image

Interesting behaviour that if you try save an absurdly large integer the location disables (as it should when making the request to save). But the request never completes, and I don't see any errors.

RobAndrewHurst (changes requested):
See comment above!

RobAndrewHurst:
image

@AlexanderGeere I am getting this error when adding a point on a scratch layer

RobAndrewHurst (approved):
Dit lyk goed my chom!

dbauszus-glx (changes requested):
The input fails on an integer filter input element.

image


The type is integer but there are no formatter params.

dbauszus-glx:
Added the /ui/locations/view module and method to the docs.

/docs/mapp/module-_ui_locations_view.html

The valChange is now a member of this module and not declared inside the view method.

image

http://localhost:3000/docs/mapp/module-_ui_locations_view.html#~valChange

dbauszus-glx:
The location.update() method must return an error if a location entry is invalid.

The toggleLocationViewEdits() methods has been taken out of the location.view method and documented.

image

**/docs/mapp/module-_ui_locations_view.html#~toggleLocationViewEdits**

The onClickEditToggle() method has been taken out of the location.editToggle element and documented.

image

The method will now make use of the location.renderLocationView() which also has been documented.

**/docs/mapp/module-_ui_locations_view.html#~renderLocationView**

dbauszus-glx:
For numeric inputs it must be possible to define a max and min. This can be within the edit property or as entry properties. The edit object is merged with the entry object in the numeric entry edit() method.

Outside the min and max the entry musst be invalid.

It must not be possible to update the location with an invalid entry.

The default formattingParams.locale is for the mapp.language.

Setting formattingParams to null should prevent formatting.

```js
{
  "title": "num",
  "field": "numeric_field",
  "type": "numeric",
  "edit": {
    "min": -100000,
    "max": 10000
  },
  "suffix": "%"
},
{
  "title": "int",
  "field": "integer_field",
  "type": "integer",
  "formatterParams": null,
  "edit": true,
  "min": -100000,
  "max": 10000
  "prefix": "£"
}
```

dbauszus-glx:
Document numericChecks method.

image

dbauszus-glx:
The /ui/locations/entries/numeric module is now documented

**/docs/mapp/module-_ui_locations_entries_numeric.html**

A condition has been added for the range slider. The value must be between min and max. Otherwise a numeric input element will be displayed.

The slider element has now been documented.

**/docs/mapp/module-_ui_elements_slider.html**

The sliderElement value will be updated by the numericInput oninput method.

dbauszus-glx (approved):

dbauszus-glx:
I have removed the formatter from filter and slider_ab.

simon-leech (changes requested):
Ignore the previous stuff - forgot to build after pulling 😬

Two things I've noticed
1. Typing a non-number in the filter boxes puts 'NaN' and throws the slider off the screen. This should use the invalid CSS and not break the slider
image

3. We've removed the formatting on the numbers in the sliders? This is something clients will notice so are we sure we want to do this?

dbauszus-glx:
The slider_ab and filter modules have now been updated.

The filter method will merge the field entry into the filter object which is provided as params argument to the slider_ab element method.

The slider_ab [like the slider] element creates a numericInput element which handles callbacks, formatting, and numeric checks.

The range type input only assign a number as value to the numericInput elements which handle the formatting on change.

The callback is only called if numeric checks pass.

RobAndrewHurst (approved):
Looks all good to me

simon-leech (changes requested):
Happy with this! 

✅ Tested again on infoj entry - type integer with min and max edit options and without.
✅ Tested again on filter - formatting is applied and invalid values use CSS to highlight this. 

❌ Tested again on infoj entry - type numeric with min and max edit options and without. I don't seem to be able to put decimals into a type numeric? 
❌ You shouldn't be able to see the save button on ToggleLocationViewEdits when any entry has invalid values - presently if you type a valid number, then keep typing to make it invalid - the save button remains able to be pressed.
❌ Placeholders aren't working at the moment for infoj entry of type numeric or integer

dbauszus-glx:
[Screencast from 02-08-24 12:49:24.webm](https://github.com/user-attachments/assets/20359516-9a61-4ef8-8080-dc0a2b20e742)

The callback for numericInputs must be run before the invalid shortcircuit. This is to ensure that the location view valchange detects invalid inputs and hides the update button if necessary.

dbauszus-glx:
Fraction input should now be possible by storing a trailing fraction character in the unformat method.

dbauszus-glx:
The placeholder should work now. But only if neither suffix, nor prefix are defined.

simon-leech (changes requested):
The only thing I noticed is that with a type numeric infoj editable entry - you can't put a 0 as the first number after the . 
For example - if you try typing 99.01 it won't let you put the .0 in

dbauszus-glx:
@simon-leech 
Zero fractions at the end of the string value should now be preserved.

simon-leech (approved):
I'm happy with this one now! 🥳🥳

PR #1383, opened 2024-07-17
Assign getLayer template error message; catch and log error in addLayer

dbauszus-glx:
An error object can not be sent with the layer json.

The Error.message can be assigned to the layer json instead.

The addLayer() method can check for an err message and log this.

image

### DIFF
--- a/lib/mapview/addLayer.mjs
+++ b/lib/mapview/addLayer.mjs
@@ -16,9 +16,8 @@ export default async function (layers) {
 
     // The layer.err will be assigned from a failure to fetch a layer template.
     if (layer.err) {
-
-      console.error(layer.err)
-      continue;
+      // Error for each of the array items.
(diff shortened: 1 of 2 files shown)


simon-leech:
I pushed a commit to change `layer.err` into an array. 
I did this as we also need to check on `layer.templates` inside the `getLayer` method. 

Example configuration for testing 
``` json
"templates":{ 
"test_1": {
"template": "path/path.json",
},
"test_2_template": {
"template": "path/path.json",
},
,
"test_3_template": {
"template": "path/path.json",
}
}, 
"layers": {
  "test_1": {
  "templates": ["test_2_template", "test_3_template"] 
}
}
```

The example above, before my commit you would only see an error for the overall layer file path being wrong. 
With this commit we now also check on the file path of anything in the `layer.templates` array, and error those too.

RobAndrewHurst (approved):

PR #1385, opened 2024-07-18
Allow for array updates with text entry arraySeparator

dbauszus-glx:
In order to set and update an array type database field the arraySeparator edit property is introduced in this PR.

```js
"edit": {
  "arraySeparator": ","
}
```

The `onInput` method of the text input element will use the arraySeparator character to split a provided target element string value into an array provided as newValue property value to the entry.

[Screencast from 18-07-24 09:00:39.webm](https://github.com/user-attachments/assets/539543de-94a6-40a2-8351-218e33da7bf1)

### DIFF
--- a/lib/ui/locations/entries/text.mjs
+++ b/lib/ui/locations/entries/text.mjs
@@ -1,3 +1,11 @@
+/**
+## ui/locations/entries/text
+
+The locations entries text module exports the default text entry as mapp.ui.locations.entries.text(entry)
+
+@module /ui/locations/entries/text
+*/
+
 mapp.utils.merge(mapp.dictionaries, {
   en: {
     loading: 'Loading',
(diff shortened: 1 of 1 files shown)


RobAndrewHurst:
Interesting!

simon-leech (changes requested):
On testing this on a layer with a display true layer dataview - when you press save on a location you hit an edit from here. 

image


The actual arraySeparator flag is working as expected though - a nice feature :)

dbauszus-glx:
@simon-leech The bug is unrelated to this feature and addressed in this PR. https://github.com/GEOLYTIX/xyz/pull/1395

simon-leech (approved):

AlexanderGeere (approved):

PR #1392, opened 2024-07-24
Changes required to implement user update with payload;

cityremade:
1) Allow request body on user/* endpoints - express.js
2) Pills ui element - stop propagation on pill clicks
3) user/update.js - it can process payload unlike before where values to update were only passed in url parameters
4) _workspace.js - now supports roles descriptions defined in the workspace.

This contains non-breaking changes that facilitate new admin view.
In order to approve compatibility with current admin panel and user updates should be reviewed.


cityremade:
Admin view now in public for testing purposes https://geolytix.github.io/public/mapp/admin_view.html

RobAndrewHurst:
Cognitive load is too high on the update.js and needs to be fixed. Putting back into draft

cityremade:
phew. cognitive load fixed.

dbauszus-glx (changes requested):
This is a dangerous security vulnerability.

Values must not come directly from body into template but should be substituted.

```js
    const updatedUser = Object.entries(req.body)
      .filter(i => i[0] !== 'email')
      .map(i => { return `${i[0]} = '${i[1]}'` })

    update_query = `UPDATE acl_schema.acl_table
    SET ${updatedUser.join(', ')}
    ${verification_by_admin}
    ${approved_by}
    WHERE lower(email) = lower('${req.body.email}');`
```

There is also some duplicate code which I want to address.

The use of the `let` is a red flag. 

mailer_options are declared and then re-assigned. This isn't necessary.

Screenshot from 2024-07-31 18-20-58

dbauszus-glx:
### Roles
The roles return from the _workspace module has been taken out from this PR. There are potential issues which need to be addressed in a separate PR and associated ticket.

### Documentation
It should be prevented to nest properties. eg. The `user.admin` property should not be nested as `req.params.user.admin`
```
@param {req} req HTTP request.
@param {req} res HTTP response.

@property {Object} [req.body] 
HTTP Post request body containing the update information.
@property {Object} req.params 
HTTP request parameter.
@property {string} params.email 
User to update
@property {string} params.field 
User record field to update
@property {string} params.value 
Update value for user record field.
@property {Object} params.user 
Requesting user.
@property {boolean} user.admin 
Requesting user is admin.
```

### Variable Declaration.

Please always use const and only fallback to let if required. eg. In this script only the password_reset needs to be re-assigned if a user if being verified.
```js
  let password_reset = ''

  if (update_user.verified) {

    // Verifying a user will also approve the user, reset password, and failed login attempts.
    Object.assign(update_user, {
      password_reset: null,
      failedattempts: 0,
      verificationtoken: null,
      approved: true,
      approved_by: `${req.params.user.email}|${ISODate}`
    })

    password_reset = `password = password_reset,`
  }
```

### Cognitive Load
The idea of cognitive load is that a method is hard to understand, document, maintain, and test. Simply throwing lines of script into nested methods is a _defeat device_ which gets rid of the sonar cloud warning but effectively increases the cognitive. This is especially true for nested functions without supporting jsdocs. Nested functions are better than functions declared in a template literal. But if functions are required to do a specific part of repeated logic they should be outside the closure from where they are called.

### Duplicate Logic
Duplicating the logic for updates from req.body or req.params does not have to increase the cognitive load.
A user_update object can be created from the request.body in one line of code. The user_update object is then decorated from request.params if available before being used to create the update query.
```js
  // Create update_user from request body or create Object with email from params.
  const update_user = req.body || {
    email: req.params.email
  }

  if (req.params.field === 'roles') {

    // The value string must be split into an array for the roles field params.
    req.params.value = req.params.value?.split(',') || [];

  } else if (req.params.field) {

    // Assign field property from request params.
    update_user[req.params.field] = req.params.value
  }
```

### SQL Injection
Variables must never be injected unchecked into SQL query templates. Where possible variable substitution in the database should be used.
```js
  // Populate properties, substitutes array for update_query.
  Object.keys(update_user)
    .filter(key => key !== 'email')
    .forEach(key => {
      substitutes.push(update_user[key])
      properties.push(`${key} = $${substitutes.length}`)
    })

  const update_query = `
    UPDATE acl_schema.acl_table
    SET 
      ${password_reset}
      ${properties.join(',')}
    WHERE lower(email) = lower($1);`
```

Where not possible, eg. for fields, tables, etc. the value must be validated prior to template injection. We can validate keys by checking that only whitelisted characters are used.
```js
  // Validate update_user keys.
  if (Object.keys(update_user).some(key => !/^[A-Za-z0-9.,_-\s]*$/.test(key))) {

    // Return bad request 400 if an update_user key contains not white listed characters.
    return res.status(400).send('Invalid key in user object for SQL update.')
  }
```

### Variable names
Please use i only for index or integer, e for event, not for anything iterative. entry for entry, key for key, param for params[], property for properties[], etc. is preferred.

dbauszus-glx (approved):
This works for me. Please check last comment in regards to testing SQL injection.

RobAndrewHurst (approved):
Nice one! 

Works well!

I have added in tests to check updating a user via a body, or with the vanilla (params) method. 

I also renamed the utils/xhr.mjs function name from `default` to 'xhr' just for readability.

PR #1395, opened 2024-07-25
Fix issue where dataview is disconnected from layer object.

dbauszus-glx:
Dataviews are not associated with the layer object if referenced from the entry.location.layer, for example in an update statement. The reason for this being the entry[1] being spread into the dataview object in the dataview decorator.

image

The effect of this being that the dataview cannot be referenced through the location.layer in the location update method.

image


cityremade (approved):
looks good to me, straightforward change.

simon-leech (approved):

PR #1396, opened 2024-07-25
findlast keyvalue dictionary entry for multiple options.

dbauszus-glx:
The keyvalue_dictionary is an array which may contain multiple entries with the same key/value.

```js
{
  "locale": {
    "keyvalue_dictionary": [
      {
        "key": "name",
        "value": "OSM",
        "default": "OpenStreetMap"
      }
    ],
    "roles":{
      "*": true,
      "A":{
        "keyvalue_dictionary": [
          {
            "key": "name",
            "value": "OSM",
            "default": "A"
          }
        ]
      },
      "B":{
        "keyvalue_dictionary": [
          {
            "key": "name",
            "value": "OSM",
            "default": "B"
          }
        ]
      }
    },
    "layers": {
      "OSM": {
        "display": true,
        "name":"OSM",
        "format": "tiles",
        "attribution": {
        }
      }
    }
  }
}
```

Adding to the dictionary with roles will create an array like this for a user with roles A and B.

image

Instead of find [the first matching entry] the findLast array method should be used to assign the last matched entry.

---
  - https://app.asana.com/0/0/1207890533562710


AlexanderGeere (approved):

simon-leech (approved):

simon-leech:
I've just pushed 3 tests for the KeyValue Dictionary Plugin that are all passing so this is all approved from me :)

PR #1402, opened 2024-08-01
Add details param for workspace/roles request.

dbauszus-glx:
The workspace/roles request returns an array of roles parsed from the workspace object.

image

This behaviour must not change in regards to backwards compatibility.

This PR documents the workspace module and exported method. The workspace roles method, and the roles.get method.

Providing the `detail` URL parameter flag will return the workspace.roles object.

image


cityremade (dismissed):
I have tested with new admin view. Some changes required there but this looks good.

cityremade:
ok one thing: we now take roles from workspace and details from separate description. I think this is prone to error. Possibly the roles should be first taken from workspace AND then description should be added. 

Example:
a role defined only within locale/layer will **not** appear if it's not defined in the descriptive roles object

cityremade (dismissed):
To change: get roles from the locales/layers and then decorate with description - otherwise we lose roles not described and this is a room for error.

cityremade (approved):

dbauszus-glx:
The roles.get utility method has been resolved into the workspace.roles method.

That utility is only useful in that single location and having to call a utility from a method to test that method is more awkward than testing the method in the first place.

simon-leech:
I've added a series of tests to test against these api calls. 
image

These make use of the `workspace.json` configuration of 
``` json
 "roles": {
        "A": "Text about A",
        "B": "Text about B"
    },
    "locale": {
        "roles": {
            "A": null,
            "B": null,
            "C": null,
            "*": null
        }
}
```

simon-leech (approved):

PR #1404, opened 2024-08-01
Vector layer format error handling

dbauszus-glx:
The vector format layer.reload() method has been unnested from the vector method and documented.

The xhr request should not attempt to add an Error instance returned from a failing data request to the source.

The featureFormats WKT method has been documented on testing.

To Test this : 
Set an incorrect dbs connection for the layer on load. 
You should now see just one error message instead of 2 that are seen off this PR.

### DIFF
--- a/lib/layer/featureFormats.mjs
+++ b/lib/layer/featureFormats.mjs
@@ -31,13 +31,26 @@ export function geojson(layer, features) {
   })
 }
 
+/**
+@function wkt
+
+@description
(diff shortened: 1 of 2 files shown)


simon-leech (approved):
Working for me - tested with an incorrect DBS connection and only get one error message.
On main currently - I get two error messages.

AlexanderGeere (approved):
Nice one!

PR #1407, opened 2024-08-05
bump ol lib

dbauszus-glx:
Bump OL to version 10.

The only breaking change is with the getFeaturesInExtent method which we don't use.


AlexanderGeere (approved):
tested on a few instances, looks good to me!

simon-leech:
The current version of `_ol` was still set to 9.1. I've updated to 10.0. 

This is required as in custom views we check on the version to make sure we are loading the right one compared to the version used in MAPP 

image

simon-leech (approved):

PR #1408, opened 2024-08-05
Geometry entry draw callback

dbauszus-glx:
It must be possible to assign a custom callback function to a draw interaction.

Drawing geometries from a geometry entry require a custom callback which does not create a new location but updates the geometry entry field.

Updated geometry entry docs to make use of `@this` and `@property` for param properties.


simon-leech (changes requested):
I don't seem to be able to modify an infoj entry geometry - i hit an error here as `this.location` is undefined.

dbauszus-glx:
The issues with methods being called from target element events, ie modify button or location.view should have been addressed now. @simon-leech

simon-leech (changes requested):
I can now modify, delete and re-draw geometries. 

One minor issue I noticed, the `helpModal` is not removed at the end of drawing interactions on a location infoj entry

dbauszus-glx:
The helpDialog should now be removed in the drawCallback assigned to the interaction by the geometry entry draw method.

simon-leech (approved):
This is working as expected for me now! :)

AlexanderGeere (approved):
Looks good!

PR #1411, opened 2024-08-07
Dataview location stroke color;

cityremade:
Added primary colour as a fallback when location.style object not found after positive tabview check;

Previously when tabview assigned in a custom view location.style would be unset therefore no location.style.strokeColor;

When no strokeColor found var(--colour-primary) assigned;


dbauszus-glx (approved):
lgtm

simon-leech (approved):

PR #1413, opened 2024-08-12
Do not assign window as caller for vector layer reload method

dbauszus-glx:
calling a bound method like this from a [window] setTimeout() method assigns the window as this.

```js
setTimeout(layer.reload, 100)
```

### DIFF
--- a/lib/layer/format/vector.mjs
+++ b/lib/layer/format/vector.mjs
@@ -124,7 +124,7 @@ export default function vector(layer) {
 
       clearTimeout(layer.timeout)
 
-      layer.timeout = setTimeout(layer.reload, 100)
+      layer.timeout = setTimeout(()=>layer.reload(), 100)
     })
   }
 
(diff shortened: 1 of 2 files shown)


tests/workspace.json
simon-leech, line 1: Does this need to change in this PR? 
Isn't this in line with the xyz_settings?

lib/layer/format/vector.mjs
simon-leech, line 1: All good here :)

simon-leech (changes requested):

tests/workspace.json
```diff
@@ -1,7 +1,7 @@
 {
     "locale": {
         "plugins": [
-            "${PLUGINS}v4.8.0/coordinates.js"
+            "${PLUGINS}/v4.8.0/coordinates.js"
```
simon-leech, line 4: Does this need to change in this PR?
Isn't this in line with the xyz_settings?

dbauszus-glx: No it doesn't need to change in this PR. In fact there should be an issue to review whether a workspace with private variables should be in the repository.

simon-leech (approved):

Issue #1154, opened 2024-02-23
mapp.ui.layers.panels.style composition


dbauszus-glx:
The style panel module creates a drawer with a legend and supporting style elements for hover and label. Changing the theme will reload the panel.

It should be possible to have the legend from the theme with the label and hover control outside the style panel.

A method to construct the style view with selected elements should be available to be called from the panel.

dbauszus-glx:
#1362 must be merged before this can be addressed.

simon-leech:
When using a `categorized` thematic with `fields` array - the `switch-all` text is not applied. This is correct as `switch-all` cannot work. 
However, there should be a default applied for the user. 

We need to ensure we test `theme.meta`.

PR #1414, opened 2024-08-12
Add icon_scaling to featureStyle methods.

dbauszus-glx:
The icon_scaling is configured by the `icon_scaling{}` object in the layer.style.

An `icon_scaling.field` property is required on features to scale icons.

The feature fields params array was previously created in three different locations.

The `fieldsArray()` method was added to the `featureFields` module to generate the `params.fields[]` array for feature data requests. This method is now referenced in the vector and MVT format modules. The icon_scaling.field has been added to fields requested for feature properties.

The featureFields module has now been documented. The reset and process methods will regard all fields in the `layer.params.fields[]` array rather than look for current theme.field. The lookup for the theme.field or fields is already done in the fieldsArray() method.

The `featureFormats` module methods will now add all layer.params.fields to the layer.featureFields. The lookup for the theme.field or fields is already done in the fieldsArray() method.

The `iconScale()` method was added to `featureStyle` module. The `icon_scaling.field` feature property value will be used for scaling icons.

The `feature.style.fieldScale` will be calculated by adding 1 to the fraction of the fieldValue and `icon_scaling.maxFactor`. Increasing this factor will increase the scale of the largest icons. The `olStyle` module method will multiply the icon.scale by the feature.style.fieldScale.

Cluster feature icons are not scaled by default. Adding the `icon_scaling.cluster` flag will sum the feature field properties of all cluster features. Setting `icon_scaling.cluster = "avg"` will divide this sum by the length of the cluster features[] array length to calculate an average.

If not implicit, eg. 100 for percent the maxFactor will be calculated in the `featureFields.process()` method as the max value from the array of the featureFields icon_scaling.field values[] array.

```js
"icon_scaling": {
  "field": "name", // required!
  "cluster": "avg", // can be true for sum.
  "maxFactor": 50 // will be calculated if not implicit.
}
```
The iconScale method will also apply the `style.zoomInScale` and `style.zoomOutScale` which are not associated with cluster layer by definition.

The `featureStyle` module has been cleaned up. Params which are properties have been removed from docs. Methods have been unnested.

A style typedef for style object has been renamed to `feature-style` typedef for clarity.

The `clusterChecks` method in the `styleParser` has been unnested from the warnings which are nested in the styleParser method themselves. This is a bit of a mess but should be addressed in a separate PR.

The clusterChecks method has now been documented. The zoomInScale and zoomOutScale are moved to the layer.style{} config from the cluster style.

The `evalIconObject` method has been removed from the styleParser as this could never have been called.

```js
  // Define default style cluster icon
  layer.style.cluster = {
    clusterScale: 1,
    icon: {
      type: 'dot'
    },
    ...layer.style.cluster
  }

  // This does not seem possible.
  // if (!layer.style.cluster.icon) {

  //   layer.style.cluster = {
  //     ...layer.style.cluster,
  //     ...layer.style.cluster.style
  //   }

  //   delete layer.style.cluster.style

  //   layer.style.cluster.icon = evalIconObject(layer.style.cluster)
  // }
```

There will NOT be an interface for the icon_scaling method in this PR. The interface will be added in a PR for #1154.


simon-leech (changes requested):
I think this should work whereby we get the `min` and `max` of the field using the `minmax` template and then scale the value proportionally across the range of min to max. 

Currently i'm testing on a field that holds values that go up to millions (a sales field) - and the icons take up the whole world in the mapview, its not usable. 

I even tried put the maxFactor down to 0.000000000000000000000000001 but still they are too large to use!

dbauszus-glx:
@simon-leech We can lookup the max value if undefined but it must be possible to manually define this.

Imagine a datasets of [10,30,50] which are percent. In order to set these to be scaled percent it must be possible define 100 as the maxFactor for scaling.

simon-leech:
@dbauszus-glx  Another question - If i leave off the `avgCluster` flag, the cluster icons are gigantic as I guess they are summed values for all features within them? 
Is it possible to say you only want to apply the icon scaling to the individual feature icon and ignore cluster features?

dbauszus-glx:
> @dbauszus-glx Another question - If i leave off the `avgCluster` flag, the cluster icons are gigantic as I guess they are summed values for all features within them? Is it possible to say you only want to apply the icon scaling to the individual feature icon and ignore cluster features?

That's correct. Sum is the default. avgCluster calculates the average. I can make scaling of cluster optional.

simon-leech:
> > @dbauszus-glx Another question - If i leave off the `avgCluster` flag, the cluster icons are gigantic as I guess they are summed values for all features within them? Is it possible to say you only want to apply the icon scaling to the individual feature icon and ignore cluster features?
> 
> That's correct. Sum is the default. avgCluster calculates the average. I can make scaling of cluster optional.

Yeah I think that would be a good feature, 
maybe we could have 
"cluster": "average" or "cluster": "sum" or if you don't include it then it just doesn't scale the cluster?

dbauszus-glx:
@simon-leech 

I've added the calculation for the maxFactor in the revised `featureFields` module methods.

Cluster feature scaling is now optional and can be sum [default] or average.

I've updated the initital comment with details.

RobAndrewHurst (commented):
Quick question what is the use case for defning the maxfactor in the style? Instead of just having it calculated?

RobAndrewHurst (approved):
Actually I can see it being used for things like percentages, indexes etc.

RobAndrewHurst (approved):
Nice! I am happy with this!

And i have a good idea of how to write tests for this!

AlexanderGeere (approved):
Looks good to me

simon-leech (changes requested):
Review ongoing - please don't merge yet

dbauszus-glx:
The icon_scaling.max will always be calculated in the featureFields process method.

The icon_scaling.scaleFactor will always be used if implicit.

The factor to calculate the scaling will be the max if not implicit as scaleFactor. A fallback of 1 is taken if neither is not falsy to prevent an item from not being displayed at all.

simon-leech (approved):
I'm happy with how this is working now - looking good! 
Thanks @dbauszus-glx  👍

tests/lib/layer/featureStyle.test.mjs
```diff
@@ -1,7 +1,57 @@
...
+         * 2. Get the layer from the mapview
+         * 3. Get an id to perform a select on.
+         * 4. Get the location.
+         * 5. Set the newValue of the icon_scale field 
+         * 6. Update the location.
+         * 7. Assert that we have a new max value
+         * 8. Reset the icon_scale field to null
+         * @function it
+        */
+        await codi.it('featureStyle: Icon Scaling', async () => {
```
RobAndrewHurst, line 26: Please can someone review this test and help me with the assertion!

RobAndrewHurst (approved):
I am happy for this to go in. 

The tests needs a set of eyes though 👀

dbauszus-glx:
I updated the featureFormats test to test for style.icon_scaling on wkt format with a minimum setup.

```js
    codi.it('Test style.icon_scaling for WKT features', () => {

      const mapview = {
        "srid": "3857"
      }

      const layer = {
        "format": "wkt",
        "params": {
          "fields": [
            "size"
          ]
        },
        "style": {
          "icon_scaling": {
            "field": "size"
          }
        },
        mapview
      }

      const features = [
        [
          1,
          "POINT(-14400 6710852)",
          null
        ],
        [
          2,
          "POINT(-14920 6711046)",
          10
        ],
        [
          3,
          "POINT(-15322 6710796)",
          15
        ]
      ]

      mapp.layer.featureFormats[layer.format](layer, features)

      return layer.style.icon_scaling.max === 15
    });
```

I am not sure about the featureStyle test structure. There should be test for the layer.featureStyle methods but these should check whether a feature and layer passed to the feature style method get the JSON style assigned which we would expect the feature to get assigned. 

Tests shouldn't involve cloud databases but should work with controlled test data. The current featureStyle.test may fail because the database cannot be accessed. 300 may not be the max field value. The location.update could fail. All these methods should be tested individually.

simon-leech:
Liking that test @dbauszus-glx  - I have just reset the featureStyle test to "TO DO" and removed the content. As the test you pushed is testing the same thing as in this PR.

RobAndrewHurst (dismissed):
The featureFormats tests for the icon scaling needs to make an assertion otherwise the test will pass every time.

RobAndrewHurst:
I have updated the test so that it now asserts the value with the codi library. There are still issues with checking that the max is recalculated when the data set is updated, which is what that previous test that was removed was doing.

RobAndrewHurst (approved):

PR #1418, opened 2024-08-14
Prevent this being the location.view element

dbauszus-glx:
Calling render event of the location.view element fails.

```js
location.view.dispatchEvent(new Event('render'))
```

`this` inside the renderLocationView() method will be location.view element in this case and not the location to which the method was bound.


simon-leech (approved):
This is working correctly for me - I tested with an `infoj` `entry` containing the `blocking: true` flag. 

On main, it failed as the `location.view` element was passed to it, on this PR it works as expected

PR #1419, opened 2024-08-15
elements.slider val/value property and layer.style.opacity slider element callback

simon-leech:
# Slider Tweaks

## Description

1. Assign val to value if not defined (as we changed this recently)
2. Fix opacitySlider
3. Add translation for opacity slider
4. Remove scaleSlider as doesn't work and superseeded by plugin

## Type of Change

## How have you tested this? 
Tested it locally.

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/elements/slider.mjs
+++ b/lib/ui/elements/slider.mjs
@@ -29,6 +29,9 @@ export default function slider(params) {
   // Range inputs require a step.
   params.step ??= 1
 
+  // Set value to val if value is not defined.
+  params.value ??= params.val;
+
   // Create numericInput element for formatting and numeric checks.
   const numericInput = mapp.ui.elements.numericInput(params)
 
--- a/lib/ui/elements/slider.mjs
+++ b/lib/ui/elements/slider.mjs
@@ -32,6 +32,10 @@ export default function slider(params) {
   // Set value to val if value is not defined.
   params.value ??= params.val;
 
+  params.title ??= ''
+
+  params.style ??= ''
+
   // Create numericInput element for formatting and numeric checks.
   const numericInput = mapp.ui.elements.numericInput(params)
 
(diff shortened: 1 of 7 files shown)


AlexanderGeere (changes requested):
If you use `edit:{ range: true }` and the value you are editing is null the slider doesn't show.

dbauszus-glx:
@AlexanderGeere This is by design.

A range must be bound by definition, and null can never be on this range. Null can never be max, nor min. You will need to set a default value if you want a range for a value which is null or undefined.

eg.
```js
{
  "title": "int",
  "field": "integer_field",
  "type": "integer",
  "formatterParams": null,
  "default": 0,
  "edit": {
    "range": true
  },
  "min": -100000,
  "max": 10000,
  "prefix": "£",
  "filter": {
    "type": "integer"
  }
}
```

That being said. The logic was bugged. I just did a commit to ensure that the range condition regards the newValue and value 0 by checking whether this is not NaN.

dbauszus-glx (approved):
This is working for me on the numeric entry as well as the opacitySlider.

AlexanderGeere (approved):

AlexanderGeere:
Changed the value check to use `??` (nullish operator) to avoid `0 || undefined` as this will evaluate to `undefined`.

RobAndrewHurst:
@simon-leech @dbauszus-glx 

Why do we assign 'a' as the data-id identifier for the slider?

`params.data_id = 'a'`

dbauszus-glx:
@RobAndrewHurst Good question. This is required for the numericInput element oninput method.

Here the sliderElement style sets the style property --a which is required to calculate the track styling.

```js
// Set styling param for the track.
params.sliderElement.style.setProperty(`--${e.target.dataset.id}`, params.newValue)
```
In the css
```css
    width: calc((var(--a) - var(--min)) / var(--dif) * (100% - 10px));
```

I've updated the params assignment here. https://github.com/GEOLYTIX/xyz/pull/1419/commits/7b335a82f62a666d1ba3050c6540e26cdbc1d820

RobAndrewHurst (approved):
Nice! I am happy with this. 

I added a lib/ui/elements/slider.test.mjs module to test the creation of the slider.

There are still a couple of things that I want to figure out how to simulate. That being events on elements like this. 

I could dispatch an event to trigger the oninput for the slider, but the event itself was lacking all sorts of crucial elements (callbacks, values etc). Something to try tackle in another pr though.

Issue #1421, opened 2024-08-19
modal -> dialog (No accessible layers in locale.)


RobAndrewHurst:
## Description

The modal displayed when a user does not have access to any locales because they lack the roles needs to be a dialog as we will need access to potential controls, like admin or logout.

![image](https://github.com/user-attachments/assets/07504c11-7f22-44b7-9a3a-d71e67450fc9)

This screen shot is an example of a user being restricted. 

- 🐛 Bug

PR #1422, opened 2024-08-19
Default view dialog fixes

RobAndrewHurst:
## Description

The no_locales and no_layers dialogs on the default view were displaying as full modals which would prevent the user from accessing any other controls in the mapview. (admin or logout).

Fixes.
1. Gave the two dialogs a target, because if a dialog does not have a target it is a modal.
2. I removed the 'px' suffix because a user may want to set the position with other units other than 'px'.
3. Fixed the 'helpDialog' to use 'px' in the config.
## GitHub Issue
https://github.com/GEOLYTIX/xyz/issues/1421

## Type of Change

### DIFF
--- a/lib/ui/elements/dialog.mjs
+++ b/lib/ui/elements/dialog.mjs
@@ -41,18 +41,18 @@ const dialog = mapp.ui.elements.dialog({
 */
 
 export default function dialog(dialog) {
-  
+
   if (!dialog.target) {
 
     // The dialog must be modal if no target is provided.
     dialog.modal = true;
 
-  // Ensure the target is an HTMLElement before proceeding
+    // Ensure the target is an HTMLElement before proceeding
   } else if (!(dialog.target instanceof HTMLElement)) return;
 
   // Close existing modal.
   document.querySelector('dialog.modal')?.close()
-  
+
   dialog.closeBtn ??= dialog.close && mapp.utils.html.node`
     <button
       data-id=close
@@ -63,7 +63,7 @@ export default function dialog(dialog) {
 
   // If headerDrag is true - add the headerDrag class to the dialog header
   dialog.headerDrag && (dialog.header = mapp.utils.html.node`<header class="headerDrag">${dialog.header}</header>`);
-  
+
   let classList = `${dialog.modal ? 'modal' : 'dialog'} ${dialog.class || ''}`
 
   dialog.node = mapp.utils.html.node`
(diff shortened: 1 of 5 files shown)


dbauszus-glx (changes requested):
Defaults should not be in pixel. `em` or event percent are preferred.

RobAndrewHurst:
> Defaults should not be in pixel. `em` or event percent are preferred.

Is this in reference to the helpDialog module?

dbauszus-glx:
> > Defaults should not be in pixel. `em` or event percent are preferred.
> 
> Is this in reference to the helpDialog module?

Yes. It's this one.
image

RobAndrewHurst:
@dbauszus-glx This has been fixed!

AlexanderGeere (approved):
Looks good!

dbauszus-glx (approved):

dbauszus-glx (changes requested):
Should the dialog be on the mapview element or the document.body?

There should be a little bit of margin to the element as this looks somewhat squashed.

image

RobAndrewHurst:
> Should the dialog be on the mapview element or the document.body?
> 
> There should be a little bit of margin to the element as this looks somewhat squashed.
> 
> image

image

I have no made it part of the body, and added in a z-index style on the dialogs to ensure they are on top of everything because a dialog element won't have that but a Modal will.

RobAndrewHurst:
@dbauszus-glx Do we want the option to have a dialog that you cannot move? Or are we starting to call a spade something else?

dbauszus-glx:
> @dbauszus-glx Do we want the option to have a dialog that you cannot move? Or are we starting to call a spade something else?

Don't think that's needed. A modal doesn't move and should be used in that case.

RobAndrewHurst:
> > @dbauszus-glx Do we want the option to have a dialog that you cannot move? Or are we starting to call a spade something else?
> 
> Don't think that's needed. A modal doesn't move and should be used in that case.

This is ready for a re-review then.

AlexanderGeere (approved):
Dialog could probably open a little to the right, but otherwise looks good to me

dbauszus-glx (approved):
I have updated the margins and css to have the dialog at a max-width of 50%, left 25%, and slightly top: 40%.

The dialog is more prominent if slightly top, centre, and not full width.

image

RobAndrewHurst:
I am happy with this!

PR #1424, opened 2024-08-20
entry without node in valChangeMethod on a new location.

dbauszus-glx:
A new location is immediately editable.

The valChange method called to check values in a new location will fail because the entry.node is undefined.


RobAndrewHurst:
This looks all good to me. I just want to write a test for the module first and then it can go in.

RobAndrewHurst:
This is quite complex to test the location view. I will need more time to write a test for this.

RobAndrewHurst (approved):

PR #1425, opened 2024-08-21
Assign roles onto update_user object

RobAndrewHurst:
## Description

The Roles were not being assigned to the `update_user` object on the mod/user/update function.

The assign will happen regardless if the field is a Roles or not as we will always want to assign that value.


dbauszus-glx (approved):
lgtm

simon-leech (approved):
Forgot to press approve earlier after testing this one, but all good 👍

PR #1426, opened 2024-08-21
Remove redundant featureFormats.cluster method

dbauszus-glx:
The undocumented featureFormats.cluster() method has been removed. The method is not referenced in the mapp library, nor any plugin script.

The origin or legacy use of the method are not known.


simon-leech (approved):
I see no reference to this also in both XYZ or our private repositories that use XYZ - so all good to remove and tidy up 👍

PR #1427, opened 2024-08-21
Update JSDocs definition in infoj module; Location typedef

dbauszus-glx:
Update `utils/paramString` module/method documentation and added as `@requires` to info module.

Update `utils/xhr` module/method documentation and added as `@requires` to info module.

Update `utils/merge` module/method documentation and added as `@requires` to info module.

Updated `infoj-entry` typedef as entry param for infoj module methods.

Changed `@param` to `@property` for infoj module method param which are properties.

Reviewed and added missing property defintions, updated property types.

Checked whether properties are optional.

Added missing function tag.

Moved function descriptions into `@description` tag.

Added `location` typedef.

This typedef is a stump which includes a function property with a `@link` inline link tag to a innerMember of a module.

```
{@link module:/location/decorate~flyTo}
```

image

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -128,6 +128,11 @@ A mapp-layer object is a decorated JSON layer object which has been added to a m
 */
 ```
 
+### @link
+The link tag creates an inline link element which can be used to create a link inline the description.
+
+The link can reference a module and it's inline members like methods. `{@link module:/location/decorate~flyTo}`
+
 ### @example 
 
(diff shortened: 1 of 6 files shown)


simon-leech (approved):
Looks good to me - like the link tag so you can easily navigate all the modules and methods! :)

RobAndrewHurst (approved):
Nice one! Keen to put more of this into practice

AlexanderGeere (approved):
Nice one!

PR #1428, opened 2024-08-22
Update user - only add `field` if present in params

RobAndrewHurst:
## Description

The params.field only needs to be added to the user_update object if its present in the params. On a request with just a body, it will not have this param.

## How to test this? 
Please run the local tests with the tests settings. (please note that we will need to turn the PUBLIC key to PRIVATE for this to work.) You will also need to be signed in as an admin user


simon-leech (approved):

dbauszus-glx (approved):
Working for me as expected.

image

PR #1429, opened 2024-08-22
Fix language differences

RobAndrewHurst:
## Description

There was a discrepancy between Spanish, Chinese Simplified and Traditional keys being used in the different dictionary objects being used throughout the application. Please make sure that if you add a language that there are base keys, and if there is already a language defined that you are using the correct key.

e.g of issue
Spanish base key = esp 
In other modules = es

## Type of Change

## How have you tested this? 
You can test this by running the local tests and you should see the dictionary tests pass.


dbauszus-glx (approved):
¡Chido!

image

PR #1386, opened 2024-07-18
entry type:layer to replace mvt_clone

dbauszus-glx:
The mvt_clone type entry had several bugs and applied an unnecessary complex logic.

The mvt_clone module has been removed and the mvt_clone entry method has been deprecated. The mvt_clone method will warn and return the new layer entry method.

A `layer.featureSet` filter has been to the featureStyle method. Feature not included in the featureSet will not be styled.

The `layer/format/mvt` module documentation has been revised. The clone logic has been removed.

The `type:layer` entry method assigns a Key property which is the orignal layer key required to get a location.

The `highlight` interaction `getFeature` method has been updated to recognise the layer.Key property.

related fixes:

Removed `layer/featureStyle` module function params.

Added missing `@requires` to `/ui/locations/entries` module.

Fixed `/ui/locations/entries/numeric` declaration.

Fixed `/layer/changeEnd` declaration.

Removed deprecated ui/locations/entries/default module.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206602587669819

PR #1431, opened 2024-08-22
Create ui.elements.layerStyle module

dbauszus-glx:
The `ui/layer/panel/style` module was a cognitive nightmare. Think templates in functions in templates in functions in templates, with if conditions gallore.

There was also a serious bug with the `changeEnd` event for labels potentially being assigned multiple times.

The `layerStyle` module has been added to the `ui.elements`.

A `layer.style.view` can be requested with selected `style.elements[]`. By default these are all elements in the same order created by the panel/drawer method.

The `layer.style.view` can be used to represent a collection of style control elements in a custom application view or in a dialog.

This must be tested for this PR to be accepted.

The different element methods that make up the style.view have been split and are exported as methods from the layerStyle module.

It should be possible to call these individually if required.

---
  - https://app.asana.com/0/0/1208157154358394


dbauszus-glx:
Put into draft until icon_scaling is merged in order to allow for icon_scaling elements to be included.

dbauszus-glx:
The icon_scaling method has now been added to the layerStyle module.

simon-leech (changes requested):
I have re-ordered the default style element order so it is the same as on the current main branch. 

``` js 
 'labels',
    'label',
    'hovers',
    'hover',
    'icon_scaling',
    'themes',
    'theme',
```

The icon scaling checkbox doesn't seem to update the map correctly. It is only updated when I actually highlight a certain feature, instead of reloading the whole layer on the map. 


Testing the elements in a custom view using this code 

``` js 
   const layer_a= Mapview.layers.layer_a;

    const layer_a_themes= mapp.ui.elements.layerStyle.themes(layer_a)

    console.log(layer_a_themes);
   
    // Add the layer toggle to the map
    sixthTop.querySelector('.map').append(layer_a_themes);

```

The element returned from `layerStyle.themes` is an array of HTML Elements. Can we return it within a `div` so that it can be directed appended to other elements?

simon-leech:
Working well for me testing in both default mapp and also custom views. 

Separate issue to resolve about the return from the `addLayer` method ensuring this is the decorated layers (have been styleParsed). 

We need about how we refresh a legend that isn't part of the `style` panel - e.g. in a custom view the legend is likely shown in a different `div` HTML Element.

dbauszus-glx:
@simon-leech Issue for the return from addLayer https://github.com/GEOLYTIX/xyz/issues/1457

Update legend can be added in a separate PR once we get to play with this which is required for https://github.com/GEOLYTIX/xyz/pull/1386

RobAndrewHurst (changes requested):
image
When looking at the different elements I noticed that if there is a hover then we have a checkbox with a data-id: **hoverCheckBox**. But if we have a hovers object which produces a dropdown, the data-id for the dropdown is **dropdown**. Should it not have a data-id similar to the hoverCheckBox element?

Maybe I am misunderstanding the purpose of how we use these data-id attributes. Are they there to be used for unique identification without using the id attribute? The typical use case that I find is that it's used as a layer to store data on an element.

image

Should we maybe make these drop downs and other elements follow suit to check boxes and other elements in how they use the data-id attribute?

dbauszus-glx:
@RobAndrewHurst This should now be resolved.

The styleElements assign data_id to all elements.

The slider and slider_ab elements will assign the data_id to group_id which is the data_id of the element group.

image

lib/ui/elements/layerStyle.mjs
```diff
@@ -0,0 +1,532 @@
...
+@property {layer-style} layer.style The layer style configuration.
+@property {Array} [style.elements] Array of method keys to order elements in the layer.style.view.
+
+@returns {HTMLElement} The layer.style.view element.
+*/
+function panel(layer) {
+
+  if (!layer.style) return;
+
+  layer.style.elements ??= Object.keys(layer.style)
```
RobAndrewHurst, line 57: Found an issue here, because it looks for object keys if you provide a property like opacitySlider which isn't an object but a boolean it doesn't assign the key to the layer.style.elements object which wont trigger th function on the panel.

RobAndrewHurst (approved):
Happy with this! Please have a look at the tests that I have written, and add to them as you feel fit!

simon-leech (approved):
Tested this on default mapp and custom views and working as expected :)

PR #1432, opened 2024-08-22
Add src test method to workspace module.

dbauszus-glx:
The test method will cache the workspace, get every locale, and every template.

The template error logged by the getTemplate method are returned with the HTTP response.

image


simon-leech (approved):
I added a new test to the `test_view` that will call this method and return all invalid paths as errors. 

All working for me :)

RobAndrewHurst (approved):

PR #1433, opened 2024-08-27
Test Workspace rework

RobAndrewHurst:
## Description

I have refactored the workspace because of the #1409 changes.
Just makes it easier and quicker merge and review changes to the test workspace.

- ✅ Testing

## How to test this?
Run the local tests and ensure that everything passes


dbauszus-glx:
Looks good. Is this the expected output for the test launch default view?

image

simon-leech (approved):
All tests passing for me - all good :)

RobAndrewHurst:
> Looks good. Is this the expected output for the test launch default view?
> 
> image

Are your ACL settings set to PUBLIC or PRIVATE? And are you an admin user on the ACL?

I want to consider adding in logic to only run these tests if the instance is PRIVATE

dbauszus-glx:
I was on PUBLIC. Could it be possible to run the frontend tests with a spoofed user object and no ACL?

RobAndrewHurst:
@dbauszus-glx I have added in a check if there is no user on the `mapp` object then don't run the user update test.

RobAndrewHurst:
> I was on PUBLIC. Could it be possible to run the frontend tests with a spoofed user object and no ACL?

So if there is no ACL and its set to public we should maybe just skip those api tests

PR #1437, opened 2024-08-27
Create proxy module for mapp.dictionary to check for mapp.language and lookup key

dbauszus-glx:
The dictionary proxy has been moved from the mapp module script to a module in the dictionaries directory.

The getter proxy will now check whether the language exists in the dictionaries, and whether the key exists in either the language or default english dictionary.

The key will be returned if no lookup is available.

The dictionary proxy must not crash.


simon-leech (changes requested):
I tested by swapping my language to Chinese. I correctly get the Chinese translation where available and otherwise get the default of English. 

I tested by swapping my language to 'TEST' which is obviously bogus. I get no crash and just see English. 

However I see 43 console warnings , I assume this is one for every key that's not available? Do we want to just have a single warning instead of filling the console with the same warning repeatedly? 
image

dbauszus-glx:
@simon-leech We can set the mapp.language to english in the default view if the issue is the number of warnings. But we shouldn't mess with debouncing warnings or altering the user/mapp language inside the dictionary proxy getter.

simon-leech (approved):

dbauszus-glx:
I added a check for the user language to the default view to reduce the number of warnings.

simon-leech (approved):

simon-leech:
Still to do before merging - write a test to test that an unknown language warns and sets to English. The other language tests are passing so just need to do this, will have a look today

simon-leech:
I just pushed a test for checking that a 'language' of "TEST" is correctly updated to "en" and does not crash - this is passing so the process is working.

Please can someone review this?

RobAndrewHurst (approved):
Happy with this!

PR #1446, opened 2024-08-29
User update req.body check for Vercel deployments ▲

RobAndrewHurst:
## Description

On deployed instances the req.body always defaults to '{}'. Therefore the check that was applied on the user_update assign for the req.body would not resolve correctly because it would be truthy and then provide an undefined email address to the update query.

```js
const update_user = (Object.keys(req.body || {}).length === 0 || req.body === undefined)
    ? { email: req.params.email }
    : req.body;
```

This is solved by adding in a check on if there are no keys in the object then it would resolve to falsy assigning the email from the params.

We also check if there is an email property on the update_user before executing the query. As if you provide a body with keys it will pass the initial check but still fail the query.

- 🐛 Bug fix (non-breaking change which fixes an issue)

**Note**
You can also test this locally with the rewrites of a deployed instance by running `vercel dev`. It will ask you some setup questions but just point them to the instance.

### DIFF
--- a/mod/user/update.js
+++ b/mod/user/update.js
@@ -104,7 +104,8 @@ module.exports = async function update(req, res) {
     return res.status(400).send('Invalid key in user object for SQL update.')
   }
 
-  if (!update_user.email) { return res.status(500).send('Email address required.') }
+  //If the client has provided a request with a body/param that does not have an email we will return a 400
+  if (!update_user.email) { return res.status(400).send('Email address required.') }
 
   const properties = []
   const substitutes = [update_user.email]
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
I think another thing to consider is we should return an error to the user if a user email was not provided in the params or the body. Because actually with this check if you provide a body with no email, but other keys the query will still fail.

simon-leech:
@RobAndrewHurst  Agreed - do you want to push that before I review properly?

RobAndrewHurst:
@simon-leech Done.

simon-leech (approved):
All working with me - tested both locally and on a deployed instance and working across all fields and roles updates 👍

dbauszus-glx (approved):
Deployed and tested here. https://geolytix.dev/geodata/api/user/admin

PR #1447, opened 2024-08-30
Missing HelpDialog for Editing a Geometry Editing

simon-leech:
It was noticed that when you are editing an `infoj` `geometry` `entry` - you are not provided with a `helpDialog` , when you do get this for drawing. 

This PR simply adds a `helpDialog` for this. 

## Type of Change

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -11,6 +11,10 @@ mapp.utils.merge(mapp.dictionaries, {
     delete_geometry: 'Delete Geometry',
     modify_geometry: 'Modify Geometry',
     edit_dialog_title: 'Editing Instructions',
+    edit_dialog_cancel_drawing: 'Cancel Drawing - Click on the cancel button to cancel the drawing.',
+    edit_dialog_remove_vertex: 'Remove Vertex - Click on a vertex to remove it.',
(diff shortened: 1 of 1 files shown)


RobAndrewHurst (approved):
Looks good to me!

RobAndrewHurst:
Actually do we want the dialog to be optional?

simon-leech:
> Actually do we want the dialog to be optional?

Isn't the current helpDialog on drawing interaction always shown, so I followed that pattern.
 
I don't think it should be optional as surely its helpful to see it for reference to the available options?

RobAndrewHurst:
We might want to consider this for an issue, and another PR

simon-leech:
> We might want to consider this for an issue, and another PR

Yeah good point on custom views. 

I think the problem is I think it should be default true, but you can turn it off if required - in case not wanted. 
But in general, I think its an important feature to show users 😅

Agreed maybe we want an issue with RFC

RobAndrewHurst (approved):

dbauszus-glx (changes requested):
The dialog does not reflect modifying geometry which is different from drawing geometry. 
image

simon-leech:
image
Pushed a commit to update the text to be relevant to geometry editing

dbauszus-glx:
Looking good now.
image

dbauszus-glx (approved):

PR #1451, opened 2024-09-02
numericChecks method for slider_ab element.

dbauszus-glx:
It must be possible to assign a custom numericChecks method to a numericInput element.

The slider_ab element has two numeric input elements `a` & `b`.

The a.newValue must not be more than the b.newValue.

The b.newValue must not be less than the a.newValue.


AlexanderGeere (approved):
An elegant solution!

simon-leech (approved):
All working for me :)

PR #1453, opened 2024-09-03
Location mock

RobAndrewHurst:
## Description

We need the ability to pass in a custom template to the location.get() function. This can be used in different use cases. The main one at the moment is used to mock locations in the testing environment. 

## Type of Change

- ✅ Testing

## How have you tested this? 
Run the local tests and Look for the location: getTest.

## Testing Checklist 
Note that the user tests will fail with a public instance. This is something that still needs to be resolved by spoofing a user.

- ✅ New Tests Added

### DIFF
--- a/lib/location/get.mjs
+++ b/lib/location/get.mjs
@@ -72,13 +72,15 @@ export async function getInfoj(location) {
     location.layer = await mapp.utils.xhr(`${mapp.host}/api/workspace/layer?locale=${location.locale}&layer=${location.layer}`)
   }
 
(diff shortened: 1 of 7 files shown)


AlexanderGeere (approved):
Nice one!

simon-leech (approved):
I just pushed one more test to ensure the location has a layer object assigned to it. 

Looking good to me :)

dbauszus-glx (approved):
Lovely work. 
image

PR #1454, opened 2024-09-03
Only update dataview with activateUpdate or no data.

dbauszus-glx:
By default the dataview update method is triggered from the tab activate method.

With the `keepDataOnActivate` flag the update method will be shortcircuited in the tabview tab.activate() method.

The dataview can persist with this flag.

### DIFF
--- a/lib/ui/Tabview.mjs
+++ b/lib/ui/Tabview.mjs
@@ -84,13 +84,12 @@ function addTab(entry) {
 
     if (entry.update instanceof Function) {
 
-      if (entry.data && entry.keepDataOnActivate) {
+      
+      if (!entry.data || entry.activateUpdate) {
 
-        // Do not update dataview entry with data.
-        return;
+        // Call dataview update method if data is falsy or activateUpdate flag is set.
+        entry.update()
       }
-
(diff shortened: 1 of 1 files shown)


AlexanderGeere (commented):
This flag should maybe be the default? `entry.keepDataonActivate ?? true`

dbauszus-glx:
@AlexanderGeere The flag would be different ie. `updateOnActivate`. This would be more elegant but the problem is that the current behaviour will be changed. Should be discussed which way is preferred.

simon-leech:
This is working well for me. 

I also think it should be the default to apply this so data is retained on activate of a tabview - and only if specified update the data every time the activate method is called.

dbauszus-glx:
I reversed the logic to use the `activateUpdate` to update on activate if the entry data is not falsy.

simon-leech (approved):

RobAndrewHurst (approved):
I am happy with this change. I don't have time as of yet to write a test for the Tabview. 

We need to get to that at some stage. :( 

Or maybe someone can get to this while I am on leave?

PR #1455, opened 2024-09-03
No obj.key template lookup and clone merge.

dbauszus-glx:
Merging an object into a template will cause an array aggregation of objects if the template is not cloned for the merge.

This may also cause cross pollination as objects merged from the template will be added to objects merged into the same template from another object.

[Screencast from 03-09-24 10:05:50.webm](https://github.com/user-attachments/assets/e8185704-8623-4f79-97e4-13074a6395ed)

Template lookup for object key is fraught with danger. There is no control to prevent a wrong template being looked up, and the warning if a template is not found cannot be stored because many object keys may not have a matching template.

### DIFF
--- a/mod/workspace/mergeTemplates.js
+++ b/mod/workspace/mergeTemplates.js
@@ -47,28 +47,29 @@ module.exports = async function mergeTemplates(obj) {
   // Cache workspace in module scope for template assignment.
   workspace = await workspaceCache()
 
-  const template = await getTemplate(obj.template || obj.key)
+  // The object has an implicit template to merge into.
+  if (obj.template) {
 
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
This is not working as expected. 

I tested with a bogus template assigned to a layer. 
I see no error or warning about this, but the layers beneath this one are not loaded in - it just stops without error or warning?

dbauszus-glx:
@simon-leech can we see the test workspace for this?

RobAndrewHurst (approved):
Happy with this now!

simon-leech (approved):
I've built the auto-update method in ignite and tested this and i'm happy

PR #1456, opened 2024-09-03
location.table may be undefined; location.get and remove tests; docum…

dbauszus-glx:
A location should not require a table property. The property can be undefined and has been added as optional with the getTemplate string property to the location typedef.

Location get tests should check on instanceof Function or Object rather than typeof string checks or mere existence (!!).

Instead of testing whether location.remove is a function it's better to call this for testing.

I check the hook and whether this is removed.

I add a removeCallback method which removes itself for testing.

Added documentation for remove method.

Removed location decorate test as this is tested from location.get.

The decorate method is only called from gazetteer outside the location.get method. Instead of running nested tests the gazetteer should be tested itself which will then ensure that the decorate test passes if called from any occurrence in the mapp script.


RobAndrewHurst (approved):

RobAndrewHurst:
Nice one!!

PR #1461, opened 2024-09-05
Enable mapviewControls in locale

dbauszus-glx:
Openlayers controls must be created from the ol.controls library.

This prevents controls being defined as string in the workspace.

In addition of providing an array of control objects in the mapview creation eg `controls:[new ol.control.Zoom()]` it should be possible to define an array of string values for the controls in the locale as mapviewControls.

This will allow to move remove the zoom control plugins from the default `syncPlugins` and add the OL zoom control to the mapview.Map.

```js
    "mapviewControls": ["Foo", "Zoom"],
    "syncPlugins": ["admin", "login"],
```

Control methods which are not a property of the ol.controls object are ignored.

### DIFF
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -119,6 +119,11 @@ export default function decorate(mapview) {
     mapview.extent = ol.proj.transformExtent([-180, -90, 180, 90], 'EPSG:4326', `EPSG:${mapview.srid}`);
   }
 
+  // Assign ol controls for mapview from locale.
+  mapview.controls ??= mapview.locale.mapviewControls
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):
The zoom controls are behind the button column, but when using `app_menu` this works well. All tests still passing as well

RobAndrewHurst (approved):
Looks good!

I added in a test to check on a locale that has mapviewControls defined that we can see the controls being added into the mapview object.

cityremade (approved):
Looks good to me. Position is dependent on css yet the top-left position is a reasonable default choice.  Easily adjusted in custom views.

PR #1462, opened 2024-09-05
Only check obj key template if not implicit

dbauszus-glx:
The workspace should not be checked for a template matching the object key if there is an implicit template property on the object.

This is to prevent a false flag exception.


RobAndrewHurst (approved):
This looks all good to me! 

I have re-run the tests and everything looks normal!

PR #1464, opened 2024-09-05
Pills - minor fix;

cityremade:
`pills` not assigned to parent params. Would error when layer was initially filtered on the same field. Second fix - when using pills keep placeholder in the dropdown.

## Description

Layer with a filter initially set in the workspace. When filtering on the same field by pills dropdown 
access to current `in` filter would fail: 

was: params.pills && pills.add(entry.title)
should be: params.pills && params.pills.add(entry.title)

Additional change: when using pills keep the dropdown placeholder. Join values only when pills not in use.

## GitHub Issue

## Type of Change

## How have you tested this? 
Please describe how to test this change and how you verified it worked. 

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -92,7 +92,7 @@ export default (params) => {
       params.selectedOptions.add(entry.option)
 
       // create pill
-      params.pills && pills.add(entry.title)
+      params.pills && params.pills.add(entry.title)
     }
 
     return li
@@ -158,7 +158,7 @@ export default (params) => {
       data-id=${params.data_id || 'dropdown'}
       class="dropdown">
         <div class="head" onclick=${params.headerOnClick}>
-          <span data-id=header-span>${headerSpan}</span>
+          <span data-id=header-span>${params.pills ? params.placeholder : headerSpan}</span>
           <div class="icon"></div>
         </div>
         <ul>${ul}`;


dbauszus-glx (changes requested):
I can see this working however. If the dropdown module is touched it wants to be reviewed and documented. There is not even a jsdocs module stump. The dropdown method itself is undocumented with hardly any useful comments.

Using conditional prevents a bug where you check whether pills are a property of params before calling pills directly. You can see that the conditional was already used in the else condition to remove.
image

Please remove the duplicate logic in the template literal. the headerSpan assignment has a complex logic and was therefore taken out of the literal. Adding additional conditional logic in regards to the headerSpan, params.pills, and params.placeholder which is already in the condition outside the literal makes this impossible to debug/understand.

headerSpan should be defined outside the literal and then assigned as `${headerSpan}` inside the literal

image
 
I define a current `in` filter with a single value which I see in the dropdown.

```
        "filter": {
          "current": {
            "id": {
              "in": [1]
            }
          }
        },
```

image

However, without the filter in the layer JSON I would see two values.

image

RobAndrewHurst (approved):
This all looks good to me! 

I have added in a pills test to test creating a pills component with 0 params and then using the add/remove function to add pills to the set.

AlexanderGeere (approved):
Nice looks good to me. new tests look good as well.

dbauszus-glx (changes requested):
Review in progress; See comments.

dbauszus-glx:
Parameter properties for the `filter_in` method docs have been changed from param to property.

The current filter for the filter.field must be excluded from the distinct_values query.

The `ui/elements/dropdown` module was missing the module block which is required to show the module documentation.

The dropdown method returned from the dropdown module has been named.

While technically an object, the dropdown method returns an HTMLElement.

Assigning the li elements to the ul variable used within nested methods of the parent method will cause confusion later down the road. It much better to assign the li elements to the li_elements variable which is only required in the construction of the params.ul element group. The params.ul element is used in the pills removeCallback method to find a li element. With good variable naming a log of the comments are redundant. params.ul for example is much better understandable than using e.target.nextElementSibling where the e.target is the header element and assuming that the ul element is the next element sibling.

dbauszus-glx:
The headerOnClick method has been unnested from the dropdown method and cleaned up.

Assigning variables which are used only once to a name just for them to be used in a single calculation which result is stored in a different variable adds unnecessary cognitive overload.

The classList toggle returns true when a class is added. This is much more concise than checking whethe the class exists and then add or remove the class based on whether the class exists.

```js
  const active = dropdown.classList.toggle('active');

  if (active === false) return;
```

dbauszus-glx (approved):

dbauszus-glx:
create ulElements method with nested liOnClick method; updated docs.

PR #1467, opened 2024-09-06
mapviewPromise method for async mapview decoration; svg_templates util

dbauszus-glx:
The `svg_templates` plugin has been moved to mapp.utils.svgTemplates.

The svg_templates plugin method has been deprecated and will simply call mapp.utils.svgTemplates to the same effect.

`mapp.utils.svgTemplates` ensures that individual svg_templates will only be loaded once.

A mapview being decorated with svg_templates in a locale and no mapp.utils.svgSymbols.templates will return the mapviewPromise method.

The async `mapviewPromise` which resolves to the mapview can make the mapview decorator async if needed. This is for legacy purposes. The mapview decorator can still by synchronous if no plugins or svg_templates need to be loaded.

The `loadPlugins` flag is no longer required. If set the `mapview.local.syncPlugins` are assigned to the `mapview.syncPlugins` which can also be directly defined as shown in the _default script.


lib/utils/svgTemplates.mjs
```diff
@@ -0,0 +1,43 @@
...
+The svgTemplates method will resolve once all templates have been parsed and loaded.
+
+@param {object} templates An object with svg_template properties to load.
+
+@returns {Promise<void>}
+*/
+export default async function svgTemplates(templates){
+
+  if (!Object.keys(templates)) return;
+
```
RobAndrewHurst, line 21: This is going to be an issue with undefined or null templates.

Going to add an additional check if templates are undefined or null.

```js
  if (!templates || !Object.keys(templates).length) return;
```

dbauszus-glx: Not sure I fully understand the issue but happy to expand the conditional check.

RobAndrewHurst: So when a workspace has sync plugins but no svgTemplates in the workspace it would hit the svgTemplates function with (undefined/null) because it's not part of the workspace. So just like we have that check in the loadPlugins utility we will also expand the check :) If not clear I can show can example tomorrow in the stand up :)

RobAndrewHurst (approved):
Nice! 

All happy with this!

And included tests to add new svgs and then test trying to added new svgs with the same key.

lib/utils/svgTemplates.mjs
dbauszus-glx: Good catch.

simon-leech (approved):
This is working for me - tested in default MAPP and also in a custom view that calls the svgTemplate method and still working as expected.

AlexanderGeere (approved):

PR #1468, opened 2024-09-06
addlayer to allow for single layer to be added.

dbauszus-glx:
The addLayer method now allows for a single layer to be added or an array of JSON layer.

A default key consisting of the format + index will be assigned to layer with a valid format.

The layer will be replaced in the layers array in order to return an array of decorated layer.

The addLLayer test has been updated to test the method.

image


RobAndrewHurst (approved):
Looks all good to me!

Thanks for adding in this test

PR #1471, opened 2024-09-10
Set icon_scaling defaults; warn for plugin config; numcheck callback.

dbauszus-glx:
The icon_scaling plugin will be deprecated.

The styleParser method will warn and assign layer.icon_scaling as layer.style.icon_scaling.

The minScale and maxScale defaults will only be assigned if not explicit in config.

### DIFF
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -187,6 +187,13 @@ export default function styleParser(layer) {
     if (layer.style.zIndex) {
       console.warn(`Layer: ${layer.key}, layer.style.zIndex has been deprecated. Use layer.zIndex instead.`);
     }
+
+    if (layer.icon_scaling) {
+
+      console.warn(`Layer: ${layer.key}, layer.icon_scaling has been assigned to layer.style.icon_scaling`)
+
+      layer.style.icon_scaling ??= layer.icon_scaling
+    }
   }
 
   function parseTheme(theme, layer) {
--- a/lib/ui/elements/layerStyle.mjs
+++ b/lib/ui/elements/layerStyle.mjs
@@ -386,11 +386,11 @@ function themes(layer) {
   const dropdown = mapp.utils.html`<div>
     ${mapp.dictionary.layer_style_select_theme}
     ${mapp.ui.elements.dropdown({
-      data_id: 'themesDropdown',
-      placeholder: layer.style.theme.title,
-      entries,
-      callback
-    })}`
+    data_id: 'themesDropdown',
+    placeholder: layer.style.theme.title,
+    entries,
+    callback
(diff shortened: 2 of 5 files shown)


cityremade (changes requested):
I see in the scale slider callback only layer.changed event and not layer reload.
Symbol is only scaled after hover/highlight interaction which is not the desired behaviour.
Looks like layer needs to be reloaded when style is manipulated.
Is there a reason why layer.reload is not in the slider callback (line 464) ?

dbauszus-glx:
I've had another thought about this and setting the cache to null feels wrong. The cache should be optional with the feature being enabled by setting the Boolean property true. https://github.com/GEOLYTIX/xyz/pull/1490

simon-leech (changes requested):
image

If you type in the box a number that is out of range, it correctly gets the invalid CSS applied. However the value is still applied as the scaling factor. This is the same for if you type in non-numeric data.  It should not be applied if its invalid. 

Also, the icon_scaling plugin has lots of other methods which include the icon scaling but also methods for the different cluster scaling parameters. I think we need to support these too before the plugin can be deprecated, as otherwise we are removing functionality. 
``` json
  "icon_scaling":{
    "icon":{},
    "clusterScale":{},
    "logScale":{},
    "zoomInScale":{},
    "zoomOutScale":{}
  },
```

dbauszus-glx:
> image
> 
> If you type in the box a number that is out of range, it correctly gets the invalid CSS applied. However the value is still applied as the scaling factor. This is the same for if you type in non-numeric data. It should not be applied if its invalid.
> 
> Also, the icon_scaling plugin has lots of other methods which include the icon scaling but also methods for the different cluster scaling parameters. I think we need to support these too before the plugin can be deprecated, as otherwise we are removing functionality.
> 
> ```json
>   "icon_scaling":{
>     "icon":{},
>     "clusterScale":{},
>     "logScale":{},
>     "zoomInScale":{},
>     "zoomOutScale":{}
>   },
> ```

Odd; They should all be there? These were added a while ago. This PR is just to ensure pariity and add the defaults.

image

simon-leech (approved):
Weird I don't know what I did wrong but I got them working now! 


I just pushed a commit to add to the slider callbacks to check that the val is within the min/max and is a numeric before applying it - this fixes the issue I described above. 
All good with me now :)

cityremade (approved):
Excellent, icons scale now on every change numeric or range 👏

dbauszus-glx:
@simon-leech I removed the min/max check in the callbacks. The numericInput invalid condition must return before executing the callback. I applied the fix in this PR.

simon-leech (approved):

PR #1472, opened 2024-09-10
Add theme fields array to fieldsArray

dbauszus-glx:
Previously the theme.fields array was added as an array value to the fieldsSet.

This was flattened in a return but will cause duplicate entries. The reason a Set is used is to prevent duplicates.

This example theme must return a fieldsArray with a single hours_index_15 value in the fieldsArray return.

```js
{
  "title": "By Hour",
  "type": "graduated",
  "theme_slider": true,
  "field": "hours_index_15",
  "fields": [
    "hours_index_00",
    "hours_index_01",
    "hours_index_02",
    "hours_index_03",
    "hours_index_04",
    "hours_index_05",
    "hours_index_06",
    "hours_index_07",
    "hours_index_08",
    "hours_index_09",
    "hours_index_10",
    "hours_index_11",
    "hours_index_12",
    "hours_index_13",
    "hours_index_14",
    "hours_index_15",
    "hours_index_16",
    "hours_index_17",
    "hours_index_18",
    "hours_index_19",
    "hours_index_20",
    "hours_index_21",
    "hours_index_22",
    "hours_index_23"
  ]
}
```


RobAndrewHurst (approved):
Looks all good to me!

PR #1477, opened 2024-09-13
Remove geometry Layers[] from array

dbauszus-glx:
I wrongly assumed that an object property deleted will automatically be removed from an array. This is not the case.

Even though the `.L` Openlayers layer properties are deleted in the pin and geometry entry method they still remain in the location.Layers[] array.

Using the ol_uid the array is not filtered before the layer object is deleted.


AlexanderGeere (approved):
Tests pass! Looks good to me

RobAndrewHurst (approved):
Looks all good to me!

PR #1480, opened 2024-09-17
Allow for custom min max query template.

dbauszus-glx:
By default the `field_minmax` query module is used to determine the bounds for a filter if not explicit in the filter config.

```js
module.exports = `
  SELECT
    ARRAY[min(\${field}), max(\${field})] as minmax
    FROM \${table}
    WHERE true \${filter};`
```

This PR adds a condition to the `generateMinMax` method which allows to define a custom query template in the workspace and the filter entry.

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -156,11 +156,12 @@ Query the min and max values for a numeric filter field.
 @property {string} filter.field Field to filter.
 @property {numeric} [filter.min] Min bounds.
 @property {numeric} [filter.max] Max bounds.
+@property {string} [filter.minmax_query] Query template for min max values.
 */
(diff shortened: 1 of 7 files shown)


AlexanderGeere (approved):
It's pretty strict as the query has to return an array called minmax, but it works.

RobAndrewHurst:
This is looking good! Need to park it unitl Monday when i get back back into it.

RobAndrewHurst:
All happy with this now!

I have cleaned up some tests too.

The main test checks for the minmax_query changes.

And then we check two additional conditions if the min is only provided and if the max in only provided. 

We should see the explicit value override the responses value.

RobAndrewHurst (approved):

RobAndrewHurst (approved):

PR #1482, opened 2024-09-18
dataview and tabview module documentation

dbauszus-glx:
Added documentation for dataview and tabview modules.

Checks for the create function now look whether the property is undefined rather than checking with ! exclamation mark.

Add a jsonDataview module to utils which allows to test dataviews without the need to load any dataview library from through plugins.

eg. the random_num query template returns a RANDOM() number on each query.

```json
"random_num": {
  "template": "SELECT RANDOM()"
},
```
This can be tested with 2 dataviews.

```json
"dataviews": {
  "json_1": {
    "dataview": "json",
    "query": "random_num",
    "target": "tabview"
  },
  "json_2": {
    "dataview": "json",
    "query": "random_num",
    "target": "tabview"
  }
}
```

[Screencast from 2024-09-23 15-52-47.webm](https://github.com/user-attachments/assets/d172859a-468b-4914-8586-a92b0c98c549)


simon-leech (approved):
image


All good with me - added a new series of tests for the Tabview module too

lib/ui/utils/_utils.mjs
```diff
@@ -48,10 +48,13 @@ import imagePreview from './imagePreview.mjs'
...
 import resizeHandler from './resizeHandler.mjs'
 
+import json from './jsonDataview.mjs'
+
 export default {
   Chart,
   Tabulator,
   idleLogout,
   imagePreview,
+  json,
```
RobAndrewHurst, line 58: Why is this called json and not jsonDataview?

dbauszus-glx: Agree. Should be jsonDataview.

dbauszus-glx: I have updated the reference.

PR #1484, opened 2024-09-18
mapp.ui.elements.alert/ confirm methods

cityremade:
mapp.ui.elements.alert
mapp.ui.elements.confirm

This is alert / confirm elements to display information to the user.

Uses the existing ui.elements.dialog.

2 parameters added to the dialog:

closeOnClick - when set to true the dialog closes when the document is clicked anywhere over the backdrop.
onClose - callback function to execute when dialog fires a close event.

```js
mapp.ui.elements.alert({
   text: "Drivetimes have been created."
})


---
  - https://app.asana.com/0/0/1208407306090518

### CODE REVIEW

File: lib/ui/elements/alert.mjs [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+@property {string} [alert.title] Text to display in the alert header. Defaults to 'Information'.
+@property {string} [alert.text] Text to display in the alert content. 
+@returns {HTMLElement} alert The alert element.
+*/
+
+export default function alert(params) {
+
+    params.title ??= 'Information';
+
+    const css = `
```
RobAndrewHurst, line 34: We are going to want to move this css into the bundled css sheet

dbauszus-glx (changes requested):
This should make use of the mapp.ui.elements.dialog() method to create the dialog itself.

cityremade:
Changes made:

alert is now a ui.elements.dialog.

2 parameters added to the dialog:

closeOnClick - when set to true the dialog closes when the document is clicked anywhere over the backdrop.
onClose - callback function to execute when dialog fires a close event.

Both documented in the dialog module.

dbauszus-glx (changes requested):
The confirm dialog in /ui/locations/view breaks with mapp.ui.elements not being a function.

image

This can be tested by removing a location with outstanding edits.

`.mapp-says` should be replaced with a semantic class like `.alert-confirm`

The jsdoc return for function that return a promise must be promise with the text explaining how the promise resolves.

lib/ui/locations/view.mjs
simon-leech, line 1: On line 271- should this `&&confirm` also use the `mapp.ui.elements.confirm` method?

simon-leech (changes requested):

simon-leech (changes requested):
Review in progress - no commits please

simon-leech (approved):
A few changes from me but I'm very happy with this, like it a lot! 

1. I moved the Ok and Cancel Buttons in the Confirm method so Ok is first as this matches the rest of the UI - e.g. see the unsaved changes dialog, we always have Ok on the left. 
image
image
2. I turned the text 'Confirm' and 'Information', the default titles when not provided into dictionary terms instead of being hardcoded, so they can be translated in the future. 
3. I wrote a series of tests for both Alert and Confirm methods that test for passing in Title and Text, and also passing in no params. They check for the presence of the correct content and buttons. 
image

RobAndrewHurst:
I have added in a test for the dialog ui element to test the close functionality. And I noticed after invoking the close that the dialog can still be found in the DOM. I am not too sure if this is something todo with dialogs, or if we are not closing correctly.

RobAndrewHurst (changes requested):
See comment above.

cityremade:
ui.elements.dialog has onClose callback where closed dialog can be removed if needed.
Removing the element will not be a case at all times.

dbauszus-glx (changes requested):
The dialog element method decorates the params argument.

```js
const dialog = {
  content: 'foo'
}

// dialog becomes the decorated dialog object with the dialog.node being the dialog node element.
mapp.ui.elements.dialog(dialog)
```

The dialog element method should return the decorated dialog.

```js
// dialog becomes the decorated dialog with the dialog.node being the dialog node element.
const dialog = mapp.ui.elements.dialog({
  content: 'foo'
})
```

The `dialog.close()` method should call the dialog element close method as well as remove the element.

```js
dialog.close = ()=>{
  dialog.node.close()
  dialog.node.remove()
}
```

dbauszus-glx (changes requested):
The confirm dialog closes when I click on the header. This shouldn't happen.

The closeBtn shouldn't be there if it shouldn't be there. Regardless whether the styling set's the button to the same colour.

dbauszus-glx:
We try to avoid negating flags. I removed the hideCloseBtn reference. Default will be no close button. If the closeBtn is defined, ie true then the property will be mutated to a html fragment. No web api node needed here. Also no tertiary statement needed.

https://github.com/GEOLYTIX/xyz/pull/1484/commits/7bb0b498987c52ee512dae61c9cdd8dc926bb12e

dbauszus-glx:
There was a lot of unnecessary stopPropagation. There must not be a click event on the dialog element. Also the dialog must not close when clicked anywhere. That defeats the logic of a confirm dialog. A confirm dialog should be boolean. E.g. Ok=true, Cancel=false. There is no tertiary whatever/close state.

There was a lot of close onClose closeDialog action. This has been streamlined.

The closeDialog method calls the onClose event method.

The dialog.close is closeDialog method.

dialog.clkoseBtn calls closeDialog method.

dialog onclose event calls closeDialog method.

there shouldn't be any stopPropagation required.

image

dbauszus-glx (approved):
This is now working for me as expected. Please see comments.

RobAndrewHurst (approved):

PR #1485, opened 2024-09-18
highlight interaction render and getLocation

dbauszus-glx:
The default getFeature method of the highlight interaction provides a dialog for selecting cluster features. In order to use this dialog with a custom method to use for the feature/location from cluster it is required to override the getLocation method and use this method in the feature selection of the default getFeature method.

This feature is actually already provided by the location.get method but was undocumented.

This PR provides the override to highlight interaction method as well as documents the related location.get and nnearest methods and the dependency hooks module.

### DIFF
--- a/lib/hooks.mjs
+++ b/lib/hooks.mjs
@@ -1,5 +1,5 @@
 /**
-## mapp.hooks{}
+## /hooks
 
 The mapp.hooks{} module parses, updates, and stores URL parameter.
 
--- a/lib/hooks.mjs
+++ b/lib/hooks.mjs
@@ -37,12 +37,11 @@ const hooks = {
 export default hooks;
 
 /**
-## mapp.hooks.parse()
-The method parses key/value pairs of URL searchParams from the document window.location.href.
-
 @function parse
-*/
 
+@description
+The method parses key/value pairs of URL searchParams from the document window.location.href.
+*/
 function parse() {
   
   const url = new URL(window.location.href);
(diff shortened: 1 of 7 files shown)


dbauszus-glx:
Updated documentation for following modules:

/hooks
/location/get
/location/nnearest
/mapview/allFeatures
/interactions/highlight

dbauszus-glx:
A `.popup .list` css rule has been added to _mapp.css in order to get rid of some inline styling.

dbauszus-glx:
The highlight interaction module condition on the featureRender wasn't working and has been removed. The feature.setStyle method was added to prevent the whole layer rendering on the changed event. However the changed event was called anyways, and when it wasn't called after zoom the style was not set on a feature.

This can be tested by hover over icon, zoom, then leave the icon. The layer should render.

dbauszus-glx:
Added a bounce logic on location.get.

The infoj must be awaited for the location decorator. The locations set is used to check whether a location is being processed. This is to prevent getting the same location multiple times while the getInfo is in the event loop.

```js
const locations = new Set()
export async function get(location, list = location.layer.mapview.locations) {

  const mapview = location.layer.mapview;

  // A getLocation function has been assigned to the current mapview.interaction.
  if (mapview.interaction?.getLocation instanceof Function) {

    mapview.interaction?.getLocation(location)
    return;
  }

  if (!location.layer || !location.id) return;

  location.hook ??= `${location.layer.key}!${location.id}`;

  if (list[location.hook]) {

    // Remove location if already in list.
    list[location.hook].remove()
    delete list[location.hook]
    return;
  }

  // Prevent location from being requested again while not yet decorated.
  if (locations.has(location.hook)) return;
  locations.add(location.hook)

  // Assign locale key to location
  location.locale ??= mapview.locale.key

  await getInfoj(location)

  mapp.location.decorate(location)

  // Location will be removed if requested again as decorated.
  locations.delete(location.hook)
```

cityremade (approved):
Works seamlessly. I have tested getLocation, getFeature and noLocationClick overrides ✅

PR #1486, opened 2024-09-18
mapp.ui.elements.confirm

cityremade:
2/2 of planned additions

mapp.ui.elements.confirm method added to elements.

This is a confirm element to display information to the user and resolve to yes/no response.
It is a framework alternative way to using window.confirm() browser function.

At the moment css is included in the element to test it separately.
Once we're happy with it it can be moved to ui.css and unified with its sibling .alert()

### DIFF
--- a/lib/dictionaries/en.mjs
+++ b/lib/dictionaries/en.mjs
@@ -1,6 +1,7 @@
 export default {
   save: 'Save',
   cancel: 'Cancel',
+  ok: 'OK',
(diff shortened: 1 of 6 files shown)


dbauszus-glx (changes requested):
The confirm method should make use of the ui.elements.dialog method for consistent styling and params configuration.

cityremade:
Change now included in #1484

PR #1491, opened 2024-09-23
getfeature call from highlight click

dbauszus-glx:
The getFeature method must only be called from the click event method if there is a current feature.

Clicking into the mapview without a current feature should not throw an error exception.


cityremade:
from this highlight style
```json
"highlight": {
      "strokeColor": "#000000",
      "strokeWidth": 2,
      "strokeOpacity": 1
    }```

I am getting a massive stroke width. I am using a custom mapview with getFeature override.

simon-leech (approved):

Issue #1494, opened 2024-09-24
Unused Template Check


RobAndrewHurst:
## Description

We need the functionality to check if there are any current templates in a workspace.templates object that is not used by any layer.

To do this we need to mark all of the default templates with a `_orign` of `core`. `custom_templates` needs to be marked with an `_origin` of `custom_templates` & `workspace_templates` need to be marked with `workspace_templates`.

It also might be a good idea to mark the `_origin` of merged templates with `merged`. (This can be open for discussion).

Then we need to mark the different types of the templates coming from core with a `_type`: [view, query, msg, mail]
This way we can also defined other types on the explicit workspace.templates entries.
eg.

```json
"templates":{
    "api_query": {
        "src": "file://query.sql",
        "_type": "API"
    }
}
```

## Type of Issue

- ✅ New Feature
- ✅ Testing

PR #1495, opened 2024-09-24
Template rework and Unused Check

RobAndrewHurst:
## Description

I have updated the test function on the workspace mod to include a list of used templates with their frequency as well as a list of templates that are not used at all.

I also add the _core property on all default templates that get assigned to the workspace.templates object.

I also noticed that the template property on a layer when merging a template into that layer would overwrite the template on the layer. To combat this I delete the template property on the template before merging into the layer.

## GitHub Issue
#1494 
## Type of Change

- ✅ Testing


dbauszus-glx:
I added the logic to mark a template object as core to a helper function so that this can be described and doesn't have to be repeated multiple times in the workspace.templates assign.

The custom_templates should not be marked as core.

```js
  /**
  @function mark_core

  @description
  The method maps the Object.entries of the templates_object param and assigns the _core: true property to the object marking this as a system template.

  @param {Object} templates_object 
  @returns {Object} templates_object with _core: true property.
  */
  function mark_core(templates_object) {

    return Object.fromEntries(
      Object.entries(templates_object)
        .map(([key, template]) => [key, { ...template, _core: true }])
    )
  }

  // Assign default view and query templates to workspace.
  workspace.templates = {

    ...mark_core(view_templates),
    ...mark_core(mail_templates),
    ...mark_core(msg_templates),
    ...mark_core(query_templates),

    ...custom_templates,

    // Default templates can be overridden by assigning a template with the same key.
    ...workspace.templates
  }
```

mod/workspace/cache.js
```diff
@@ -95,25 +95,33 @@ async function cacheWorkspace() {
...
+  /**
+  @function mark_core
+
+  @description
+  The method maps the Object.entries of the templates_object param and assigns the _core: true property to the object marking this as a system template.
+
+  @param {Object} templates_object 
+  @returns {Object} templates_object with _core: true property.
+  */
+  function mark_core(templates_object) {
```
RobAndrewHurst, line 107: Nice, good use of a method to mark them.

RobAndrewHurst:
image
The templates that are core are not being marked as core now.

RobAndrewHurst:
> image The templates that are core are not being marked as core now.

Oh wait, because they can be. Because these are coming from custom templates.

dbauszus-glx (dismissed):
The checkTemplate methods needs to be documented as it isn't clear what this is doing.

image

The check fails on this test workspace. OSM template is used.

```json
{
  "templates": {
    "OSM": {
      "test": true
    }
  },
  "locale": {
    "layers": {
      "OSM": {
        "template": "OSM",
        "display": true,
        "format": "tiles",
        "attribution": {
        }
      }
    }
  }
}
```

image

dbauszus-glx:
@RobAndrewHurst It is possible to overwrite _core templates like with custom templates. But the overwritten ones should not be marked as core. The _core flag is only to indicate system templates which come from the templates directory itself.

RobAndrewHurst:
> @RobAndrewHurst It is possible to overwrite _core templates like with custom templates. But the overwritten ones should not be marked as core. The _core flag is only to indicate system templates which come from the templates directory itself.

Yeah I agree with this.

RobAndrewHurst:
> @RobAndrewHurst It is possible to overwrite _core templates like with custom templates. But the overwritten ones should not be marked as core. The _core flag is only to indicate system templates which come from the templates directory itself.

I have added in different types on the entries added to the workspace.templates.

The core templates will have 'core'. custom_templates will be 'custom_templates' & templates that come from the workspace itself be it from a layer itself or just in the workspace.templates object itself will have 'workspace_templates'.

I then filter out the workspace_templates as those are the templates only required for the testing, but with these additional '_type' properties we can do further drilling. 

Please have a look, and let me know if you agree with this or if we can apply this can of logic in a more elegant way.

dbauszus-glx:
> > @RobAndrewHurst It is possible to overwrite _core templates like with custom templates. But the overwritten ones should not be marked as core. The _core flag is only to indicate system templates which come from the templates directory itself.
> 
> I have added in different types on the entries added to the workspace.templates.
> 
> The core templates will have 'core'. custom_templates will be 'custom_templates' & templates that come from the workspace itself be it from a layer itself or just in the workspace.templates object itself will have 'workspace_templates'.
> 
> I then filter out the workspace_templates as those are the templates only required for the testing, but with these additional '_type' properties we can do further drilling.
> 
> Please have a look, and let me know if you agree with this or if we can apply this can of logic in a more elegant way.

I like that. Much better to use _type with a string value.

Can you add a global template typedef with all the properties eg. template, _type, src, err, etc.

simon-leech (approved):
I'm happy with this - easily explainable and the tests are working well to help clean up!

AlexanderGeere (approved):
Nice One

PR #1501, opened 2024-09-25
Pass filter to minmax

simon-leech:
# Pass Filter to minmax

## Description

I noticed that the default `field_minmax` query contains `WHERE TRUE ${filter}`, however we never pass the `filter` as a parameter into the query. 
So the layer filter is never being used in this query. 
This is a bug as we need to ensure the filtering occurs only on the data used on the layer itself, not on all data on the table on the database.

I have resolved this by simply adding it to the list of parameters passed to the query. 

## Type of Change

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/tests/lib/ui/layers/filters.test.mjs
+++ b/tests/lib/ui/layers/filters.test.mjs
@@ -48,7 +48,7 @@ export async function filtersTest(mapview) {
             //Delete the min value from the filter
             delete filter.min;
 
-            //Delte the lte value of the current filter otherwise it will not be changed in the next test
+            //Delete the lte value of the current filter otherwise it will not be changed in the next test
             delete layer.filter.current[filter.field].lte;
         });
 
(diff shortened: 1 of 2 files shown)


tests/lib/ui/layers/filters.test.mjs
```diff
@@ -67,5 +67,22 @@ export async function filtersTest(mapview) {
             codi.assertEqual(minInput.value, '100', 'The min should return 100.');
             codi.assertEqual(maxInput.value, '1000', 'The max should return 1000');
         });
+
+        await codi.it('Numeric Filter: layer.current specified as `lte: 200` and `gte: 800`', async () => {
```
RobAndrewHurst, line 71: Nice one! Thanks :)

AlexanderGeere (approved):
I added a test for the change. Looks good to me!

dbauszus-glx (changes requested):
Review in progress

dbauszus-glx (approved):
lgtm

PR #1502, opened 2024-09-25
Add template typedef review usage test

dbauszus-glx:
_type is a property of the template typedef addded in this PR.

I merged updateTemplateUsage and checkTemplate to become the documented update_templateUsage method.

The undocumented updateTemplateUsage method was only called by the undocumented checkTemplate method which in turn updated the templateUsage param.

There are still a few things I want to check.

Having a custom_templates object and templateUsage object which are essentially the same but the latter has a counter feels odd.

The templates object can just be cloned with the count being added once.

It doesn't feel right to iterate through layers in locales to update_templateUsage on each iteration within an iteration.

The update usage should be done once after all locales and layers have been parsed.

I am worried that templates used in an entry, or dataview in a layer do not count to the usage if we only check the layer.templates and layer.template.

---
  - https://app.asana.com/0/0/1208407306090517

### DIFF
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -333,10 +333,10 @@ async function test(req, res) {
 
       if (layer.err) test.errArr.push(`${layerKey}: ${layer.err}`)
     }
-  }
 
-  // Test locale and all of its layers as nested object.
-  templateUse(workspace.locales, test);
+    // Test locale and all of its layers as nested object.
+    templateUse(locale, test);
+  }
 
   // From here on its 🐢 Templates all the way down.
   for (const key of Object.keys(workspace.templates)) {
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -359,9 +359,9 @@ async function test(req, res) {
   // Reduce the test.used_templates array to count the occurance of each template.
   test.results.usage = Object.fromEntries(test.used_templates
     // sort by usage
-    .sort((a,b) => {
-      if (a>b) return 1
-      if (a<b) return -1
+    .sort((a, b) => {
+      if (a > b) return 1
+      if (a < b) return -1
       return 0
     })
     .reduce((acc, e) => acc.set(e, (acc.get(e) || 0) + 1), new Map()));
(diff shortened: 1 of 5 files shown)


RobAndrewHurst:
```js
//We get the initial custom templates from the workspace.templates as we want to see usage of templates before we merge templates from layers.
    custom_templates = {
      ...Object.fromEntries(
        Object.entries(workspace.templates).filter(([key, value]) => value._type === 'workspace_template')
      )
    };
```

`We get the initial custom templates from the workspace.templates as we want to see usage of templates before we merge templates from layers.
`

I added in this comment where I get the workspace_templates from the workspace.templates object before we merge the layer templates as we want to see what's explicit templates are in the workspace.templates that either need to remain or be removed. 

If a template is being merged from a layer then chances are it's something that is needed for the workspace. But if its just old legacy config that is lying around this could help us 'couper le gras'.

If we have other ideas in how we could handle this, then lets discuss it :)

RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/issues/1494

I have updated the templates test spec with additional things we discussed.

simon-leech (changes requested):
It is valid to have no `workspace.templates` object, for example if all the templates are defined in the `locale.layers` and/or on `layers` themselves. 

This currently crashes in the `cache.js` module as there is no check on the presence of `workspace.templates`.

simon-leech:
``` json 
   "template": {
          "key": "metrics_dwelling_privaterented_perc_thematic_2021",
          "template": "1-999"
        },
```

With a configuration as above, the used_templates array will contain 
``` js
"used_templates": [ 
"1-999": 1
]
```
for example. It should use the key of the template instead

RobAndrewHurst (approved):
Nice one! I am happy with this!

dbauszus-glx:
The template object: key should not be removed from unused templates and added to usage statistics as these are added to the templates object and not immediately used. This prevents double counting.

Instead the key should be added to the overwritten_templates set if the template existed as a _type=workspace template in the first place.

simon-leech:
Just pushed a commit to sort the templates so related naming of templates are seen together - this is working as expected for me now - thank you!

simon-leech (approved):

RobAndrewHurst (approved):

Issue #1499, opened 2024-09-25
Comma Decimal in Numeric Formatter for Slider Element


dbauszus-glx:
The numeric formatter fails on slider element input elements for a user with a locale that uses comma decimals, eg. DE 🇩🇪

The comma seperator isn't displayed and the return is an integer instead of a float.

![image](https://github.com/user-attachments/assets/83d3ec1f-d441-476a-bae7-02b0d7b2761c)

dbauszus-glx:
I believe the issue to be unformatStringValue method of the numericFormatter module

![image](https://github.com/user-attachments/assets/f2b823f9-69e7-4620-8f7f-5a7b0864b3f8)

![image](https://github.com/user-attachments/assets/55db4ff4-3def-4b68-ab03-e729fa14688d)

RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/pull/1503

This is currently awaiting testing and approval.

PR #1503, opened 2024-09-26
numericFormatter locale issue

RobAndrewHurst:
## Description

The numeric formatter fails on slider element input elements for a user with a locale that uses comma decimals, eg. DE 🇩🇪

The comma seperator isn't displayed and the return is an integer instead of a float.
## GitHub Issue

https://github.com/GEOLYTIX/xyz/issues/1499

## Type of Change
- ✅ Testing

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ New Tests Added

---
  - https://app.asana.com/0/0/1208407306088410

### DIFF
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -108,7 +108,7 @@ The numericChecks method checks whether a provided numeric value is a number, la
 function numericChecks(value, params) {
 
   // Check whether value is a null.
-  if (value === null) return false;
+  if (params.onRangeInput && value === null) return false;
   // Check whether value is a number.
   if (isNaN(value)) return false;
 
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -107,6 +107,8 @@ The numericChecks method checks whether a provided numeric value is a number, la
 */
 function numericChecks(value, params) {
 
+  // Check whether value is a null.
+  if (value === null) return false;
   // Check whether value is a number.
   if (isNaN(value)) return false;
 
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -55,7 +55,7 @@ function oninput(e, params) {
   params.stringValue = e.target.value
 
   // Assign numeric newValue.
-  params.newValue = mapp.utils.unformatStringValue(params)
+  params.newValue = params.onRangeInput ? params.stringValue : mapp.utils.unformatStringValue(params);
 
   if (params.numericChecks(params.newValue, params)) {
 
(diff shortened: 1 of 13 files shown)


RobAndrewHurst:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

I think this is what we are looking for to solve this.

RobAndrewHurst:
I have added in a number formatter on both the formatNumericValue & unformatStringValue functions. The formatter is an object that is created from a given locale. This is how we determine how numbers are formatted.

```js
// Create a number formatter using the specified locale
const numberFormatter = new Intl.NumberFormat(params.formatterParams.locale);
```
 
 I then use the formatter to break down the number into parts (digits, decimal seprator etc.) 
```js
const parts = numberFormatter.formatToParts(1.1);
```

After that I find the decimal separator which is used in the locale, by looking for the 'decimal' type in the parts array.
```js
// Find the decimal separator used in this locale
const decimalSeperator = parts.find(part => part.type === 'decimal')?.value;
```

Then we replace the locale-specific decimal separator in the stringValue with a period. This normalises the number so that Javascript can perform a parseFloat.
```js
// Replace the locale-specific decimal separator with a standard period
const normalisedValue = stringValue.replace(decimalSeperator, '.');
```
Then we replace any characters that aren't digits, periods, or minus signs. This will clean up thousand separators. 
```js
// Remove any non-numeric characters, except for period and minus sign
const cleanedValue = normalisedValue.replace(/[^\d.-]/g, '');
```

RobAndrewHurst:
Changes to how sliders get formatted as changed. 

The value coming from a slider to the numericInput function `input` will not be formatted unlike a value from a numericInput. Therefore we need a check if we need to format that value with the utility. 

I have added in a flag `onRangInput` from the slider onRangeInput function to indicate to the input function that we do not need to format the value.

simon-leech (approved):
I pushed one commit, the `oninput` was not calling `onRangeInput` correctly. I have updated it to this line of code: 
```js 
oninput=${e => onRangeInput(e, params)}>`
```

I have tested this using English, German, French and Thai in XYZ and all working for me :)

AlexanderGeere (approved):
Lgtm

simon-leech (changes requested):
On an editable infoj entry, with a value of say 9999 - it is impossible to set this back to null. this should be possible. 

 This fails as CSS invalid gets applied so you can't save it.

When using a range slider on the infoj entry too, if the value is null - then the slider does not appear, which it should.

RobAndrewHurst:
> When using a range slider on the infoj entry too, if the value is null - then the slider does not appear, which it should.

image

This is because of this logic

cityremade (approved):
looks good to me. Icon scaling with default set with 2 decimals reformats ok. Also when editing numeric location fields incorrect separator is prevented 👍

simon-leech (approved):
All good with me now - thanks @RobAndrewHurst

AlexanderGeere (approved):

dbauszus-glx (approved):
lgtm

PR #1506, opened 2024-09-26
Match equals

simon-leech:
# Match Using = 

## Description

The `match` filter type was incorrectly using ILIKE, which is in fact case insensitive and thus not a simple match. ILIKE should be used for the `like` filter. = should be used for the `match` filter.

ILIKE is slower as postgres indexes don't always use it, so match should use a simple `=` that is recognised and indexes are used for this. 

I have also updated the sqlFilter tests to test each filter type and assert on the expected filter built. 
image

## Type of Change
- ✅ Documentation
- ✅ Testing

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/mod/utils/sqlFilter.js
+++ b/mod/utils/sqlFilter.js
@@ -28,7 +28,7 @@ const filterTypes = {
       .map((val) => `"${col}" ILIKE \$${addValues(`${val}%`)}`)
       .join(' OR ')})`,
 
-  match: (col, val) => `"${col}"::text ILIKE \$${addValues(val)}`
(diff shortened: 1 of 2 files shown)


dbauszus-glx (approved):
Working for me on text
image

Integer
image

and numeric fields
image

RobAndrewHurst (approved):
Very nice! And good updates of tests in the cli.

PR #1508, opened 2024-09-27
Infoj Order Correction

RobAndrewHurst:
## Description

The lookup for the infoj_order checks on key, then field, and finally query. This fails when a key exists but doesn't match.

We have created a set of the different fields that will then be compared with a has check on the entry.

## Type of Change

- ✅ Testing

## How have you tested this? 
Run the local tests.
## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -45,33 +45,47 @@ export default function infoj(location, infoj_order) {
   // Create object to hold view groups.
   groups = {}
 
-  // The infoj_order may be assigned to the layer.
-  infoj_order ??= location?.layer?.infoj_order
-
   // infoj argument is provided as an array of strings to filter the location infoj entries.
   const infoj = Array.isArray(infoj_order) ?
-    infoj_order.map(_entry => {
-
-      if (typeof _entry === 'string') {
-
-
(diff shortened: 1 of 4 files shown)


dbauszus-glx (approved):
Looking good.

I moved the map method into documented function to make this eassier readable.

image

I updated the test to be in order of infoj > order > execution > results > expected > test for readability especially when setting breakpoints.

image

Note that you can create an array from an elements list allowing to use map functions instead of pushing into results in a for loop.

```js
      const results = Array.from(infoj.children)
        .map(el => el.firstChild.innerText.trim())
```

AlexanderGeere (approved):
lgtm

RobAndrewHurst:
> Looking good.
> 
> I moved the map method into documented function to make this eassier readable.
> 
> image
> 
> I updated the test to be in order of infoj > order > execution > results > expected > test for readability especially when setting breakpoints.
> 
> image
> 
> Note that you can create an array from an elements list allowing to use map functions instead of pushing into results in a for loop.
> 
> ```js
>       const results = Array.from(infoj.children)
>         .map(el => el.firstChild.innerText.trim())
> ```

Nice one!

PR #1510, opened 2024-09-30
Update sqlFilter mod

RobAndrewHurst:
## Description

https://github.com/GEOLYTIX/xyz/security/code-scanning/208

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.11.0β**
+**v4.11.0**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 ![Codi Unit Tests](https://github.com/GEOLYTIX/xyz/actions/workflows/unit_tests.yml/badge.svg)
--- a/mod/query.js
+++ b/mod/query.js
@@ -318,6 +318,17 @@ async function executeQuery(req, res, template, query) {
     return res.status(400).send(`Failed to validate database connection method.`)
   }
 
+  // Return without executing the query if a param errs.
+  if (req.params.SQL.some(param => param instanceof Error)) {
+
+    paramsArray = req.params.SQL.map(param => param instanceof Error? param.message : param)
+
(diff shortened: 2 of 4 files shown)


AlexanderGeere (approved):
Lgtm

simon-leech (approved):

dbauszus-glx (changes requested):
Is there a test case for the filter?

dbauszus-glx (changes requested):
I updated the isValid description.

image

What about integer? Are integer always typeof bigint?

Would an object ever be a valid value? I can not think about a filter where an object would be valid?

How can I make the test fail?

Let's say I provide 'foo' string as a value for a numeric filter check. String is valid type but the filter crashes.

The type should be provided for the test. Ie. I add a value for an integer filter then type must be int or bigint. I add a value for a numeric filter then value must be numeric.

RobAndrewHurst:
> I updated the isValid description.
> 
> image
> 
> What about integer? Are integer always typeof bigint?
> 
> Would an object ever be a valid value? I can not think about a filter where an object would be valid?
> 
> How can I make the test fail?
> 
> Let's say I provide 'foo' string as a value for a numeric filter check. String is valid type but the filter crashes.

Good question about the types. 

We appear to get an array of params as a param when using ni or in filters. So the type of object would be valid.

Doesn't number cover integers and floats?

RobAndrewHurst:
image

dbauszus-glx:
I would expect something like this?

https://github.com/GEOLYTIX/xyz/pull/1510/commits/8edd8400cafcaca5e18874195294cab6e386901f

If addValue numeric then test for number, if string test for string, if array then test for array.

dbauszus-glx (approved):
This is working for me.

I added a check on the error to not execute the query.

https://github.com/GEOLYTIX/xyz/pull/1510/commits/07de3c3c83d6215e5343a50fef4f9675304a2627

Error is logged in process console.
image

And returned to client.
image

PR #1529, opened 2024-10-03
Update Layer Test

simon-leech:
# Layer Test Fix and TLC

## Description

The Layer Test needed some TLC, comments have been added to explain the various stages of the test. 

The requirement for a geometry to be passed to the `newLocation` is valid. We provide either the `location.Pin` geometry, or just use the center of the mapview if there's no pin geometry.

The check on only deleting if `deleteLocation` is provided has been removed, we should always clear out locations generated in the test script.

## Type of Change

- ✅ Testing

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/public/tests/integrity.test.mjs
+++ b/public/tests/integrity.test.mjs
@@ -1,5 +1,7 @@
 import { base } from './_base.test.mjs';
 import { layerTest } from './layer.test.mjs';
+import { workspaceTest } from './workspace.test.mjs';
 
 const mapview = await base();
 await layerTest(mapview);
+await workspaceTest(mapview); 
--- a/public/tests/layer.test.mjs
+++ b/public/tests/layer.test.mjs
@@ -28,152 +28,149 @@ export async function layerTest(mapview) {
 
   await codi.describe(`${mapview.host} : Layer Test`, async () => {
 
-    for (const key in mapview.layers) {
+    for (const key of Object.getOwnPropertyNames(mapview.layers)) {
 
-      if (mapview.layers.hasOwnProperty(key)) {
+      await codi.it(`Layer test : ${key}`, async () => {
 
-        await codi.it(`Layer test : ${key}`, async () => {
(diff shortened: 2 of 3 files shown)


dbauszus-glx:
The integrity=true flag must be set in the URL params for the testview in order to test this.

dbauszus-glx (changes requested):
I struggle reading through the execution of the layer.test.mjs

Is there a reason why the template paths and layer test are two seperate checks? Shouldn't the template paths be checked within the layer test?

I don't understand the delayFunction which is only used in the themes test.

Why is there a 1000ms required? That should be commented on.

```js
            for (const theme in layer.style.themes) {
              console.log(`Testing theme ${theme}`);
              layer.style.theme = layer.style.themes[theme];
              layer.reload();
              await delayFunction(1000);
            }
```

I don't understand the use of the default_zoom const. Can this be commented.

```js
const default_zoom = mapview.view?.z || 0
```

There is a lot of heavy nesting. At some points there are 10+ deep nestings where there is no way of knowing how deep I am when checking.

Some nesting can be resolved by using a `for of Object.getOwnPropertyNames` loop.

I don't understand this condition. This needs to be commented on. What's with maplibre and tiles format?

```js
if (!['maplibre', 'tiles'].includes(layer.format) && layer.infoj) {
```

Is it possible to put this whole location ? test into a function which is easier to read?

image

dbauszus-glx (approved):
I replaced the delay method with a one liner.

From my point this working and I understand what the tests are doing.

RobAndrewHurst:
> I replaced the delay method with a one liner.
> 
> From my point this working and I understand what the tests are doing.

Nice thanks! I can see why it can be a one liner. 

Each to their own :)

PR #1531, opened 2024-10-03
Force workspace cache

dbauszus-glx:
The [workspace] cache will be set to null, forcing re-caching with the force param flag.

This allows to clear the cached workspace everytime the workspace test method is called.


RobAndrewHurst (approved):
This all works for me.

I added in a test in the local test that will check if we get the  same amount of entries in the workspace after two runs of the workspace test is executed.

simon-leech (approved):
All good with me :)

PR #1534, opened 2024-10-03
compose util documentation

dbauszus-glx:
The compose method wants to be in its own module file for this.

This wants a code example as well.

```js
async function logLayer(layer) {
  await layer
  console.log(layer)
}

mapp.layer.decorate = mapp.utils.compose(
  mapp.layer.decorate.bind(),
  logLayer)
```

### DIFF
--- a/lib/utils/_utils.mjs
+++ b/lib/utils/_utils.mjs
@@ -18,16 +18,7 @@ const areSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(
 
 import csvDownload from './csvDownload.mjs'
 
-/**
-## mapp.utils.compose()
-@function compose
-@memberof module:/utils
-*/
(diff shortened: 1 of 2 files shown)


RobAndrewHurst (approved):
Nice happy with this!

I have also added in a few tests to check the utility.

```js
/**
 * ### Should compose functions from left to right
 * @function it
 */
codi.it('Should compose functions from left to right', () => {
    //Different functions to use
    const addOne = x => x + 1;
    const double = x => x * 2;
    const square = x => x * x;

    //Creating the compose chain
    const composed = mapp.utils.compose(addOne, double, square);

    codi.assertEqual(composed(3), 64, 'We expect this order of opperation when composing functions left to right : ((3 + 1) * 2)^2 = 64');
});
```

PR #1535, opened 2024-10-04
Destroy redirect cookie

dbauszus-glx:
This PR attempts to destroy the redirect cookie.

The redirect cookie is created to redirect (set the response header location) in the login module.

The cookie method which is called to renew a cookie must not attempt to create a new cookie and return the login view which will cause the creation of a redirect cookie on public instances. Instead the cookie should return false without parameters and only attempt to create a new cookie with the create url parameter flag.

Some param, property, and typedef confusion has been resolved in the function docs.

This PR adds the redirect value to the request params to be substituted in the english default login view. This is a test to get eventually rid of the redirect cookie.


dbauszus-glx:
Removed this replace as it is unclear what this is doing and why. `.replace(/([?&]{1})msg={1}[^&]+(&|$)/,'')`

RobAndrewHurst (approved):

AlexanderGeere (approved):

PR #1539, opened 2024-10-07
Remove filter method in filter dropdown

dbauszus-glx:
The card element and filter panel module and method have now been documented.

Checks for types have been added.

The removeFilter method has been added to the layer filters module. The removeFilter calls the apply filter method rather than duplicating code from the apply filter method.

Elements are assigned to the layer filter object rather than const within the panel method.

The dropdown callback method has been reduced in complexity.

---
  - https://app.asana.com/0/0/1208488860427079


RobAndrewHurst (approved):
I am happy with this!

I also updated the filters test to use the `removeFilter` method and all works as expected!

RobAndrewHurst:
Fixed 2 issues on this PR.

The drop down elements did not have a string value if there was no title on the entry. It now uses title, then label then field.

Then I moved up a check for filter that are of type string. This will ensure the filter type is correctly defined before checking against the filter methods on layer.ui.filters.

simon-leech (approved):

PR #1542, opened 2024-10-08
remove cloudinary provider module

dbauszus-glx:
The cloudinary provider module which requires the image/document payload to be passed through the xyz host has been removed.

The signer module has now been reviewed and documented.

The cloduinary signer module has now been reviewed and documented.

The module will not fail silently but return an error message.

The cloudinary entry is now using the cloudinary signer module.

The cloudinary entry is now reviewed and documented.

The cloudinary entry requires a cloudinary account. However, no third party module is required to sign requests, hence the cloudinary signer is included in the XYZ core and the cloudinary entry method is included in the mapp location entries.


cityremade (approved):
Works as expected.
I've replaced alerts and confirms with ui elements plus more human wording on removal confirm;

RobAndrewHurst (changes requested):
image

when removing a document that uses cloudinary the alert says I am going to remove an image which doesn't align with the action.

We should change this to something more fitting for documents

RobAndrewHurst (changes requested):
image

All of these dictionary entries except the remove_item_confirm are not used anywhere in the core, and should be removed.

RobAndrewHurst:
> Works as expected. I've replaced alerts and confirms with ui elements plus more human wording on removal confirm;

This can't work with both images and documents.

lib/ui/locations/entries/cloudinary.mjs
```diff
@@ -207,85 +251,123 @@ function imageLoad(e, entry) {
...
+    body: data
   })
 
   if (!response || response.error) {
-    const errorDetail = response?.error?.message ? ` Error: ${response.error.message}` : '';
-    const errorMessage = `${mapp.dictionary.document_upload_failed}${errorDetail}`;
-    alert(errorMessage);
+    const errorDetail = response?.error?.message? `Error: ${response.error.message}` : '';
+
+    alert(`Cloudinary document upload failed! ${errorDetail}`);
```
RobAndrewHurst, line 331: We will also want to make sure that all alerts are now using the core defined alerts in the module!
```suggestion
    alert(`Cloudinary document upload failed! ${errorDetail}`);
```mapp.ui.elements.alert();

RobAndrewHurst (approved):
I have have replaced the remove_item_confrim entry on the dictionary with two items. One for images and the other for documents.

They are then passed to the trash method as params.

AlexanderGeere (changes requested):
Writing tests for the module

AlexanderGeere (approved):

RobAndrewHurst (approved):
Happy with this!

PR #1550, opened 2024-10-09
Assign and check filter.li element in dropdown callback

dbauszus-glx:
The filter panel dropdown assigns the (e.target) li element to the filter object and sets the selected class.

The selected class is removed from the li element in the filter remove method.

The filter remove method is  called when the li is already selected in the dropdown callback.


RobAndrewHurst (approved):
Nice!

This works well!

simon-leech (approved):
All working for me - nice one! :)

Issue #1548, opened 2024-10-09
Slider Styling Revamp 🚧 


RobAndrewHurst:
# Slider Styling Revamp 🚧 

We are looking to restyle the core slider element that will be used for both normal and ab sliders.

### Vanilla Slider
1. On vanilla slider elements put the label and input on the same line.
2. Input element needs to dynamically resize to value.  

### AB Slider
1. The input elements for the ab slider need to be put on the same line.
3. The input elements needs to dynamically resize to accommodate the size of the value in the inputs.
4. Remove greater than or less than labels (These are self explanatory).
5. Add an In title on ab sliders that will indicate that we are filtering between these two values.

### Slider
1. The Slider element itself will remain the same.

Bellow is an example of the ab slider that we are wanting to replicate. The vanilla slider will take influence from this as well.

![image](https://github.com/user-attachments/assets/3d55d8ad-01e7-4a2e-88ef-a41a62b4a1f7)

simon-leech:
![image](https://github.com/user-attachments/assets/44685355-0a5a-4b65-8c72-2c94067cae64)

I was just playing and just throwing another potential option out there.
What about putting the min / max input boxes at the start/end of the slider. They be fixed in place but to me at least it is implicit they are min/max and you wouldn't need the 'In' word too?

RobAndrewHurst:
[...]

PR #1552, opened 2024-10-10
Input Elements: CSS changes

AlexanderGeere:
## Description

Changes the css for input elements.
1. The input boxes above a slider now appear on either end of the slider elements thumbs.
2. Added a new parameter to numericInput which lets the input element grow and shrink dynamically according the length of the value.
3. Input elements are now borderless except the bottom and have a slight grey background.
4. The Range editor will place the slider below the input box.
5. Added padding to the dialog element and to the content element of the dialog to avoid scrolling.
## GitHub Issue

[#1548](https://github.com/GEOLYTIX/xyz/issues/1548)

## Type of Change

## How have you tested this? 
Tested using the multi_filter PR and any instance that has filters and editing.

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -26,7 +26,7 @@ export default function numericInput(params) {
 
   const value = mapp.utils.formatNumericValue(params)
   const width = params.dynamicWidth ? (value.length + 1.3) + 'ch' : '100%'
-  const style = `text-align: right;width:${width}`
+  const style = `text-align: center; width: ${width}`
 
   const numericInput = mapp.utils.html.node`<input
     data-id=${params.data_id}
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -24,12 +24,16 @@ export default function numericInput(params) {
   params.data_id ??= 'numeric-input'
   params.numericChecks ??= numericChecks
 
+  const value = mapp.utils.formatNumericValue(params)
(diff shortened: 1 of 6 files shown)


dbauszus-glx:
Do we need the light-off-black color? I can't really tell the difference between light-secondary and light-off-black.

image

dbauszus-glx (approved):
https://github.com/GEOLYTIX/xyz/pull/1552/commits/f5d691b40d859b1035f1a7e68cd2f12bdeb17333

I set the text align to centre as the width is not perfect depending on the character widths.

I also removed --color-light-off-black to not use a colour once when there is another colour --color-light-secondary which looks almost the same.

The dialog content should not have additional padding by default causing a mis-alignment with the header. Padding to the content can be added with the content if a misalignment is the desired effect.

I added the position: fixed to the dropdown css. This seems to work like the absolute position hence there is no need to set this to absolute first and then to fixed in a dialog.

I removed the padding on label row as this is not required with the default style.

image

RobAndrewHurst (approved):
Very Happy with this!

Nice @AlexanderGeere

simon-leech (changes requested):
All working well for me! 

Just one question / comment - should we align the styling of textarea input with text and number? It stands out at the moment as different but the new styling is much cleaner so I would suggest we keep them all inline :)

image

simon-leech (approved):

PR #1558, opened 2024-10-11
callback on invalid numericInput

dbauszus-glx:
The location.view valChange event method checks for `.some` invalid entry in the location.layer infoj array.

However the valChange method is only triggered in the entry.callback method.

It is required for the numericInput module method to return with the callback method if the input is invalid.

### DIFF
--- a/lib/ui/elements/numericInput.mjs
+++ b/lib/ui/elements/numericInput.mjs
@@ -81,7 +81,7 @@ function oninput(e, params) {
   }
 
   // The invalid input should not be formatted.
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
I played about with an infoj entry containing a range of input types. 
I can confirm this is working as expected, when I input an invalid value into a numeric field - regardless of if other entries are edited or not, it correctly hides the save button,

I have tested both using toggleLocationViewEdits and without and both are working for me

AlexanderGeere (approved):
Working as expected

PR #1564, opened 2024-10-14
query params from template, layer, locale, or workspace

dbauszus-glx:
For security reasons the dbs should not be provided as a request parameter.

Any references to the dbs request parameter have been removed from the mapp library in this PR.

The layer.dbs will be set to the locale.dbs which will be set to the workspace.dbs.

The layer.dbs or workspace.dbs will be used as fallback if the template object has no dbs property.

---
  - https://app.asana.com/0/0/1208538873788583


dbauszus-glx:
The layer parameter should always be provided if available.

simon-leech:
This is working for me

Note when testing: 
Queries that are ran in custom_views that don't use the workspace.dbs will need the layer and locale providing as params to use that dbs

simon-leech (approved):

simon-leech:
Also please don't merge until @RobAndrewHurst  has had a look after his hols!

RobAndrewHurst (approved):
Updated the tests, and I am all happy with these changes.

Issue #1566, opened 2024-10-15
Filter Panel: Reset all filters


AlexanderGeere:
## Description
The multi filter needs a reset all button. We should add this to core as part of the filter

## Type of Issue 
- ✅ New Feature
- ✅ Documentation

PR #1567, opened 2024-10-15
Filter panel: reset all function

AlexanderGeere:
## Description
Adds the ability to reset the filters back to their starting positions.

Reset all will leave the filter in the dom, but just remove any user provided input and clear the parameters from the filter. 

## GitHub Issue

[#1566](https://github.com/GEOLYTIX/xyz/issues/1566)

## Type of Change
- ✅ Documentation

## How have you tested this? 
Tested on the demo multi_filter.  Can be tested on `bugs_testing/filter/all_filters.json`. To test the multi_filter please use the PR.
(The multi filter is present on `bugs_testing/filter/all_filters.json`)

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

---
  - https://app.asana.com/0/0/1208550579293270

### DIFF
--- a/lib/ui/elements/card.mjs
+++ b/lib/ui/elements/card.mjs
@@ -21,7 +21,7 @@ export default function card(params) {
   
   params.data_id ??= 'card'
 
-  params.card = mapp.utils.html.node`
+  const card = mapp.utils.html.node`
   <div 
     data-id=${params.data_id}
     class="drawer">
@@ -38,5 +38,5 @@ export default function card(params) {
     </header>
     ${params.content}`
 
-  return params.card;
+  return card;
 }
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -92,6 +92,23 @@ function reloadCallback(layer) {
   return layer;
 }
 
+/**
+@function removeFilter
+
+@description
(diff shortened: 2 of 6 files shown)


simon-leech (changes requested):
I noticed we use the same styling `margin-bottom: 5px;text-decoration: underline` for both buttons. 
Can we move that to a class to be accessible for other elements too? 

Would you be able to write a test for this too so we can confirm its working as expected?

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/1567/commits/ca7a70418ebb8ca3a809d734bd76b5ec10ef6d6c

Instead of altering the filter on reset it is much easier to delete the filter and replace the card.

The delete filter has been moved into a method to avoid code duplication with removeFilter method.

The docs have been updated as only the deleteFilter method needs to know the filter properties and remove / reset just need to know the html elements.

The card element had to be updated. The card cannot be assigned to the properties. This would make replacing the element impossible.

dbauszus-glx (approved):
Looks good to me. But I also did a few changes so this must be reviewed by someone else in detail.

RobAndrewHurst (approved):
Happy with this!

PR #1572, opened 2024-10-16
Null Geometry - Does not refresh geometry entry on the map

simon-leech:
# Pull Request Template

## Description

If I have a `geometry` entry that is a `dependent` of another field. 
If the `geometry` entry is already turned on, and when refreshed as a `dependent` it is then null, it is not removed from the mapview. 

This PR addresses that by checking if entry.value is null, and removing the layer if it exists.

## Type of Change

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -93,6 +93,14 @@ export default function geometry(entry) {
 
   entry.zIndex ??= entry.location?.layer?.zIndex + 1 || 99
 
+  // If the value is null, the entry must be removed from the mapview (if it exists).
+  // This is possible if the entry is already drawn, and then is a dependent of another entry.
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
https://github.com/GEOLYTIX/xyz/pull/1572/commits/3232e14626debf7b3cedeb648006dd4452246c2b

The layer.display does not have to be altered before the return which prevents the entry being added to the location view.

Setting the display to false will prevent a an api geometry from being drawn on a new location with display true.

The entry layer can be removed before the return in the `(!entry.value && !entry.draw && !entry.api)` condition check.

RobAndrewHurst (approved):
Updated tests!

Happy with the changes!

PR #1583, opened 2024-10-17
Theme filter and layer.featureFilter to filter features in style.

dbauszus-glx:
It should be possible to toggle categories in a legend without applying a filter.

This was already possible through the featureFilter but untested, awkward, and undocumented.

The featureFilter method has now been removed.

The categorized theme method will set the feature style to null if the cat.style is null.

The featureStyle method will shortcircuit after the theme method if the feature.style is null.

The `theme.filter` indicates that a theme [style only] filter should be applied.

The onclick method for the cat element label in the categorized legend cat iteration has been unnested into its own method catToggle.

The catToggle method sets the cat style to null if toggled and restores the style from the cat._style property if toggled off.

The filterAdd and filterRemove methods will add or remove the cat value to the cat field filter. These methods will shortcircuit if the theme.filter is not set.

The theme.filter will be set from style.filter for all themes in a layer.style.


simon-leech (changes requested):
Using a layer with multiple thematics: 
1. I set a cat to disabled, it is styled as null and not added to the filter current✅
2. I swap the thematic, all features on the other thematic are styled ✅ 
3. I swap back to the original thematic, the original cat that was disabled now has a style of null but this means I can't turn it back on, as it has no style. ⚠️
4. Also, the original cat that was disabled is not disabled (greyed out) ⚠️

simon-leech:
Also, is there a better name than `filter` we can use? While testing I found it quite confusing applying a `filter: true` to prevent it being added to the `filter.current` ?

dbauszus-glx:
> Also, is there a better name than `filter` we can use? While testing I found it quite confusing applying a `filter: true` to prevent it being added to the `filter.current` ?

There is probably a better name but the filter flag has been in use for at least a year for this use case but was never documented. Happy to use a different flag and then assign the filter flag in the styleParser and warn if used.

dbauszus-glx:
The onclick label event method will now set unset the disabled flag on the cat object. The _style is used for the legend icon creation if available. This should ensure that the theme.filter remains when switching to a different theme and back.

simon-leech (approved):
All working for me - happy with this change! 👍 

Would like to rename the filter but the functionality is working

simon-leech:
> > Also, is there a better name than `filter` we can use? While testing I found it quite confusing applying a `filter: true` to prevent it being added to the `filter.current` ?
> 
> There is probably a better name but the filter flag has been in use for at least a year for this use case but was never documented. Happy to use a different flag and then assign the filter flag in the styleParser and warn if used.

What about calling it `styleFilterOnly` which is set to true when you wish to only apply the filter to the layer style (ie not add to the layer filter)?

dbauszus-glx:
@simon-leech https://github.com/GEOLYTIX/xyz/pull/1583/commits/4733e0c0f706e0093e8f81ba774dad0fac6d2be3

RobAndrewHurst (approved):

Issue #1309, opened 2024-06-06
Cognitive overload of XYZ api.js module


dbauszus-glx:
The XYZ api method has not been reviewed in ages and has a massive cognitive overload.

https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=1264&id=GEOLYTIX_xyz&open=AY_uWqKATpgutnEmB8VD

The warning in regards to the await for the async auth method seems to be a sonarcloud ai halucination.

dbauszus-glx:
The user object request param must never be assigned by a request. https://github.com/GEOLYTIX/xyz/security/code-scanning/211

Error messages returned in response to a failed request should be escaped.

PR #1588, opened 2024-10-21
layer query and templates

dbauszus-glx:
The geojson query template has been flagged as a layer query which requires a layer request parameter.

The getLayer method will attempt to retrieve a layer template from the templates if the layer param is not found in the locale.layers.


dbauszus-glx:
I dismissed the security alert as not relevant. This should be addressed but not in this PR.

The issue is related to the params.user object which is assigned in the API method. User roles may be merged into a layer/template object and an error.message maybe returned from a failed query.

The security of the user object must be controlled in the API module which is yet to be tested and documented.

Message objects returned from failed requests should be escaped which may require a library. I will revise this in the relevant PR. https://github.com/GEOLYTIX/xyz/issues/1309

RobAndrewHurst (approved):
Very happy with this!

Added in a test to check for a template that does not have a layer param in the params!

AlexanderGeere (approved):
Ran integrity tests on an instance, checked reports. Looks good to me

cityremade:
query endpoint is widely used to request data not related to a specific layer, such as non-spatial data based on location id.
How is this going to be supported?

cityremade (approved):
No objections;

PR #1601, opened 2024-10-23
remove mapp.layer changeend method

dbauszus-glx:
Several inline/inliteral methods have been unnested.

The mapview changeEnd event is no longer nested in the mapview decorator but bound to the mapview object as mapview.changeEnd()

This function is bound to the mapview object in the decorator and triggered by the mapview.Map moveend event.

The mapp.layer.changeEnd module has been removed. The changeEnd function has been added to the layer/decorate module. The method will be added to the mapview.Map changeEnd eventlistener for layer with tables or viewport param.

The condition for the layer.tables and viewport param has been removed from the mvt and vector format methods.

The mvt changeEndLoad method has been renamed to wktPropertiesLoad with the documentation linking to the featureFormats-wkt_properties and featureStyle-featureProperties methods to explain the functionality better.

A changeEnd event will be dispatched when a new location is created in the mapview.

The plugins/zoomBtn method documentation has been cleaned up. There was some confusion on params/properties.

The /ui/layers/view module and layersView method has been documented and cleaned up. The zoomToRange and changeEnd methods have been unnested and documented.


cityremade:
does it affect deleting features from editable layers? when I remove a feature from a layer the deletion is complete but the feature stays on the map until reload.

dbauszus-glx:
> does it affect deleting features from editable layers? when I remove a feature from a layer the deletion is complete but the feature stays on the map until reload.

Yes, it should be possible to delete a geometry entry or delete a location with associated geometry.

[Screencast from 2024-10-25 14-45-47.webm](https://github.com/user-attachments/assets/983d9c9e-57a2-463b-b6b6-88410720921b)

RobAndrewHurst (approved):
All happy with this!

simon-leech (changes requested):
There is a bug with editing a location on an `mvt` layer. 

Steps to replicate: 
1. MVT Layer with a categorized thematic on field_a 
2. Select a location on this layer, and update the value of field_a to a different option in the thematic.
3. The locations styling is not updated correctly on the mapview, the theming is not updated.
4. If you zoom out or in, the thematic will update, but when you go back to the original zoom, it will be as it was initially

dbauszus-glx:
@simon-leech I create a separate issue for this. https://github.com/GEOLYTIX/xyz/issues/1624

simon-leech (changes requested):
Re-review in progress

simon-leech (approved):
All looks good to me - tested on a range of different layer types, updating the fields, adding new locations, deleting locations and seemed to be all good :)

Issue #1610, opened 2024-10-28
Feature Properties reference correction


RobAndrewHurst:
## Description

The feature properties referenced in the highlight and graduated modules are incorrect.
This is because of a change we did to spread the properties into the featureFormats module. 

- ✅ Bug

PR #1612, opened 2024-10-28
Fix feature properties

RobAndrewHurst:
## Description

The properties are now spread into the OL element and we need to correct the feature properties references. 

## GitHub Issue

#1610 

## Type of Change

### DIFF
--- a/lib/layer/featureFormats.mjs
+++ b/lib/layer/featureFormats.mjs
@@ -81,7 +81,7 @@ export function wkt(layer, features) {
     const properties = {}
 
     //To ensure that we know that id and geometry are first, and then additional params get added to the set.
-    let wkt_params_fields = Array.from(new Set(['id', 'geometry', ...layer.params.fields]));
+    const wkt_params_fields = Array.from(new Set(['id', 'wkt_geometry', ...layer.params.fields]));
 
     // Populate featureFields values array with feature property values.
     wkt_params_fields?.forEach((field, i) => {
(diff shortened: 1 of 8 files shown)


dbauszus-glx:
I have done a review of all OpenLayers feature getProperties() use cases.

https://github.com/GEOLYTIX/xyz/pull/1612/commits/afe5a35733023bb538608e538ec33607aa0b4194

These should always be stored as const featureProperties with the OL feature as F.

Please double check the cluster label assignment.

This is now working for me but was bugged.

image

dbauszus-glx (approved):
Working for me now.

dbauszus-glx (changes requested):
The latest change of removing the +2 in the WKT featureFormats shift breaks the format.

RobAndrewHurst:
> @RobAndrewHurst I don't understand the shift change.
> 
> I have two fields retailer, and store.
> 
> image
> 
> This means they are indexed 2 and 3. With 0 being the id and 1 being the geometry.
> 
> image
> 
> Removing the +2 in shift breaks WKT featureFormats method for me.
> 
> image

Yeah we noticed that now! We have come up with a fix!

Will document the fix now!

dbauszus-glx (approved):
Looking much cleaner now.

AlexanderGeere (approved):
Nice one!

simon-leech (approved):
Happy with this - tested on a range of layer formats

PR #1617, opened 2024-10-29
deprecate mbtiles format

dbauszus-glx:
The mbtiles format dependent on a legacy version of the maplibre library has now been removed.


simon-leech (changes requested):
Can StyleParser check for this layer format and warn (can it parse to the correct format)? 
Currently if you have an mbtiles layer then it will just crash the whole thing?

dbauszus-glx:
> Can StyleParser check for this layer format and warn (can it parse to the correct format)? If you have an mbtiles layer on this PR, it just crashes without any helpful debugging or anything.

Not possible since the styleParser checks on vector styles only and mbtiles do not have a style object.

dbauszus-glx:
@simon-leech You can add a deprecation warning in lib/layer/format/_format.mjs

```js
/**
@function mbtiles
@deprecated

@description
The mbtiles layer format method has been deprecated.
*/

function mbtiles() {

  console.warn(`The mbtiles layer format method has been deprecated.`)
}
```

PR #1625, opened 2024-10-29
Drawing  ui - circle settings in a dialog;

cityremade:
## Description
Layer Drawing UI

Circle drawing current settings are displayed in a dialog when the button in the drawing panel is hit. 
They can be changed or reused before drawing interaction starts - or can be cancelled.
Putting these settings within an expandable panel is not clear that anything important is nested there 
so showing default/recent settings to the user is a possible solution to both accessibility and clear layout in the panel.

Screenshot 2024-10-29 at 16 56 45

## GitHub Issue

## Type of Change


RobAndrewHurst:
An issue needs to be created or assigned to this pr

PR #1628, opened 2024-10-30
Explain the WORKSPACE AGE in more detail

dbauszus-glx:
The getTemplate and cacheWorkspace methods are now linked with an example and link to the query module why it is not recommended to set the WORKSPACE_AGE to 0.


cityremade (approved):

AlexanderGeere (approved):

RobAndrewHurst (changes requested):
Do you think we should include a recommended workspace age in ms that a developer could use to with seeing local changes reflect quicker?

dbauszus-glx:
@RobAndrewHurst I wouldn't know a recommended value. The workspace age is only required if a workspace is likely to change. By default the workspace doesn't change hence there is no default age. I don't want to set this to an hour, or a day and then expkain that this is just an example default not required which can be overwritten.

RobAndrewHurst (approved):

PR #1633, opened 2024-10-30
Remove data-id assignment for layer-view element

dbauszus-glx:
A data-id is not a class!

A data-id is a pseudo id as the nomination suggests.

The data-id for a layer view/drawer element should be the layer.key.

The data-id for a [layer] group view/drawer should be the group id.

There should not be a queryselector for `data-id=layer-drawer` which is better solved with a combo class eg. `.layer-view.drawer`


AlexanderGeere (approved):

cityremade (approved):
this looks good.
Click on layer toggle now has better access to layer key 👍

RobAndrewHurst (approved):

PR #1642, opened 2024-10-31
CSV Upload Utility

simon-leech:
# CSV Uploading

## Description

This PR provides a new utility `lib.utils.csvUpload`. This module takes an input and parameters and stores the data to the database in chunks. 
The module returns a Promise, and an `outcome` object. 
The `outcome` object will contain a `success` flag, of true or false indicating the success of the import, and a message, which can be displayed to provide information on the reason for failure. 

## GitHub Issue
#1676 

## Type of Change

- ✅ Documentation

## How have you tested this? 
Tested on `bugs_testing/plugins/csv_upload/workspace.json` 

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


AlexanderGeere (changes requested):
The upload button on the entry doesn't seem to get the id, when I attempt the upload the id is undefined, and nothing goes into the table. Side note it does say the import was successful.

dbauszus-glx (changes requested):
This is way too complicated for general plugin / utility.

There is a situation where a very specific update query is executed from the resolution of multiple promises with insert quaeries which have to modules and must receive a body with an array property arr.

I would recommend to cleanup, test, and document POST queries with an array body first.

This is a valid select statement that will return to you the array sent as POST body.

```json
    "select_body": {
      "template": "SELECT %{body}",
      "value_only": true
    },
```

image

image

Any rerences to the updateQuery should be removed from the csvUpload utility. These should be handled in the plugin which I would add in a seperate the PR.

The csvUpload utility should be in line with the csvDownload utility which merges multiple array responses into a CSV file.

The csvDownload method should be fully documented before another even more complex utility is added.

The specification for the csvUpload utility method should be to split a csv into seperate array bodies provided as POST body to a SQL query and return an array of responses from the individual queries once all queries are resolved. It must be possible to provide queryparams for the individual queries as well as a parameter for the payload to use in the array split. It should be possible to define whether all queries should be run in parallel or whether the queries should be sent consecutively.

This can be tested with very small csv file and providing a small enough payload param to force this csv to be split into at least two parts to be sent to the simple select %{body} statement which returns the body.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,222 @@
...
+  );
+
+  const url = `${mapp.host}/api/query?${paramString}`;
+
+  // Push upload query into promises array.
+  promises.push(
+    mapp.utils.xhr({
+      method: 'POST',
+      url,
+      body: JSON.stringify({ arr: chunk }),
```
RobAndrewHurst, line 219: Testing!

RobAndrewHurst:
Test

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,229 @@
...
+/**
+@function csvUpload 
+@async
+
+@description
+This function uploads a CSV file to a database store.
+
+@param {File} file The CSV file to upload.
+@param {Object} params The parameters object.
+@property {Array} params.schema The schema array of column types.
```
RobAndrewHurst, line 18: Do we need to be so specific as to show an array? in docs?
What is the naming standard of these types. Is it 'function' or 'Function'. Is it 'string' or is it 'String'.

RobAndrewHurst, line 18: Do we need to be so specific as to show an array? in docs?
What is the naming standard of these types. Is it 'function' or 'Function'. Is it 'string' or is it 'String'.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,229 @@
...
+@property {String} [params.headerCheckErrorMessage] The error message to display if the number of columns in the file does not match the schema length.
+
+@returns {<Promise>} The outcome object and any messages returned by the database.
+*/
+
+export default async function csvUpload(file, params = {}) {
+  // Create an outcome object that will be returned at the end of the function.
+  // This means other function that make use of this method can check if the import was successful.
+  // They can then handle the success or failure of the import and show the user a relevant message.
+  const outcome = { success: true };
```
RobAndrewHurst, line 33: Should we not maybe want this to default to false. As that makes more sense from an error handling perspective.
It doesn't make sense to me that we start with a condition that we have successfully uploaded a csv.

lib/utils/csvUpload.mjs
dbauszus-glx: It doesn't really matter whether the type in capitals or not since we have been using either throughout.

lib/utils/csvUpload.mjs
dbauszus-glx: I would like the logic with the outcome object to be changed but looking at this at the moment. Need to get a better understanding with some test cases which do not invlolve a plugin.

RobAndrewHurst:
Just putting this back into draft until @dbauszus-glx is ready with his changes.

dbauszus-glx:
> Just putting this back into draft until @dbauszus-glx is ready with his changes.

Thanks. I should have done this yesterday already.

dbauszus-glx:
I created a csv_upload utility for the json dataview to test the upload without the need for a plugin.

image

This dataview will simplly pump the rows into the select body query which returns the body from the post request to the sender.

```json
"dataviews": {
  "csv_json": {
    "display": true,
    "target": "tabview",
    "label": "Json",
    "dataview": "Json",
    "toolbar": {
      "jsonfile": true,
      "csvupload": {
        "header": true,
        "query": "select_body",
        "template": {
          "key": "select_body",
          "template": "SELECT %{body}",
          "value_only": true
        }
      }
    }
  }
}
```

I made a few changes to the csvUpload module.

The chunkSize is assigned to the params if nullish/undefined.

A check must be performed on the params.query prior to returning the Promise.

The headerCheck params property has been retired. If the header flag param is set then the file has a header.

The splitCSVIntoRow method has been renamed to splitRowIntoFields to indicate the purpose of the method.

The schemaMethods object is defined in the module, not inside the splitRowIntoFields method which doesn't require the params object.

'Text' is assumed as schema method for all fields if the schema params is undefined.

The processChunks method has been renamed to postRequestData as this method will post the data received and return the XHR utility method [promise].

The data chunks are pushed into the chunks array. The promises array is not required.

If async the chunks array will be mapped to return the XHR utility promises from the postRequestData method to the Promise.all() returning the indicidual responses.

If not async the chunks array will be iterated through in a for...of loop and the xhr from the postRequestData method will be awaited prior to the next chunk being loaded.

dbauszus-glx (changes requested):
I feel there is a much simpler approach to splitting csv rows with a regex command. Iterating through each character feels inefficient.

dbauszus-glx (approved):
I am good with this but have written a lot of the logic. This must be tested by other people as well.

dbauszus-glx:
I think the CSV parser itself should be managed in a seperate ticket if needed. This is a lot more complex than I thought.

This is a good article why you shouldn't roll your own CSV parser. https://secretgeek.net/csv_trouble

I think in the future we should import a module like PapaParse for the job. https://github.com/mholt/PapaParse

AlexanderGeere (approved):
All working for me!

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+@property {Boolean} [params.async] If set to true, the data is uploaded asynchronously. Default is false.
+@property {Boolean} [params.header] If set to true, the first row is treated as a header row and not uploaded.
+
+@returns {<Promise>} The outcome object and any messages returned by the database.
+*/
+export default async function csvUpload(file, params = {}) {
+  if (!params.query) {
+    console.warn(`Query required`);
+    return;
+  }
```
RobAndrewHurst, line 36: Do you think we should throw an error here rather than just return??

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+
+  return new Promise((resolve) => {
+    const reader = new FileReader();
+
+    reader.readAsText(file);
+
+    reader.onload = async (e) => {
+      // Split text file into rows on carriage return / new line.
+      const rows = e.target.result.trim().split(/\r?\n/);
```
RobAndrewHurst, line 45: Could be a good idea to move this logic in the onload into its own function? Good for readability.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+
+      // The file has a header row.
+      if (params.header) {
+        const headerRow = rows.shift();
+
+        params.fields ??= headerRow.replaceAll('"', '').split(',');
+      }
+
+      const chunks = chunkRows(rows, params);
+
```
RobAndrewHurst, line 55: This is a bit difficult to read. Maybe we could break it up?

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+*/
+function chunkRows(rows, params) {
+  const chunks = [];
+
+  params.chunkSize ??= 1024 * 1024 * 4; //kbyte?
+  let chunkSize = 0;
+  let chunk = [];
+
+  rows.forEach((row) => {
+    // Split row into fields array.
```
RobAndrewHurst, line 97: What does kbyte mean to us? Are we asking if this is in kilobytes? Or could it be?

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+    const fields = splitRowIntoFields(row);
+
+    // Create row string for values from by passing field value to schema method.
+    const fieldsRow = `(${fields.map((v, i) => schemaMethods[params.schema?.[i] || 'Text'](v)).join()})`;
+
+    // Determine blob size of the row.
+    const rowSize = new Blob([fieldsRow]).size;
+
+    // Check whether the chunk would exceed lambda payload limit.
+    if (chunkSize + rowSize >= params.chunkSize) {
```
RobAndrewHurst, line 107: Maybe we could break this up as it's also quite complex to read.

RobAndrewHurst (changes requested):
Noticed some readability and complexity issues.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+@property {String} params.query The query to upload the data.
+@property {Boolean} [params.async] If set to true, the data is uploaded asynchronously. Default is false.
+@property {Boolean} [params.header] If set to true, the first row is treated as a header row and not uploaded.
+
+@returns {<Promise>} The outcome object and any messages returned by the database.
+*/
+export default async function csvUpload(file, params = {}) {
+  if (!params.query) {
+    console.warn(`Query required`);
+    return;
```
RobAndrewHurst, line 35: Should we rather throw an error here instead of a return.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+@param {array} rows An array of rows extracted from the CSV file.
+@param {object} params Parameter for chunking the rows array.
+@property {Number} [params.chunkSize] The chunk size in bytes to upload the data in. Default is 4MB.
+@property {Array} [params.schema] The schema array of column types.
+@returns {Array} An array of field arrays chunked to size.
+*/
+function chunkRows(rows, params) {
+  const chunks = [];
+
+  params.chunkSize ??= 1024 * 1024 * 4; //kbyte?
```
RobAndrewHurst, line 92: What does this mean to us?

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+    console.warn(`Query required`);
+    return;
+  }
+
+  return new Promise((resolve) => {
+    const reader = new FileReader();
+
+    reader.readAsText(file);
+
+    reader.onload = async (e) => {
```
RobAndrewHurst, line 43: Maybe we should move this logic into it's own named function to help with readability.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+  params.chunkSize ??= 1024 * 1024 * 4; //kbyte?
+  let chunkSize = 0;
+  let chunk = [];
+
+  rows.forEach((row) => {
+    // Split row into fields array.
+    const fields = splitRowIntoFields(row);
+
+    // Create row string for values from by passing field value to schema method.
+    const fieldsRow = `(${fields.map((v, i) => schemaMethods[params.schema?.[i] || 'Text'](v)).join()})`;
```
RobAndrewHurst, line 101:
```suggestion
const schemaType = params.schema?.[i] || 'Text';
const convertedValue = schemaMethods[schemaType](v);
const fieldsRow = `(${convertedValue.join()})`;
```
This might want to be broken up to make it easier to read.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+
+    reader.onload = async (e) => {
+      // Split text file into rows on carriage return / new line.
+      const rows = e.target.result.trim().split(/\r?\n/);
+
+      // The file has a header row.
+      if (params.header) {
+        const headerRow = rows.shift();
+
+        params.fields ??= headerRow.replaceAll('"', '').split(',');
```
RobAndrewHurst, line 51: This could also probably benefit from being broken up. Seems quite hard to read.

lib/utils/csvUpload.mjs
```diff
@@ -0,0 +1,194 @@
...
+The method returns a parameterised query object from the XHR utility method.
+
+@param {Array} data Data for the post request body.
+@param {Object} params Parameters for the post query.
+@returns {<Promise>} Returns promise from XHR query utility.
+*/
+function postRequestData(data, params) {
+  params.queryparams ??= {};
+
+  const qp = mapp.utils.queryParams(params);
```
RobAndrewHurst, line 183:
```suggestion
  const queryParams = mapp.utils.queryParams(params);
```
I think we need a more obvious variable name here.

RobAndrewHurst (changes requested):
I found some style changes and some complexity issues that we might want to address.

lib/utils/csvUpload.mjs
dbauszus-glx: No. An error should be thrown if a method errs. Warning is for missing configuration.

lib/utils/csvUpload.mjs
dbauszus-glx: Breaking up the replaceAll and the split method?

RobAndrewHurst: Yeah! If we think it's 100% necessary :)

lib/utils/csvUpload.mjs
dbauszus-glx: If an array must be an array it should be defined as array type. If an array property could also be a different type of object then object is fine. The same goes for boolean which are also type Object.

lib/utils/csvUpload.mjs
dbauszus-glx: This logic has been removed.

lib/utils/csvUpload.mjs
dbauszus-glx: Like so?

```js
const headerRowWithOutDoubleQuotes = headerRow.replaceAll('"', '')
params.fields ??= headerRowWithOutDoubleQuotes.split(',');
```

RobAndrewHurst: Sounds good!

dbauszus-glx: I'd like a second opinion on this. Personally I feel the suggestion to verbose.

RobAndrewHurst: It's just an option, but not necessary :)

RobAndrewHurst (approved):
All happy with this!

Issue #1621, opened 2024-10-29
Count of Locations Meeting Criteria When Filtering


simon-leech:
## Description

In the multi-filter plugin we have the Location Count for locations that meet the criteria of filtering. 
As all the components are provided in the core, the query, the functionality, we should add this to the default filtering on a layer as itsa  great feature!

## Type of Issue 
Please delete options that are not relevant, and select all options that apply. 
- ✅ New Feature

Issue #1434, opened 2024-08-27
Dictionary language lookup


dbauszus-glx:
A language lookup in the dictionary should default to the english key if requested language doesn't exist for the requested key.

The key word / phrase itself should be returned if this doesn't exist.

The dictionary method must not crash. No matter what is requested in which language.

Requesting key="layers" for language "es" crashes the mapp process.

![image](https://github.com/user-attachments/assets/2abe4c8c-4a6e-4742-be88-459f0f059398)

PR #1675, opened 2024-11-04
Filter Panel: Location match amount display

AlexanderGeere:
## Description

A count of the locations that match the filter criteria will now reflect on the filter panel.
Screenshot 2024-11-04 112035

This can be hidden by supplying: `layer.filter.location_count.hidden = true`.

## GitHub Issue
[#1621](https://github.com/GEOLYTIX/xyz/issues/1621)

## Type of Change

## How have you tested this? 
Tested locally on `bugs_testing/ui_elements/workspace.json`. Can tested on any layer with a filter. 

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


lib/ui/layers/panels/filter.mjs
```diff
@@ -15,6 +15,7 @@ mapp.utils.merge(mapp.dictionaries, {
     layer_filter_greater_than: 'Greater than',
     layer_filter_less_than: 'Less than',
     layer_filter_set_filter: 'Set filter',
+    layer_filter_location_match: 'Location(s) match your criteria'
```
RobAndrewHurst, line 18: this is going to want to wait for the Dictionaries issue #1434 as all these merges are not happening in the core anymore

RobAndrewHurst (changes requested):

lib/ui/layers/panels/filter.mjs
AlexanderGeere: I removed this for line for now.

lib/utils/locationCount.mjs
```diff
@@ -0,0 +1,19 @@
+export async function locationCount(layer){
```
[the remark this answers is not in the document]
AlexanderGeere: Done!

dbauszus-glx (dismissed):
Modules should not have a default export and export.

This looks odd.
```js
import filter, { updateCount } from './panels/filter.mjs'
```

I'd just export the filterPanel and updateCount as regular export from the module.

```js
import { filterPanel, updateCount } from './panels/filter.mjs'
```

And then assign the filterPanel method to the filter method property in the panels{}

I wouldn't bother with layer.filter.location_count.hidden flag and just show this. We don't have flags to hide the clear or reset button.

The locationCount module should be in the ui utils not the general mapp.utils since this is only used by ui methods.

RobAndrewHurst:
Is this something we want to be optional or even configurable?

simon-leech:
> Is this something we want to be optional or even configurable?

I don't think so - I think it should be core like the Clear and Reset buttons. Its a helpful feature to provide for UX

RobAndrewHurst:
> > Is this something we want to be optional or even configurable?
> 
> I don't think so - I think it should be core like the Clear and Reset buttons. Its a helpful feature to provide for UX

My thought behind this was to include custom count messages, and also the option to turn off if we have crazy data sets that we just don't want to include

RobAndrewHurst (approved):

dbauszus-glx (changes requested):
I still want to look at this.

dbauszus-glx (approved):
This wants another review.

The updateCount method does not have to await the layer and does not need to return the layer. The method should not exported as layers.panel.updateCount which is meaningless.

The method is a filter method and should be assigned as mapp.ui.layers.filters.updateCount(layer)

This method should be called on the changeEnd event which is triggered by the applyFilter method.

The changeEnd event listener is assigned in the filterPanel method when we know that the view as well as the count element do exist.

The locationCount utility method should respect if the viewport flag is set in the layer.filter configuration. This enables viewport queries on layers such as MVT which do not have a viewport queryparams.

simon-leech (changes requested):
Found a few little issues: 

1. When the layer is off and you expand the filter panel - you end up just seeing the 'Locations match your criteria' text. This should be hidden until you turn on the layer. 
image
2. When you turn on the layer, the number is not updated until you zoom in / out - this should be updated when you toggle display of the layer. 
3. I'm seeing a lot of errors in the console from the location_count query, it looks like the table parameter is not being passed to it. 
4. Can we update the documentation to make it clear that you need to pass 'viewport' to get the number updating when you zoom in / out - as this is basically required otherwise its pretty confusing for the user.

RobAndrewHurst (approved):

simon-leech (approved):
All good with me :)

Issue #1679, opened 2024-11-05
svgTemplates not being passed to mapview


RobAndrewHurst:
## Description

The svgTemplates from the workspace are not being passed from the locale into the mapview promise.
The reference on the locale is also incorrect. It should be 'svgTemplates' not svg_templates for consistency.

## Type of Issue 
- ✅ Bug

PR #1680, opened 2024-11-05
Fix svgTemplates reference in mapview

RobAndrewHurst:
# Pull Request Template

## Description
The mapview on the default script was not passing in the svg_templates in from the locale.

There is inconsistency when it comes to the naming of different variables. It's a mix between snake_case & camelCase.

```js
syncPlugins,
svg_templates
```

We might want to discuss how we handle this for the future, and set some eslint checks. 

## GitHub Issue

#1679 

## Type of Change

## How have you tested this? 
You can run the local tests

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Updated Existing Tests

### DIFF
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -55,7 +55,7 @@ The mapview decorator may return the async mapviewPromise method which must be a
 @param {object} mapview JSON params for a new mapview.
 @property {string} [mapview.host=mapp.host] The host domain/path for queries.
 @property {object} mapview.locale The locale defintion for the mapview.
-@property {array} [mapview.svgTemplates] SVG templates to load as mapp.utils.svgSymbols.templates
+@property {array} [mapview.locale.svgTemplates] SVG templates to load as mapp.utils.svgSymbols.templates
 @property {array} [mapview.syncPlugins] locale.plugins[] must be loaded prior to the execution of syncPlugins.
 
 @returns {mapview} Decorated Mapview.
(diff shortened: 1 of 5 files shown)


AlexanderGeere (approved):
Looks good to me. Im a fan of camel case, so plus one for that.

RobAndrewHurst:
> The mapview decorator method should not require svgTemplates nor syncPlugins as mapview properties.
> 
> 
> 
> It should be enough for these properties to be defined on the locale.
> 
> 
> 
> The loadPlugins flag feels legacy as it shouldn't matter.
> 
> 
> 
> The mapviewPromise check should be conditional on whether the locale has syncPlugins or svgTemplates.

It's always a promise because syncPlugins is defaulted to an array on the _default.html view

dbauszus-glx:
I fixed the mapview properties https://github.com/GEOLYTIX/xyz/pull/1680/commits/716dd4a69eebf503483a74aaeac9beb1e183d0ef

and update from patch branch for the bcrypt removal.

dbauszus-glx (approved):
This is now working as expected and much clearer.

RobAndrewHurst:
> If svg_templates is legacy naming, should we be warning on it and assigning it to the svgTemplates (that's currently happening)?
> 
> Otherwise it will never be removed as it's not going to be obvious that anything needs changing?

It's not legacy really. It's still used, but in the code it doesn't actively reference svg_templates it makes it svgTemplates in the mapview locale for consistency. I think going forward we must try enforce camelCase rather than snake_case if that's what we want.

dbauszus-glx:
@simon-leech Legacy maybe the wrong term. It's an alternative spelling. Adding the warning would be additional code which we could never remove and we would always have to keep the assignment from one to the other. Hence we just do the assignment with a reference in the documentation.

simon-leech (approved):

PR #1681, opened 2024-11-05
Template fields fieldsmap

dbauszus-glx:
Layer query templates should not parse the workspace for templates matching fields.

The _.label hasn't been used for a while since the label has been added to the fields parameter.

The fieldsMap method has been added to the query module. The method is called from the layerQuery method and assigns the fieldsMap Map object to the request parameter. The method attempts the lookup of a workspace.templates template string.

The infojMap method has been added to the query module. The method is called from the layerQuery method and assigns the infojMap Map object to the request parameter. The method attempts the lookup of a workspace.templates template string.


RobAndrewHurst:
The infojMap method has a [for..of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop that was using the incorrect syntax to get to the next iteration. 

It was using `break` which would stop the execution and go to the next first statement after the loop.

To fix this we need to use `continue`. `continue` stops the execution and goes to the next iteration of the loop.

RobAndrewHurst (approved):

AlexanderGeere (approved):

simon-leech (changes requested):
Ongoing review

simon-leech (approved):
I wrapped the field in brackets when pushed into the fieldsMap and infojMap and that resolves the issue I noticed - working for me now. 

Will need re-review by others so I have requested  that tooo

RobAndrewHurst (approved):
All happy with this!

PR #1685, opened 2024-11-06
Update feature_info plugin with css

dbauszus-glx:
A default css will now be assigned to the plugin config which is set on the content element.

```js
plugin.css ??= 'padding: 1em; max-width: 250px; overflow: auto;'
```

image

The plugin is now fully documented.

### DIFF
--- a/lib/plugins/feature_info.mjs
+++ b/lib/plugins/feature_info.mjs
@@ -1,9 +1,14 @@
 /**
- *  
+## /plugins/feature_info
+
+The feature_info plugin module exports the feature_info plugin method.
+
 Dictionary entries:
 - toolbar_current_location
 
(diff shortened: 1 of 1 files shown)


AlexanderGeere (changes requested):
Clicking to disable the plugin throws an error: 
`e.btn.firstChild.classList is undefined`
which can be fixed with
`plugin.btn.firstChild.classList?.remove('active')`

On line 100 in the callback

dbauszus-glx:
@AlexanderGeere It's firstElementChild innit? Can you check against last commit.

AlexanderGeere (approved):

simon-leech (approved):
All working for me 😅

PR #1689, opened 2024-11-07
acl is null checks and module export

dbauszus-glx:
A condition on null is more expressive than a condition to check whether a module is not a function.

The acl module should return null instead of an empty object if the ACL cannot be initialised.

The acl module must return null before checking whether the connection variable is an array split from the PRIVATE or PUBLIC process.env.

This must be documented in the ACL module docs.


dbauszus-glx:
I removed the cognitive overload in the login mod by creating the loginBody method.

RobAndrewHurst (approved):

AlexanderGeere (approved):

PR #1691, opened 2024-11-07
Warn on entry.layer not in locale

dbauszus-glx:
The layer entry as a replacement for the mvt_clone will have an entry.layer property used to lookup a layer in the mapview.locale.

The [layer] entry method must return and should warn if the entry.layer property is defined but the layer is not found in the mapview.locale.

I also noticed that the [entry] layer.tables zoom restriction is not respected on first execution of the method but only once the changeEnd event is executed. This has been fixed in this patch PR as well.

### DIFF
--- a/lib/ui/locations/entries/layer.mjs
+++ b/lib/ui/locations/entries/layer.mjs
@@ -43,6 +43,12 @@ export default function layer(entry) {
     const layer = entry.mapview.locale.layers
       .find(layer => layer.key === entry.layer)
 
+    if (!layer) {
+
+      console.warn(`Layer [${entry.layer}] not found in mapview.locale`)
+      return;
+    }
+
     // Spread locale layer into entry.
     entry = {
       ...structuredClone(layer),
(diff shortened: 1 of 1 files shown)


lattaai3:
Hello, I tried to fix the issue

This is what I did:

Modified the layer entry method to warn and return if the entry.layer is defined but not found in mapview.locale. Also fixed the zoom restriction issue for layer.tables.

I also created pull request: https://github.com/GEOLYTIX/xyz/pull/1693


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

dbauszus-glx:
I notice another bug testing this with a featureSet clone. A [layer] entry may have a featureSet or featureLookup object but not both. Otherwise the featureSet will be ignored.

dbauszus-glx:
The open_library function receives argument e, documented as param event which is never used.

image

What is the icon param for the icon_click method?

image

I just don't understand the method. 

image

AlexanderGeere (approved):
Added tests for checking the zoom restriction and the warning. 

Looks good to me!

simon-leech (approved):
Working correctly for me :)

Issue #1607, opened 2024-10-28
S3 provider/signing module


dbauszus-glx:
The S3 provider should be moved to the signing API as it signs requests and does not receive payload.

The method wants to be fully documented.

PR #1698, opened 2024-11-12
API Sign: Change the S3 Provider into a signer

AlexanderGeere:
## Description
Provides an api endpoint `api/sign/s3` Which returns a signed URL. 

## GitHub Issue
[#1607](https://github.com/GEOLYTIX/xyz/issues/1607)

## Type of Change
- ✅ Documentation

## How have you tested this? 
Tested with `bugs_testing/sign/s3_workspace.json`

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


mod/sign/s3.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  if (!Object.hasOwn(commands, req.params.command)) {
+    return res.status(400).send(`S3 command validation failed.`)
+  }
+
+  return commands[req.params.command]()
+}
+
+async function objectAction(reqParams, objectCommand) {
```
RobAndrewHurst, line 63: We need some documentation on this function

mod/provider/s3.js
```diff
@@ -1,102 +0,0 @@
-/**
```
RobAndrewHurst, line 1: I think let's also reroute the user with a deprecation warning for this endpoint.

RobAndrewHurst (dismissed):

dbauszus-glx (dismissed):
The S3 library should be installed as an optional dependency. The module should export null if the dependency is not available using npm install --no-optional.

dbauszus-glx:
How are requests handled where signing fails? eg. missing resource, missing credentials.

mod/provider/s3.js
AlexanderGeere: Addressed in latest push 👍

AlexanderGeere:
> How are requests handled where signing fails? eg. missing resource, missing credentials.

Signing will always succeed even if the credentials are wrong, calling the signed url will return a 403.

mod/sign/s3.js
```diff
@@ -0,0 +1,149 @@
...
+let
+  S3Client,
+  PutObjectCommand,
+  GetObjectCommand,
+  DeleteObjectCommand,
+  ListObjectsCommand,
+  getSignedUrl;
+
+//Check for credentials 
+if (!process.env?.AWS_S3_CLIENT) {
```
RobAndrewHurst, line 19: How do we handle public buckets that don't require any credentials?

package.json
```diff
@@ -20,10 +20,12 @@
     "test-docs": "jsdoc --configure jsdoc_test.json --verbose"
   },
   "license": "MIT",
-  "dependencies": {
+  "optionalDependencies": {
```
RobAndrewHurst, line 23: Moved the dependencies into the optionalDependencies object.

if we want to not install the dependencies we will have to give the flag `--omit=optional` when installing

RobAndrewHurst: > `npm install --no-optional` should do the same.

This works but is being phased out. It started getting phased out from npm 7. It still works but they want developers to start using --omit=optional

RobAndrewHurst: The omit flag was introduced because it's more flexible. 

You can do things like

`npm install --omit=dev,optional`

RobAndrewHurst: We would want to start adding this kind of install command on certain instances. So that if we know we don't need the s3 bucket packages and all of the dev dependencies we would just use one flag instead of 

`npm install --no-optional --omit=dev`

dbauszus-glx:
I don't understand this bit. Can you comment on this for me to test.

```js
  if (provider[req.params.provider]?.deprecated){
    return await provider[req.params.provider].fn(req, res)
  }
```

mod/sign/s3.js
AlexanderGeere: I'll test what happens with public buckets

dbauszus-glx:
I have updated the S3 module docs with reference to the credentials required from AWS and optional dependencies.

image

mod/sign/s3.js
AlexanderGeere: Public buckets can just be hit on the url. [7071f1b](https://github.com/GEOLYTIX/xyz/pull/1698/commits/7071f1b0c6fa9ec563c9c415a5ed5570134cd541) provides a function that returns the public url if no credentials are supplied

RobAndrewHurst:
Happy with that! :)

dbauszus-glx:
I bumpred the optional dependencies to their current versions 3.691.0

AWS recommends to use ListObjectsV2Command in development. The ListObjectsCommand is only kept for legacy purposes.

I also changed the s3 signer to return the signed URL as a text which it is.

dbauszus-glx:
Rather than providing command maps, and action dictionaries to translate commands and properties back and forth it will be much easier to talk directly with the SDK using the commands as documented in the S3 documentation.

https://github.com/GEOLYTIX/xyz/pull/1698/commits/9cce4f17d1e651693a0ff978a04a02bb25db2baf

dbauszus-glx (approved):

RobAndrewHurst (approved):
Nice I am happy with this! 

I added in a note for developers that they must not use this sign if they are interacting with a public bucket, and gave them an example of how they could use it.

simon-leech:
Don't feel like I'm best placed to review this one so removed my request. If you two @dbauszus-glx and @RobAndrewHurst are happy, that's great with me!

PR #1708, opened 2024-11-13
Roles should be stripped from layer JSON

dbauszus-glx:
Roles should be stripped from locale and or layer JSON.

I have updated the the utility method docs.


dbauszus-glx:
The notIncludesNegatedRole should be unnested and documented.

https://github.com/GEOLYTIX/xyz/pull/1708/commits/234f709528ba76732d0ae43fe00988274ab98d57

RobAndrewHurst (approved):

PR #1713, opened 2024-11-14
Shortcircuit keyvalue_dictionary parseObject on mapview key

dbauszus-glx:
A mapview with the locale itself may be assigned to a plugin object.

This would cause the keyvalue_dictionary parse method to parse the locale object until the maximum callstack size is exceeded.

I also provide an example for the keyvalue_dictionary in the module jsdocs block.

### DIFF
--- a/lib/plugins/keyvalue_dictionary.mjs
+++ b/lib/plugins/keyvalue_dictionary.mjs
@@ -45,7 +45,7 @@ Parses an object and replaces its string values with dictionary entries.
 function parseObject(obj, dictionary) {
   for (const [key, value] of Object.entries(obj)) {
 
-    // Prevents crash where the mapview with the local itself maybe nested in a locale plugin object.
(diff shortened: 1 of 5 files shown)


lib/plugins/keyvalue_dictionary.mjs
```diff
@@ -30,6 +44,10 @@ Parses an object and replaces its string values with dictionary entries.
 */
 function parseObject(obj, dictionary) {
   for (const [key, value] of Object.entries(obj)) {
+
+    // Prevents crash where the mapview with the local itself maybe nested in a locale plugin object.
```
simon-leech, line 48: Should this comment read 'locale object' instead of 'local object' ?

simon-leech (approved):
I correct a comment and tested this - working as expected and no more crashes :)

RobAndrewHurst (approved):
lgtm

PR #1714, opened 2024-11-15
Add layer.featureFormat option

dbauszus-glx:
It should be possible to define the layer.featureFormat in order to allow for custom featureFormats without the need to create a custom layer format.

The featureFormat for vector layer will be assigned from the format if not explicit.

This PR allows to set the featureFormat in the layer json.

```js
"circles": {
  "name": "Center of Mass Circles",
  "format": "geojson",
  "featureFormat": "centerOfMassCircle",
  "template": "nuts3",
  "attribution": {
     "© eurostat": "https://ec.europa.eu/eurostat/web/nuts",
     "turf.js": "https://turfjs.org/"
  }
}
```


simon-leech (changes requested):
Review in progress; adding tests

simon-leech (approved):
All good with me - I pushed a test update too that: 
1. Ensure that a base vector layer has a featureFormat defined (nullish assignment testing here).
2. Create a layer with a custom featureFormat and ensure this is correctly provided.

dbauszus-glx:
This could be tested by creating a mock featureFormat.

This for example would add the layer_key property to each feature instead of the properties from json feature.

```js
function geojson_layer(layer, features) {

  const formatGeojson = new ol.format.GeoJSON

  return features.map((feature) => {

    return new ol.Feature({
      id: feature.id,
      geometry: formatGeojson.readGeometry(feature.geometry, {
        dataProjection: 'EPSG:' + layer.srid,
        featureProjection: 'EPSG:' + layer.mapview.srid,
      }),
      layer_key: layer.key
    })

  })
}
```

RobAndrewHurst (approved):
All happy with this!

PR #1724, opened 2024-11-18
mapview.addLayer merge/clone issue

dbauszus-glx:
I was under the assumption that mapp.utils.merge() creates a clone if objects are merged into an empty object. This is not the case.

In this PR I pass structured clones of the layer and locale.layer to the merge() utility method in order to make sure that the layer added to the mapview is indeed a clone.

### DIFF
--- a/lib/mapview/addLayer.mjs
+++ b/lib/mapview/addLayer.mjs
@@ -45,7 +45,7 @@ export default async function addLayer(layers) {
   for (let i = 0; i < layers.length; i++) {
 
     // The JSON layer is merged into the locale.layer
-    const layer = mapp.utils.merge(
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):
Tests are a passing. Observing the correct behaviour in the bugs testing workspace.

simon-leech (approved):
This is working as expected for me - all the layers are added correctly, correct themes applied and legends in the correct order

RobAndrewHurst (approved):
I am all happy with this PR. 

I added in a check to see if a layer is identical to another layer even if they have the same configurations. 

This will expect to resolve to false, as they cannot be directly equal.

Issue #1651, opened 2024-11-01
Cloudfront Signer


dbauszus-glx:
A cloudfront signer module should provide the signedUrl to the cloudfront provider.

The cloudfront signer and provider modules must export null if the KEY_CLOUDFRONT is undefined.

Failed fetch or failed signing requests must return the error instead of failing with just logging the error to the process console.

PR #1727, opened 2024-11-19
Signing API: Cloudfront Signer

AlexanderGeere:
## Description
Provides a signing api for requests to cloudfront. The provider will now use the signing api to get the signed url to make requests to cloudfront

## GitHub Issue
[#1651](https://github.com/GEOLYTIX/xyz/issues/1651)

## Type of Change

## How have you tested this? 
This can be tested by changing the plugin key in an env to `cloudfront` instead of `file` and using the cloudfront url instead of the local file reference in the `PLUGINSREPO` key. 

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


dbauszus-glx:
I added status codes for returns from the provider and signer module.

404 for requests for modules that don't exist. eg. foo

405 for requests where the module existy but is not configured.

The message reflects the status and prints the missing param.

https://github.com/GEOLYTIX/xyz/pull/1727/commits/f1a42a1a6ec0adc2e783629feecb1f4d6d64f64e

dbauszus-glx:
image

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/1727/commits/1ad418729376ec92e4e519ac109a54ad2615818b

The privateKey should read in the try..catch the catch should always console.error the error and export null not throw an error and take the process down.

image

dbauszus-glx (changes requested):
It's almost there. Please review my changes in previous comments.

You need to manage the err caught in the signer. If I providea key with the right name but wrong content I get two error messages and the process dies.

image

There should be one logged error which is passed back from the signer and handled in signer/provider module.

AlexanderGeere:
@dbauszus-glx [2e9c1eb](https://github.com/GEOLYTIX/xyz/pull/1727/commits/2e9c1eb982892b970ce2e821c580bf525caeac55) Should prevent the process from crashing and you should only see one error per failed request.

RobAndrewHurst (approved):

Issue #1717, opened 2024-11-15
Layers not cloned when added to multiple map views


simon-leech:
When using a custom view that creates multiple map views, there is some odd behaviour with layers added to map views. 

Workflow 
- Get the locale 
- Set layers to locale.layers
- Create a map view 
- Add layers to it
- Create second map view 
- Add same layers to this
- Change a theme on the second map view on a layer, the behaviour does not work, it may change the theme or perhaps the theme is reversed multiple times incorrectly too as the layer is decorated over and over. 

Will provide a replicable example next week

RobAndrewHurst:
I am looking into this!

PR #1728, opened 2024-11-19
Template bug

RobAndrewHurst:
## Description

> The layerQuery must be run before the getTemplate request in the query module. This ensures that a query template defined in the layer [template] will be loaded if the workspace is not cached.
> 
> This can only be tested to pass.

## GitHub Issue

#1717

## Type of Change
- ✅ Testing

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -35,7 +35,7 @@ const getLayer = require('./workspace/getLayer');
 @description
 The [SQL] query method requests a query template from the getTemplate method and checks whether the requesting user is permitted to execute the query.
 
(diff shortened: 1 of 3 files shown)


dbauszus-glx:
The layerQuery must be run before the getTemplate request in the query module. This ensures that a query template defined in the layer [template] will be loaded if the workspace is not cached.

This can only be tested to pass.

simon-leech (approved):
This is working for me - I moved some templates from the workspace onto the layer and re-deployed the application and they are correctly loaded in. 

i tested this both in default MAPP and also in a custom view and its working on both

Issue #1718, opened 2024-11-18
Ability to hide the `zoomToExtent` button on a zoom-restricted layer


simon-leech:
## Description

Currently we always add the `zoomBtn` if the layer is zoom restricted. 
However, we should allow configuration to turn this off if the functionality is not required / contextually appropriate. 

``` js
 // Create a div for the magnifying glass icon
  layer.zoomBtn = layer.tables
    && mapp.utils.html.node`<button 
      data-id="zoom-to"
      title=${mapp.dictionary.zoom_to}
      class="mask-icon search"
      onclick=${() => zoomToRange(layer)}>`

```

I propose we instead do this, by making it an OR assignment we can make it `false` to hide it.
e.g. 
``` json
"zoomBtn": false
```

``` js
 // Create a div for the magnifying glass icon
  layer.zoomBtn ||= layer.tables
    && mapp.utils.html.node`<button 
      data-id="zoom-to"
      title=${mapp.dictionary.zoom_to}
      class="mask-icon search"
      onclick=${() => zoomToRange(layer)}>`

```

([Asana Ticket](https://app.asana.com/0/1208315489758086/1208753705495099/f))
## Type of Issue 
Please delete options that are not relevant, and select all options that apply. 
- ✅ New Feature

dbauszus-glx:
This shouldn't be shown by default and not shown with the false flag which is counter intuitive to mapp json.

The default should be to not show the button and show if configured.

This also opens a can of worms in regards to other elements. If there is an option for the zoomButton, should the zoomToExtentBtn be optional as well.

Should all of these flags be add to the layer JSON or perhaps should there be a config object.

layer.view === null, no view is created. Otherwise view will be the view element. We cannot change this.

dbauszus-glx:
The only option is see to not break the current behaviour without spamming more and more flags into the layer json is by creating a default object layer.viewConfig = {zoomBtn, zoomToExtentBtn, panelOrder} etc. And you can provide viewConfig taking away the options to change the behaviour.

dbauszus-glx:
All this should happen in a viewConfig method in the module to reduce the complixity of the view method exported from the module.

PR #1741, opened 2024-11-21
Layer View: Hide zoom button 

AlexanderGeere:
## Description
The layer zoom does not always make sense contextually. This PR gives us the ability to hide it by specifying `layer.zoomButton`.

Specifying `layer.zoomButton` should turn this:
Screenshot 2024-11-21 094211

into:
Screenshot 2024-11-21 094333

## GitHub Issue
[#1718](https://github.com/GEOLYTIX/xyz/issues/1718)

## Type of Change
- ✅ Documentation

## How have you tested this? 
Can be tested locally on any instance that has a zoom restricted layer.

## Testing Checklist 
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/view.mjs
+++ b/lib/ui/layers/view.mjs
@@ -73,6 +73,7 @@ The layerView method will create and assign a layer.view node to the layer objec
 @param {layer} layer A decorated mapp layer.
 @property {Object} layer.view No layer view will be created with the view property null.
 @property {Object} layer.drawer The layer view will not be in a drawer element with the drawer property null.
+@property {Boolean} [layer.zoomButton] Controls whether the layers zoom button is used or not.
 @property {array} [layer.panelOrder=['draw-drawer', 'dataviews-drawer', 'filter-drawer', 'style-drawer', 'meta']] The order of layer view panel elements.
 
 @returns {layer} The layer object is returned after the layer view has been created.
@@ -138,12 +139,15 @@ export default function layerView(layer) {
     }}>`
 
   // Create a div for the magnifying glass icon
-  layer.zoomBtn = layer.tables
+  layer.zoomButton ??= layer.tables
     && mapp.utils.html.node`<button 
       data-id="zoom-to"
       title=${mapp.dictionary.zoom_to}
       class="mask-icon search"
       onclick=${() => zoomToRange(layer)}>`
+  
+  //If zoomButton is false, set it to undefined so nothing gets rendered
+  layer.zoomButton ||= undefined
 
   // Add on callback for toggle button.
   layer.showCallbacks.push(() => {
@@ -159,7 +163,7 @@ export default function layerView(layer) {
     <h2>${layer.name || layer.key}</h2>
     ${layer.zoomToExtentBtn}
     ${layer.displayToggle}
-    ${layer.zoomBtn}
+    ${layer.zoomButton}
     <div class="mask-icon expander"></div>`
 
   // An empty drawer element can not be expanded.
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
I understand and it is working to set `layer.zoomButton` to false to remove it. 

However setting it to `true` errors. I think we need to explicitly check for a value of `false` and otherwise include it. 

Would it be possible too to add a test for this to ensure its working as expected going forwards?

dbauszus-glx (changes requested):
I don't like setting false flags. The default should be false, and the zoomButton should only be added if the flag is set to true. I'll update the issue.

simon-leech:
We can't change the default functionality as in 99% of cases this is required. could we instead introduce a new flag `hideZoomButton` and set that to true only when required to hide it? Otherwise default would be to show it?

dbauszus-glx:
@simon-leech  I update the ticket. The spec shouldn't be worked out in the PR.

AlexanderGeere:
@dbauszus-glx Created a viewConfig function which will default everything to on. This function also adds the various button the layer view

@simon-leech This now works as follows: to turn off a particular button you should add `layer.viewConfig` and then exclude the button you wish to turn off.  

zoomToExtentBtn, panelOrder,displayToggle, zoomBtn these are the four things you can control.

simon-leech (changes requested):
Not providing a `panelOrder` flag crashes the process. 
`panelOrder` is not optional true / false, it should have a default array order or be possible to overwrite that order. 

Also `layer.panelOrder` is currently supported so this must be assigned to the `viewConfig` object and warned on to prevent a breaking change.

simon-leech (approved):
I'm happy with this now - tested the various options and pushed some updated tests to test `panelOrder`, `zoomBtn` and `displayToggle`.

RobAndrewHurst (approved):

dbauszus-glx (changes requested):
viewConfig.panelOrder should be an array property, not boolean. The provided array should be used instead of the default.

It should be possible to define the panelOrder in the layer json or inside the viewConfig in the layer json. If defined directly in the layer json the config should be moved [not copied] into the viewConfig.
```js
panelOrder: ['meta', 'style-drawer']
```

`layer.drawer` should be assigned to the viewConfig as the check is after the viewConfig method.

layer.classList should be assigned to the viewConfig as well.

dbauszus-glx:
@AlexanderGeere re panelOrder. https://github.com/GEOLYTIX/xyz/pull/1741/commits/3fcc53652ca67745bc15aa7c47deacb63d99e3bchttps://github.com/GEOLYTIX/xyz/pull/1741/commits/3fcc53652ca67745bc15aa7c47deacb63d99e3bc

AlexanderGeere:
> viewConfig.panelOrder should be an array property, not boolean. The provided array should be used instead of the default.
> 
> It should be possible to define the panelOrder in the layer json or inside the viewConfig in the layer json. If defined directly in the layer json the config should be moved [not copied] into the viewConfig.
> 
> ```js
> panelOrder: ['meta', 'style-drawer']
> ```
> 
> `layer.drawer` should be assigned to the viewConfig as the check is after the viewConfig method.
> 
> `layer.classList` should be assigned to the viewConfig as well.

I moved the layer.drawer and layer.classList into viewConfig, so the layer.drawer is created in viewConfig now.

dbauszus-glx:
I updated the viewConfig conditions.

https://github.com/GEOLYTIX/xyz/pull/1741/commits/9ff57f9075ddde18c6174c2e552b0627aaa3be26

The drawer header button must be created in the view method after the layer.drawer condition check since the header might not exist.

dbauszus-glx (approved):
This is working for me.

@AlexanderGeere  can you run another check with the last commits to the viewConfig method.

AlexanderGeere:
> This is working for me.
> 
> @AlexanderGeere can you run another check with the last commits to the viewConfig method.

@dbauszus-glx Looks good to me

PR #1742, opened 2024-11-21
Check for layer query template

dbauszus-glx:
The workspace must be cached prior to be required in the fieldMaps method.

The layer query can not check for template.layer property.

This must happen after the layer query and the getTemplate method have been executed.


simon-leech (approved):
Working for me

Issue #1716, opened 2024-11-15
Template loading issues


simon-leech:
## Description

This is also related to template objects not being merged if they are nested.

Templates defined on a layer in a `template` object aren't always found on a vercel deployment.
Sometimes they return a 500 error on DEV, but if you open up the query in a new tab, there is a data returned. 
This is a required fix as we made the change for templates being moved to layers instead of the workspace. 
Example below 

``` json
  "dataviews": {
        "hide": true,
        "dv": {
            "label": "Query",
            "display": true,
            "query": "query_path",
            "template": {
                "key": "query_path",
                "src": "filepath.sql"
            },
            "target": "tabview",
            "viewport": true,
            "mapChange": true,
            "queryparams": {
                "filter": true
            },
            "toolbar": {
                "download_csv": true
            },
            "events": {
                "rowClick": {
                    "util": "select",
                    "zoomToLocation": true
                }
            },
            "table": {
                "placeholder":  "Hello",
                "layout": "fitColumns",
                "selectable": true,
                "pagination": true,
                "paginationSize": 50,
                "columns": []
}
}
}
```

## Type of Issue 

- ✅ Bug

RobAndrewHurst:
I think the reason this is happening is because of serverless Cold Starts. 

Vercel's serverless functions start fresh on each cold start, resetting our module-level variables. I think we might need to use a more persistent caching solution. Something in vercel's environment like vercel's KV.

dbauszus-glx:
Is this the workspace variable? Cold starts shouldn't have an effect on this. The variable would be undefined which should be the same as the age. The only thing I could think about is the age being set low and the workspace age is reached during the process of query request.

RobAndrewHurst:
> Is this the workspace variable? Cold starts shouldn't have an effect on this. The variable would be undefined which should be the same as the age. The only thing I could think about is the age being set low and the workspace age is reached during the process of query request.

in the tests conducted there is no age

RobAndrewHurst:
When running "vercel dev" locally I can see the workspace is cache multiple times on one load of the page

dbauszus-glx:
This is a bug in the query module. A query template maybe assigned to the workspace when the layerQuery method is executed. The layerQuery method requires a [query] template.  The query template may not be accessed from the workspace if the layer has not been cached yet. This may happen when a new process is spawn. `vercel dev` local can replicate this since a new process is spawned for every request.

PR #1743, opened 2024-11-21
fix merge template object issue

RobAndrewHurst:
# Object Template parsing correction

The `assignWorkspaceTemplate` function was calling itself with the actual values of each object key, which would result in the templates never to be assigned to the workspace.templates as the key & value is needed on the entry to make the assignment on the workspace.

## GitHub Issue

#1716 

## Type of Change


dbauszus-glx (approved):
The fixes the observed bug.

simon-leech (approved):
Working for me - thanks @RobAndrewHurst

Issue #1694, opened 2024-11-08
Date and datetime format methods are private in date entry.


cityremade:
At the moment unix timestamp => date/datetime formatters are private inside date element module.

Conversion from string date to unix is straightforward like this:
`Math.floor(new Date(<string date>).getTime() / 1000)`

Yet going from unix timestamp into string date is more problematic as editable date input only takes specific format.

How to solve:

inside ui/locations/entries/date.mjs `formats` (line 8) moved into utils so that can be reused outside date entry.

**Update**: the plugin in question for apps v480 assigns these methods to mapp.utils so that they can be used in the plugin context.

Still to implement for the framework.

dbauszus-glx:
ui/locations/entries/date.mjs must also be documented. It would be good to move the dateString, timeString, and formats object/method to their own utils module.

```js
const dateString = entry => new Date((entry.newValue || entry.value) * 1000)
  .toLocaleDateString(entry.locale, entry.options); // day only as configured in workspace

const timeString = entry => new Date((entry.newValue || entry.value) * 1000)
  .toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' }); // time only as 00:00

// formatted value for input element.
const formats = {
  datetime: entry => `${new Date((entry.newValue || entry.value) * 1000).toLocaleDateString('fr-CA')}T${timeString(entry)}`, // YYYY-MM-DDT00:00
  date: entry => `${new Date((entry.newValue || entry.value) * 1000).toLocaleDateString('fr-CA')}` // YYYY-MM-DD
}
```

I suggest a temporal module as there will be a temporal web api in the future.

eg.

```js
mapp.utils.temporal = {
  dateString,
  timeString,
  datetime, // from formats object
  date // from formats object
}
```

RobAndrewHurst:
> ui/locations/entries/date.mjs must also be documented. It would be good to move the dateString, timeString, and formats object/method to their own utils module.
> 
> ```js
> const dateString = entry => new Date((entry.newValue || entry.value) * 1000)
>   .toLocaleDateString(entry.locale, entry.options); // day only as configured in workspace
> 
> const timeString = entry => new Date((entry.newValue || entry.value) * 1000)
>   .toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' }); // time only as 00:00
> 
> // formatted value for input element.
> const formats = {
[...]

PR #1748, opened 2024-11-21
Temporal module;

cityremade:
Date formatting functions placed into one object within mapp.utils.temporal{}.

Previously they were only available inside date/datetime location entries and inaccessible outside.
With date and times stored as unix epoch they can be accessed as utilities in custom scripts/views/plugins.

Adds function which converts date string to unix epoch which previously wasn't reusable.

## GitHub Issue
https://github.com/GEOLYTIX/xyz/issues/1694

Tested locally.
Non-breaking change as these methods were not used outside date/time entry module.
Updated documentation on date formatting;

All date and datetime read only and editable entries should receive correct formatting
in location view and editable date inputs.


simon-leech:
@cityremade Can you change the base branch - this shouldn't be main. It should be patch (v4.12.4) or minor (v4.13.0)? 😅

dbauszus-glx (changes requested):
I have updated the docs for the entry/date module.

I also added the properties for the params in the temporal utils module.

I don't fully understand the `fr-CA` default? Should this not be english as default derived from the user language.

lib/utils/temporal.mjs
```diff
@@ -0,0 +1,93 @@
...
+Returns time only as 00:00.
+
+@param {infoj-entry} entry type:date or type:datetime infoj-entry typedef object.
+@property {integer} entry.value date[time] as unix epoch.
+@property {integer} [entry.newValue] updated entry.value which has not yet been saved.
+@returns {String}
+*/
+function timeString(entry) {
+  return new Date((entry.newValue || entry.value) * 1000)
+    .toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' })
```
RobAndrewHurst, line 47: image

we are going to want to use the [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) object rather as this is going to be way more efficient.

RobAndrewHurst (changes requested):

cityremade:
In order to set the date on date input we need it YYYY-MM-DD format hence `fr-CA`.
These methods make it possible to set current value on input of type date, otherwise we get empty fields.

lib/utils/temporal.mjs
cityremade: does it mean formatting by toLocaleString should be replaced entirely?

RobAndrewHurst: Yeah! We did the same for the numeric formatter

cityremade:
String date formatting now handled by `Intl.DateTimeFormat`
Date/date time inputs now use only naive ISO Date derived from unix timestamp.
timeString function removed.
All dates are handled by one function dateString and custom values will be applied in entry configuration according to Js docs ( entry.locale + entry.options{} )

```
new Intl.DateTimeFormat('en-GB', {
    dateStyle: 'full',
    timeStyle: 'long',
    timeZone: 'Australia/Sydney',
  }).format(date),
);
// Expected output: "Sunday, 20 December 2020 at 14:23:16 GMT+11"
```

RobAndrewHurst:
I renamed the module 'temporal' as it was a bit vague to the layman of what the module is actually doing. So I gave it a clearer name of `dateTime`

RobAndrewHurst:
I have also update the object name passed into the function on the utility. It was specifying that the object was only going to come from infoj entries which shouldn't be true for a utility that can be used by other elements potentially. So I have made the object passed in a more generic approach that can apply to more situations than just infoj entries.

RobAndrewHurst (approved):
I am all happy with this!

I have added in tests to check each function.

dbauszus-glx:
The dateTime utils referenced in the date entry module needed changing to temporal.

dbauszus-glx (approved):
Works well. Only change I did was to take the language as locale from the mapp.user if not explicit on entry [params].

This shows the correct formatting with dots for german [de] locale.

image

PR #1758, opened 2024-11-25
Userlocal jsonparser

dbauszus-glx:
This PR replaces the userIDB plugin with a more general userLocal plugin.

The userLocale plugin provides a test interface to save/overwrite a locale, or to remove/delete a locale. Currently the locale can only be stored in a userIndexedDatabase through the userIndexedDB utility module. Removal will destroy the database though it should be possible to set this to null.

The _default view script will check whether the userLocal is configured on a locale an attempt to get the locale from the userLocal utility module.

The userLocale utility module requires the userIDB module to get, save, and remove the store user locale.

The jsonParser module is required to parse the mapview.locale and assign the mapview layers to this locale.

The jsonParser module strips properties which cannot be stored with the JSON locale in an object store.


cityremade:
as of now when I just use the following
```
"userLocale": {
                "store": "userIDB"
            }
```

I see a flood of errors of missing plugins while I don't have any plugins in my test workspace.

simon-leech (dismissed):
Passing in an array of strings as part of the mapview, e.g. 
``` js
"plugins":[
"test.js",
"test_1.js" 
]
```

Crashes in the `excludeProperty` function in the `typeof value === 'object'` if statement.

dbauszus-glx:
@simon-leech I am unable to replicate this. 

image

dbauszus-glx:
To prevent circular reference issues the jsonParser will assign a `__parsed` flag to each object which is being parsed to prevent the object from being parssed again. The flags will be removed before the object iss returned.

dbauszus-glx:
The listview assigns the group object as group property value. This messes with rebuilding the mapview / layer listview from the locale.

The string property for the group name would be turned into an object in the userLocale. This object would not be recognised as a group name [string] trying to populate the layer listview.

lib/utils/userIndexedDB.mjs
```diff
@@ -1,96 +1,18 @@
...
- * @param {Object} store 
- * @returns {Promise} OpenDBPromise
- */
+@function openDB
+@param {Object} store 
+@returns {Promise} OpenDBPromise
+*/
 export async function openDB(_store) {
 
   const store = _store.toString()
```
RobAndrewHurst, line 18: Do you think that the store name should be passed into the `indexedDB.open` and not default to the mapp.user.email + mapp.user.title straight away if there is no db with the store name? Ie we can make this more configurable on the plugin?

dbauszus-glx: Yes. I think that's a good idea.

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/1758/commits/8ae5e196d6b8d550ad183f7411f502f218899d61

@RobAndrewHurst The userIndexedDB utils should receive a params object with the db and user properties defined.

The user will be anonymous if the mapp.user object is false.

The db will be taken from the instance [workspace.title]. The workspace.title will be assigned from the process.env.TITLE which has a default in the api module.

dbauszus-glx:
Self referenced theme objects will be lost. The styleParser must assign a key to the theme object properties of the theme.

The key should be assigned as string property for __parsed object properties. The theme object will then be correctly assigned when the locale is loaded from them indexeddb store.

The iteration of themes in the styleparser must be resolved first before this can be merged.

https://github.com/GEOLYTIX/xyz/pull/1779/commits/9101926a01888b37b5c71fd260931ea46bf41322

dbauszus-glx:
The exludeProperty check must happen before the assignValue.

Instead of checking for a number to be not not a number the check should be on type of object at the end of the assignValue method.

The maplibregl object must not be assigned to the layer object.

simon-leech (approved):
I'm happy with this now 👍  
Brilliant work @dbauszus-glx !

Issue #1732, opened 2024-11-19
saml2 optional dependency; export and warning.


dbauszus-glx:
The saml2 library should be an optional dependency.

The module should export as null if the package is not available or the process.env is not configured for saml2.

There should be no warning if saml2 exports as null.

dbauszus-glx:
This should also resolve #463 which has been reopened.

PR #1760, opened 2024-11-25
saml2 optional dependency. 

RobAndrewHurst:
# saml2 optional dependency (Export & warning)

The `saml.js` module will no longer show a warning if the package `saml2-js` package is not installed.
It will only do so by assessing if ther are any `SAML_*` process.env keys present and if the pacakge is not installed.

```js
  //Check if there are any SAML keys in the process.
  const samlKeys = Object.keys(process.env).filter(key => key.startsWith('SAML'));

  //If we have keys then log we that the module is not present
  if (samlKeys.length > 0) {
    console.log('SAML2 module is not available.')
  }
```

It will also export the module as null if the package is not installed.

## GitHub Issue

[#1732](https://github.com/GEOLYTIX/xyz/issues/1732)

### DIFF
--- a/mod/user/saml.js
+++ b/mod/user/saml.js
@@ -62,8 +62,20 @@ try {
     sign_get_request: true,
   });
 
+  module.exports = saml;
+
 } catch {
-  console.log('SAML2 module is not available.')
+
+  //Check if there are any SAML keys in the process.
+  const samlKeys = Object.keys(process.env).filter(key => key.startsWith('SAML'));
+
+  //If we have keys then log we that the module is not present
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Updated a spelling mistake in the documentation. 

Tested without SAML keys, no warning.
Tested with SAML keys, warning seen.

dbauszus-glx:
@RobAndrewHurst logger jwt require can be after saml require. No need to check and bundle if not required. https://github.com/GEOLYTIX/xyz/pull/1760/commits/26e517e7aa59189790346ec70810f19525515069

PR #1762, opened 2024-11-26
Tabview BtnRow must be set to display block in the `activate`

simon-leech:
The activate event method which is triggered when a tabview tab is added/shown must check for and display a toolbar associated to a dataview in the tab which is being activated.


dbauszus-glx:
The activateTab method should be a documented function in its own right bound to the entry object.

dbauszus-glx (approved):

PR #1763, opened 2024-11-26
`globalThis` replacing `self`.

RobAndrewHurst:
# `globalThis` replacing `self`

I have replaced the self. assignment for the more moden [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)

The main benefit of this is that we will be able to see the exact references and have better intelliSense for the `mapp` & `ui` objects.

Bellow are some examples of what you should see in vscode.

image

image

image

image

### DIFF
--- a/lib/mapp.mjs
+++ b/lib/mapp.mjs
@@ -54,7 +54,7 @@ if (window.ol === undefined) {
   }
 }
 
-self.mapp = {
+globalThis.mapp = {
   ol: _ol,
 
   version: '4.13.0-alpha',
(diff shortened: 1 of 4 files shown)


simon-leech:
@RobAndrewHurst  In the `_test.html` and `_mapp.test.mjs` we assign 
`self.codi` and `self.mappTest`- should these references to `self` also be changed?

cityremade (approved):

RobAndrewHurst:
> @RobAndrewHurst In the `_test.html` and `_mapp.test.mjs` we assign `self.codi` and `self.mappTest`- should these references to `self` also be changed?

Yeah! I will do that now!

PR #1766, opened 2024-11-27
Fix geojson json build object string

dbauszus-glx:
The json_build_object syntax is wrong in the geojson query template.


RobAndrewHurst (approved):
Updated tests and happy with this!

AlexanderGeere (approved):

PR #1774, opened 2024-12-16
Adjust min width of 2nd div in location view grid column

dbauszus-glx:
This fixes the issue where an inline dropdown button element messes with the grid layout.

image


RobAndrewHurst (approved):

simon-leech (approved):
All working for me :)

PR #1779, opened 2024-12-18
Graduated theme parser

dbauszus-glx:
Reversing the order for graduated theme categories was a terrible idea. This causes the order to be reversed multiple times if a theme is parsed multiple times.

The graduated_breaks check should not warn and always be correct. "less_than" order is assumed.

The categories are ordered according to their numeric values and direction, eg. less_than / greater_than.

Themes can be ammended and parsed multiple times with the order always being correct.

The styleParser method has multiple nested methods which have been moved to the module root and documented.

The styleObject method has been renamed to catStyle to indicate what the method actually does.

### DIFF
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -134,241 +134,81 @@ export default function styleParser(layer) {
   if (layer.style?.labels) {
     layer.style.label = typeof layer.style.label === 'object' ? layer.style.label : layer.style.labels[layer.style.label || Object.keys(layer.style.labels)[0]];
   }
+}
 
-  function warnings(layer) {
+/**
+@function warnings
 
-    if (!layer.style.default) {
+@description
+The warnings method parses the layer style configuration and warns on legacy configurations while trying to rectify these issues.
 
-      console.warn(`Layer: ${layer.key} has no implicit default style. Please add style.default.`)
+@param {layer} layer A json layer object.
+@property {Object} [layer.cluster] Cluster configuration for a point layer.
+@property {Object} [layer.hover] Legacy configuration for style.hover.
+@property {Object} [layer.icon_scaling] Legacy configuration for style.icon_scaling.
+@property {layer-style} layer.style The mapp-layer style configuration.
+@property {feature-style} style.default Default feature style.
+@property {integer} [style.zIndex] Legacy configuration for layer.zIndex.
+*/
+function warnings(layer) {
 
-      // Assign default style for vector layer.
-      // Non cluster layer do not have a default icon style
-      layer.style.default = layer.cluster
-        ? {
-          icon: {
-            type: 'dot'
-          }
-        } : {
-          strokeColor: '#333',
-          fillColor: '#fff9',
+  if (!layer.style.default) {
+
+    console.warn(`Layer: ${layer.key} has no implicit default style. Please add style.default.`)
+
+    // Assign default style for vector layer.
+    // Non cluster layer do not have a default icon style
+    layer.style.default = layer.cluster
+      ? {
+        icon: {
+          type: 'dot'
         }
-    }
-
-    if (layer.style.default.style) {
-
-      console.warn(`Layer: ${layer.key} has a style object within the default style configuration.`)
-
-      layer.style.default = layer.style.default.style
-
-      delete layer.style.default.style
-    }
-
-    iconObject(layer.style.default)
-
-    // Handle deprecated layer.hover configuration.
-    if (layer.hover) {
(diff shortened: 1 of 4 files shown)


dbauszus-glx:
The numbered module patch documentation has been split into the method documentation blocks where and then the checks actually happen.

The checks theme[s], hover[s], label[s] have been standardized. Assign the the first key as string property if undefined, and assign the lookup object if property is type string.

There should be only a single check for themes and only a single iteration through the themes object keys. This was happening twice in direct succession.

lib/layer/styleParser.mjs
```diff
@@ -98,31 +84,56 @@ export default function styleParser(layer) {
...
-  if (layer.style?.theme) {
-    parseTheme(layer.style.theme, layer)
-  }
+  parseTheme(layer.style.theme, layer)
 
   if (layer.style?.themes) {
+
     Object.keys(layer.style.themes).forEach(key => {
+
+      if (layer.style.themes[key].skip) {
```
RobAndrewHurst, line 93: It appears that the layer in the scope is not defined.

lib/layer/styleParser.mjs
```diff
@@ -132,312 +143,393 @@ export default function styleParser(layer) {
...
-    ...layer.style.cluster
-  }
+  iconObject(cat.style)
 
-  if (layer.style.cluster.zoomInScale) {
+  // Assign default icon if no cat style icon could be created.
+  if (!cat.style.icon && defaultStyle.icon) {
 
-    layer.style.zoomInScale = layer.style.cluster.zoomInScale
+    console.warn(`Layer:${layer.key}, Failed to evaluate icon: ${JSON.stringify(cat)}. Default icon will be assigned.`)
```
RobAndrewHurst, line 508: here is the issue for the layer not being defined.

dbauszus-glx:
@RobAndrewHurst the issue should be fixed in this commit.

https://github.com/GEOLYTIX/xyz/pull/1779/commits/0563ea83de0cc5437b80e9df0a4110f201293e68

The layer should be provided as argument to the caStyle method for the warning. The defaultStyle clone can then be created inside the method when needed.

lib/layer/styleParser.mjs
```diff
@@ -132,312 +143,395 @@ export default function styleParser(layer) {
...
-      }
-    }
+  iconObject(layer.style.default)
 
-    if (typeof theme.cat === 'object') {
+  // Handle legacy layer.hover configuration.
+  if (layer.hover) {
+    console.warn(`Layer: ${layer.key}, layer.hover{} should be defined within layer.style{}.`);
+    layer.style.hover = layer.hover;
+    delete layer.hover;
```
RobAndrewHurst, line 229: It appears that the legacy property deletions are happening in the different functions but are not changing the actual layer.

RobAndrewHurst: In the local scope of the function I can see the deletion happening, but after we kick out of the function the property is persists.

lib/layer/styleParser.mjs
```diff
@@ -132,312 +143,395 @@ export default function styleParser(layer) {
...
+  // Handle legacy layer.icon_scaling configuration.
+  if (layer.icon_scaling) {
 
-        cat.label ??= key
-        cat.value ??= key
+    console.warn(`Layer: ${layer.key}, layer.icon_scaling has been assigned to layer.style.icon_scaling`)
 
-        return cat
-      })
+    layer.style.icon_scaling ??= layer.icon_scaling
```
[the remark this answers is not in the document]
dbauszus-glx: It doesn't really matter if deleted or not as this would not affect the upstream workspace.

RobAndrewHurst (approved):
All happy with this! 

Added in tests to check for the styleParser test cases, and also added in a test to check for the theme order of graduated themes.

let me know what you think of the suggestion I added in the comments above.

lib/layer/styleParser.mjs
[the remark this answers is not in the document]
RobAndrewHurst: Ok cool!

RobAndrewHurst:
@dbauszus-glx I reduced the styleParsers complexity by moving the hover/s logic into it's own method called `handleHovers`.

this has been documented as well.

Lemme know what you think.

dbauszus-glx:
@RobAndrewHurst I am ok with this as long as it works. Having the method in the hover object feels wrong. The hover object itself should just be a bit of configuration like the label style, but this should be addressed in a different issue.

simon-leech (approved):
I'm happy with this - working as expected on the test cases i've used

PR #1780, opened 2024-12-19
syncFields array and documentation

dbauszus-glx:
The dependents array defined as a string for geometry entries would throw an error since the syncFields method expects a fields array as params argument.

The method has now been documented and the fields param is turned into an array if not provided as an array.


simon-leech (approved):
All good with me :)

RobAndrewHurst (approved):

PR #1783, opened 2024-12-20
Reduce mapview complexity

dbauszus-glx:
The mapview decorator method is currently the most complex in the mapp library.

Items relating to the openlayers view should be handled in the viewConfig method which creates the mapview.view object which can be used as options argument for the ol.View constructor.


simon-leech (approved):
This is working as expected for me on existing instances. 
Tested using both MAPP 'standard' view and also in MAPP 'custom' views with mapviews attached and it is working as previously.

cityremade (approved):
I've tested on a custom view with multiple maps

RobAndrewHurst (changes requested):
image

The decorate function has a complexity of 18.

simon-leech (approved):
Tested again and working as expected on both default MAPP and also in a custom view with multiple mapviews associated

RobAndrewHurst (approved):
LGTM!

PR #1785, opened 2025-01-02
Bump OL to 10.3.1 to include clear cache fix.

dbauszus-glx:
This patch PR upgrades OL to 10.3.1 which includes a fix to clear the LRU cache properly. 

https://github.com/openlayers/openlayers/pull/16328

The workaround setting the OL source property manually has been removed. OL properties should never be directly modified but only through the OL api.

The associated test has been removed since it is meaningless. The test set a property to foo before calling a method which sets the property to an empty object and then checking whether the property is set to an empty object. The test did not actually look at the tile cache and associated trigger.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -84,8 +84,6 @@ export default function MVT(layer) {
     layer.source.clear()
     layer.source.refresh()
 
-    // Reset source tiles to force refresh.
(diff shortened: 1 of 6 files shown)


simon-leech (approved):
This is all as expected for me.

cityremade (approved):
Looks good to me.

Issue #1784, opened 2025-01-02
Entry Dataviews on Location Update


simon-leech:
## Description
When calling the `location.update()` method, this subsequently calls the `renderLocationView` method which rebuilds the `location.infoj` in the `location` panel. 
However, `dataviews` are not being refreshed properly. 
 
In `entries/dataview.mjs`: 
1. We remove the `entry.display` flag when the data is null - however if this data is subsequently provided, say via a db update on the back-end, the display flag is gone. This should be stored as a temporary variable so it can be reset if required. 
2. An `entry` may have a `chkbox` associated with it for display of the `dataview`. This needs to be enabled when the `disabled` flag is removed if data is present. 
3. An `entry` may have a `chkbox` associated with it for display of the `dataview`. This needs to be set to checked if the `display` flag is set to true. 
4. An `entry` may have a `tab` associated with it (displaying in the `tabview`). This needs to be enabled when the `disabled` flag is removed if the data is present.
5. `entry.locationViewTarget` is set to `location ${entry.class}`. Most entries don't have a class, so this should use `entry.key` or `entry.query` to avoid creating `location undefined`. 

## Type of Issue 
Please delete options that are not relevant, and select all options that apply. 

- ✅ Bug

PR #1786, opened 2025-01-02
Entry Dataviews Refresh

simon-leech:
## Description
When calling the `location.update()` method, this subsequently calls the `renderLocationView` method which rebuilds the `location.infoj` in the `location` panel. 
However, `dataviews` are not being refreshed properly. 
 
In `entries/dataview.mjs`: 
1. We remove the `entry.display` flag when the data is null - however if this data is subsequently provided, say via a db update on the back-end, the display flag is gone. This should be stored as a temporary variable so it can be reset if required. 
2. An `entry` may have a `chkbox` associated with it for display of the `dataview`. This needs to be enabled when the `disabled` flag is removed if data is present. 
3. An `entry` may have a `chkbox` associated with it for display of the `dataview`. This needs to be set to checked if the `display` flag is set to true. 
4. An `entry` may have a `tab` associated with it (displaying in the `tabview`). This needs to be enabled when the `disabled` flag is removed if the data is present.
5. `entry.locationViewTarget` is set to `location ${entry.class}`. Most entries don't have a class, so this should use `entry.key` or `entry.query` to avoid creating `location undefined`. 

**The checkbox creation method has been moved into the `entry` or `layer` methods - this is to ensure that when a dynamic dataview is provided, the whole UI and dataview are rebuilt.** 

## GitHub Issue
#1784 

## Type of Change

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/panels/dataviews.mjs
+++ b/lib/ui/layers/panels/dataviews.mjs
@@ -109,6 +109,22 @@ export default function dataviews(layer) {
       dataview.remove()
     }
 
+    // Create checkbox control for dataview.
+    dataview.chkbox = mapp.ui.elements.chkbox({
+      data_id: dataview.key,
+      label: dataview.label,
+      checked: !!dataview.display,
+      disabled: dataview.disabled,
+      onchange: (checked) => {
+
+        dataview.display = checked
+
+        dataview.display
+          ? dataview.show()
+          : dataview.hide()
+      }
+    })
+
     if (mapp.ui.Dataview(dataview) instanceof Error) return;
 
     // Display dataview if layer and dv have display flag.
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -106,22 +106,6 @@ export default async function Dataview(_this) {
 
   _this.hide ??= hide
 
-  // Create checkbox if a label is provided.
-  _this.chkbox = _this.label && mapp.ui.elements.chkbox({
-    data_id: _this.key,
-    label: _this.label,
-    checked: !!_this.display,
-    disabled: _this.disabled,
-    onchange: (checked) => {
-
-      _this.display = checked
-
-      _this.display
-        ? _this.show()
-        : _this.hide()
-    }
-  })
-
   // Create dataview toolbar
   dataviewToolbar(_this)
  
--- a/lib/ui/Tabview.mjs
+++ b/lib/ui/Tabview.mjs
@@ -73,8 +73,8 @@ The decorator method will add `show()` and `hide()` methods for the tab.
 */
 function addTab(entry) {
 
-  // The entry already has a tab.
-  if (entry.tab) return;
+  // The entry already has a tab, and is not flagged as dynamic.
+  if (entry.tab && !entry.dynamic) return;
 
   const tabview = this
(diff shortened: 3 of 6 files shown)


dbauszus-glx (changes requested):
This should work with a dependent dataview receiving the data from the field.

```json
{
  "title": "test",
  "type": "json",
  "field": "json_field",
  "dependents": "dependent_dataview",
  "edit": true
},
{
  "type": "dataview",
  "dataview": "Json",
  "target": "tabview",
  "label": "json_field",
  "field": "dependent_dataview",
  "fieldfx": "json_field",
  "activateUpdate": true
}
```

However this does not work with the updateDataview method requiring a query while the data may have been provided to the entry otherwise.

dbauszus-glx:
The jsonDataview should set the data on creation. https://github.com/GEOLYTIX/xyz/pull/1786/commits/18155c11f22e83945a49bf5115fffd4337b053f9

simon-leech:
@dbauszus-glx  This should all be resolved now, it is working with the json method in either the `location` or `tabview`, with or without a `checkbox` and also working with a query `dataview`, in either the `location`, `tabview` and with or without a `checkbox`.

lib/ui/layers/panels/dataviews.mjs
```diff
@@ -109,6 +109,22 @@ export default function dataviews(layer) {
       dataview.remove()
     }
 
+    // Create checkbox control for dataview.
+    dataview.chkbox = mapp.ui.elements.chkbox({
```
RobAndrewHurst, line 113: So this is going to create a checkbox even if there isn't a label?

RobAndrewHurst: Only reason I ask is that it seems like duplicate code pretty much apart from the label check.

RobAndrewHurst: image

I see. It gets assigned if not provided further above.

Different because an entry might not have a label.

still feels weird with the duplicate code.

RobAndrewHurst (approved):
Happy with this for now. 

The duplicate code feels odd though.

In the future could we get a break down of all the changes and the lines they permit to instead of exactly what the issue says?

dbauszus-glx:
@RobAndrewHurst 

The reason for the duplicate code is that the chkbox element is specific to the dataview location entry or the dataview layer panel.

The location view entry may not have a checkbox. The layer panel will always have a checkbox.

The call to the checkbox element is one line of code. I much rather have this line of code in the modules where the element is needed than in a shared module which will require more complex conditions for the function calling module in regards to the element.

The checkbox itself is not required by the dataview and it should be possible to call the dataview module to create a dataview without the checkbox being created on the dataview object.

dbauszus-glx:
@simon-leech I don't understand this bit.

```js
  // Dataview is dependent on other field entries.
  if (entry.dependents) {
    console.warn(
      `The dataview type entry key:${entry.key} may be dependent on other entries but has no dependents.`
    );
  }
```

dbauszus-glx (approved):

AlexanderGeere (approved):

Issue #1764, opened 2024-11-27
Password check on login screen


dbauszus-glx:
The green/red shift for invalid password on the login screen may be impossible for user with a colour deficiency.

[Screencast from 2024-11-27 09-41-01.webm](https://github.com/user-attachments/assets/929b73ee-662a-4e0b-9783-2894126f909d)

PR #1790, opened 2025-01-03
12 Character Password update

simon-leech:
Updated the default login and register views to use a minimum length of 12 characters.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -84,8 +84,6 @@ export default function MVT(layer) {
     layer.source.clear()
     layer.source.refresh()
 
-    // Reset source tiles to force refresh.
-    layer.source.sourceTiles_ = {};
     layer.featureSource.refresh()
 
     if (callback instanceof Function) callback(layer)
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -1,38 +1,7 @@
 /**
-### mapp.layer.styleParser()
+### /layer/styleParser
 
-The `styleParser` module method is responsible for parsing and validating the layer.style object for assigning feature styles.
-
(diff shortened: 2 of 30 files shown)


cityremade (approved):
I see 10s changed to 12s 👍

dbauszus-glx (changes requested):
This does not work at the moment.

The information in regards to the length is shown as a tooltip which I won't see on a touchscreen.

The red light / green light indicator is invisible to people with a colour deficiancy. https://github.com/GEOLYTIX/xyz/issues/1764

The info text must emphasise that the password must be reset if the current password does not match the complexity schema.

image

We need to create a login template which uses a dictionary endpoint in the future. Having to change multiple language view templates for login and registration is unsustainable. I will create a ticket for this.

simon-leech:
Updated the login page: 
1. Added message about the password change
2. Added error message when email or password are invalid to prevent issues with accessibility. 
image


Please could this be reviewed before I do the same change on all the non-english login pages too?

RobAndrewHurst:
What happens to the message in a months time? Do we have to create another PR to remove it?

simon-leech:
> What happens to the message in a months time? Do we have to create another PR to remove it?

I would guess so yeah - I'm not sure how else to handle this really 😅

RobAndrewHurst:
image

The Choose language label is quite far away from the drop down. I might want to be closer to the drop down itself.

simon-leech:
@RobAndrewHurst Good point - I've moved it next to it 👍 
image

RobAndrewHurst (approved):
LGTM!

dbauszus-glx (approved):
I am happy for this to be merged. There are a couple of tweaks but I want to look at these after we are able to load the dictionary into a single login/registration view template.

PR #1793, opened 2025-01-03
Allow for layer entry without query [featureSet / featureLookup] configuration

dbauszus-glx:
The layer entry method as the sucessor to the mvt_clone method should not be limited to featureSet or featureLookup entries.

The featureSet and featureLookup require data to be provided as an array from either a field, inside the entry configuration, or from a [parameterised] query.

A layer entry without the featureSet or featureLookup flag should show the all features as defined by the layer filter.

```json
{
  "type": "layer",
  "name": "OA Layer",
  "layer": "oa"
},
{
  "type": "layer",
  "name": "featureSet Layer",
  "layer": "oa",
  "featureSet": true,
  "field": "oas"
},
{
  "type": "layer",
  "name": "featureLookup Query Layer",
  "layer": "oa",
  "featureLookup": true,
  "query": "oa_featurelookup",
  "queryparams": {
    "id": true
  },
  "template": {
    "key": "oa_featurelookup",
    "template": "select oas[1] as id from public.points where id = ${id}"
  }
}
```

[Screencast from 2025-01-03 12-40-54.webm](https://github.com/user-attachments/assets/7ff671e8-c658-425d-b20f-973bfa726e55)


cityremade:
I need this documented better. When I recreate this setup with retail places I see errors and errors.
What is "field" in the second setup?

dbauszus-glx:
@cityremade  I updated the layer entries method documentation.

http://localhost:3000/xyz/mapp/module-_ui_locations_entries_layer.html

image

In the example the field is an array field which has an array of feature ids for only those features to be rendered.

image

image

cityremade:
Regarding use cases,
should layer entry still appear if no response is returned from the related query?
Example is no intersecting non-seamless zone like retail place or pitch.

dbauszus-glx:
> Regarding use cases, should layer entry still appear if no response is returned from the related query? Example is no intersecting non-seamless zone like retail place or pitch.

No; The layer toggle should be disabled. I will make a change to method to disable the toggle.

cityremade (approved):
thanks, I am happy with this.

AlexanderGeere:
@dbauszus-glx The toggle seems to remain enabled if featureLookUp is true and a query is supplied but the query returns no data, 
Screenshot 2025-01-09 114506
should line 111 maybe be `||` instead of `&&`. 

If you check id=1 on `bugs_testing/infoj/layer_entry.json` on the points layer you will see the toggle remains active even though the query doesn't return any data.

dbauszus-glx:
@AlexanderGeere This should now be fixed in the last commit. https://github.com/GEOLYTIX/xyz/pull/1793/commits/159d2c9a42fc847679372b096a7a2aeb425bf065

simon-leech (changes requested):
I see a few issues with this I think: 

1. A layer that is a direct clone does not have the dataviews that are on the original layer. 
IE if the original layer was OA and had an OA tableview, its not show in the entry version of this layer. Should it be? 
2. When selecting a location from the cloned whole layer, queries in the infoj fail as they aren't passing the right `layer` param. 
http://localhost:3000/x/api/query?id=1020001287&template=query&locale=UK&layer=oa%7Csites!796
You see here it adds the layer param as `oa|sites!796` which then fails the layer lookup on the query.

dbauszus-glx:
@simon-leech Dataviews are not supported, editing is not yet supported. We will look at editing in the future but I think dataviews in layer in location from another layer is too much nesting.

I will look at the selection issue.

dbauszus-glx:
@simon-leech 
The issue with selection and highlight should be resolved in the last commit. The Key with capital K will be assigned as key in the location get.

The key for multiple layers from the same mapview layer will be assigned from the OL UID which is guaranteed to be unique in the layer entry method.

simon-leech (approved):
All working for me - nice one :)

PR #1796, opened 2025-01-06
processEnv - initial module;

cityremade:
Introducing a process env module which passes environment variables and assigns defaults in order to avoid referencing process.env across the codebase.


dbauszus-glx:
@cityremade 

Can you move the mapp_env module to the mod/utils directory and rename this to processEnv since this isn't really related to the mapp library but deals with the process environment for the xyz node modules.

Can you also move the globale env typedef from the api module into the processEnv utility module.

I had a look at the sonarcloud issues. 

image

This can be fixed by assigning true or false boolean in the processEnv module.

dbauszus-glx:
@cityremade 

image

You can fix the optional chaining warning like so:

image

RobAndrewHurst:
I have some ideas for us avoiding importing the env utility module everywhere. I will do this tomorrow

Issue #1778, opened 2024-12-18
Test Plugin 🔌


RobAndrewHurst:
# Test Plugin 🔌

We need the ability to run these tests core/custom with a plugin.

## Requirements

- The plugin will need to have the ability to be configured.
  - Ability to run core.
  - Show/hide summary.
  - quite/loud.
  - add in custom test functions.
- Trigger
  - With button.
  - Via means of web hooks.
  - Other alternatives welcome. (cron job would be awesome too).
- Run a specific test or suite on a layer.
  - Also need to have similar kind of configurations as the locale plugin.

simon-leech:
**Integrity Test**
* The integrity test should be accessible via the URL parameter `integrity=true` 
* The integrity test forms the basis for layer/location testing on an application and can be ran either locally or on a deployed application. 
**Requirements**
The integrity test should loop through every locale available to that user and run the tests.
The tests should attempt to do these things: 
1. Turn on every layer accessible to the user
2. Select a location on every layer accessible to the user (turning on all dataviews too)
3. Add a new location on every editable layer accessible to the user 
4. Delete that new location on every editable layer accessible to the user 
5. Turn on all filters on every layer accessible to the user
6. Turn on all themes on every layer accessible to the user 
7. Run the Template Test on the Workspace API to check for failing or missing templates 

Each test should return a pass or fail that is stored to the layer describe. 
This includes errors for failing themes and failing filters. 


A nice to have too would be the ability to test the gazetteer by just providing a 'test' word and checking for any errors.

dbauszus-glx:
This has been resolved in linked PR.

PR #1801, opened 2025-01-07
Provide user roles param for admin requests.

dbauszus-glx:
This PR introduces the roles request parameter for admin user to test roles on any request, eg. checking layer access in a locale.

http://localhost:3000/api/workspace/locale?locale=test&layers=true&roles=foo,bar


simon-leech (approved):
This is exactly what we need - working as expected!

In v4.13.0 as part of this ongoing issue (https://github.com/GEOLYTIX/xyz/issues/1778#issuecomment-2557011198) we can implement this into the `integrity tests` to enable us to return a list of layers / functionality by role! 🪄

RobAndrewHurst (approved):
LGTM!

PR #1806, opened 2025-01-13
Fix SAML Logout Redirect Issue 🩹

RobAndrewHurst:
# Fix SAML Logout Redirect Issue

## Issue

The SAML logout redirect was behaving differently between local Express and Vercel deployments due to how `res.redirect()` was being handled.

## Changes

- Replaced `res.redirect()` with explicit header setting for consistent behavior
- Updated logout callback to use explicit status codes and headers

### Before

```javascript
function logoutCallback(res) {
  try {
    res.setHeader(
      "Set-Cookie",
      `${process.env.TITLE}=; HttpOnly; Path=${process.env.DIR || "/"}; Expires=Thu, 01 Jan 1970 00:00:00 GMT`,
    );
    res.redirect(`${process.env.DIR || "/"}`);
  } catch (error) {
    console.error("Logout validation failed:", error);
    return res.redirect("/");
  }
}
```

### After

```javascript
function logoutCallback(res) {
  try {
    res.setHeader(
      "Set-Cookie",
      `${process.env.TITLE}=; HttpOnly; Path=${process.env.DIR || "/"}; Expires=Thu, 01 Jan 1970 00:00:00 GMT`,
    );
    res.setHeader("location", `${process.env.DIR || "/"}`);
    return res.status(302).send();
  } catch (error) {
    console.error("Logout validation failed:", error);
    res.setHeader("location", "/");
    return res.status(302).send();
  }
}
```

## Testing

- Verified logout behavior on local Express server
- Verified logout behavior on Vercel deployment
- Confirmed consistent redirect behavior across platforms

## Why

Using explicit header setting and status codes provides more consistent behavior across different platforms compared to using the Express `redirect()` helper method.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -110,7 +110,7 @@ const routes = {
 
 process.env.COOKIE_TTL ??= '36000';
 
-process.env.TITLE ??= 'XYZ | MAPP'
+process.env.TITLE ??= 'XYZ | MAPP';
 
 process.env.DIR ??= '';
 
--- a/mod/user/saml.js
+++ b/mod/user/saml.js
@@ -189,7 +189,6 @@ Authentication Flow:
 @param {req} req.params - Route parameters
 
 @param {Object} res - HTTP response object
-@param {res} res.redirect - Redirect function
 @param {res} res.send - Send response function
 @param {res} res.setHeader - Set response header
 
@@ -236,7 +235,6 @@ function saml(req, res) {
 @description Handles the metadata response 
 
 @param {Object} res - HTTP response object
-@param {res} res.redirect - Redirect function
 @param {res} res.send - Send response function
 @param {res} res.setHeader - Set response header
 **/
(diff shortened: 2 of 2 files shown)


dbauszus-glx (changes requested):
I notice that there is some confusion between param and property in the SAML module.

RobAndrewHurst:
@dbauszus-glx I have updated the docs. Always make that silly mistake of forgetting properties of params.

dbauszus-glx (changes requested):
You still need to update the types.

for the res.redirect property the type would be function, not res.

RobAndrewHurst:
> You still need to update the types.
> 
> for the res.redirect property the type would be function, not res.

I am confused by this as the res.redirect function has been completely removed.. It's not in the code or in the docs

dbauszus-glx:
> > You still need to update the types.
> > for the res.redirect property the type would be function, not res.
> 
> I am confused by this as the res.redirect function has been completely removed.. It's not in the code or in the docs

I didn't pull the last commit. All good.

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #1809, opened 2025-01-15
css structure (sync with minor)

cityremade:
New css directory:
```
css/
  base/
  elements/
  layout/
 _mapp.css 
 _ui.css
```
Ideally _mapp.css would be imports only.
At the moment is still has a few clauses and I am not sure where to move them

Inside _mapp.css:

Should this one be global?
`
a > img {
  height: 100%;
}`

Which legend container is it? If it's used in the layer legend then it shouldn't sit in _mapp.css
`.legend`


dbauszus-glx (changes requested):
There seem to be some changes to the ui.css when I build the css after pulling the changes.

dbauszus-glx (approved):
This is looking good now.

dbauszus-glx (changes requested):
I would perhaps remove tabulator css as this should really be added in the plugin.

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #1811, opened 2025-01-15
The btnRow should be set to 'flex' not 'block'

simon-leech:
We were incorrectly setting the `btnRow` to be `display:block` when it should be `display:flex`. 
This was messing with the styling of the `btnRow`, and was overwriting the `ui.css` stylesheet incorrectly. 

image

This PR simply changes the references from `flex` to `block` to resolve it.

### DIFF
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -154,7 +154,7 @@ function show() {
   }
 
   //Show toolbar buttons if there are any
-  this.btnRow?.style.setProperty('display','block')
+  this.btnRow?.style.setProperty('display','flex')
 
   this.target.style.display = 'block'
 }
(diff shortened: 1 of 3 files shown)


cityremade:
I guess it needs to be included here as well https://github.com/GEOLYTIX/xyz/pull/1809

simon-leech:
> I guess it needs to be included here as well #1809

This PR is for patch and it's not touching any CSS, only fixing the assignment of 'block' to 'flex' in the js files. So we will be good once this is approved and merged to just merge patch into 1809 👍

dbauszus-glx (commented):
Is that required in a tabview?

simon-leech:
> Is that required in a tabview?

I'm sorry @dbauszus-glx , I don't understand what you're asking here? I've just replaced setting to block to setting to flex where it was already currently being applied.

dbauszus-glx:
@simon-leech It is required. Ignore previous message. I just did a test. 
image

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #1813, opened 2025-01-15
Filter Visual Changes

simon-leech:
1. Added new filter icon
2. zoomToExtent renamed zoomToFilteredExtent
3. zoomToFilteredExtent only shown when filter current has a length
4. The zoomButton for zoom restricted layers has been moved to the left of the display toggle too, to ensure things stay in the same place across layers. 
5. The zoomButton has a new icon that includes a plus in the centre to distinguish between our search icon, and a zoom in icon - this was causing confusing for users. 

Filter Icon will show when the layer has any filters applied to it. To indicate this to the user.
User can click the button to zoom to the extent of the filtered dataset.
image

The zoomtoFilteredExtent button can be omitted by passing a `layer.viewConfig` object without it in. 
e.g. 
``` json
 "viewConfig": {
    "displayToggle": true,
    "zoomBtn": true
  },
```


simon-leech:
Please note that this is currently a patch PR as it was done this way so I could deploy a test instance for user feedback. 
So please don't merge this PR into patch. 

This PR will be re-created on the minor branch once #1809 has been merged

PR #1815, opened 2025-01-16
processEnv utility module (sync with minor)

cityremade:
Introducing a process env module which passes environment variables and assigns defaults in order to avoid referencing process.env across the codebase.


express.js
```diff
@@ -4,85 +4,88 @@ const express = require('express');
 
 const cookieParser = require('cookie-parser');
 
+const rateLimit = require('express-rate-limit');
```
RobAndrewHurst, line 7: I added in a rate limiter into the express server.

express.js
```diff
@@ -4,85 +4,88 @@ const express = require('express');
...
+
 const app = express();
 
+const limiter = rateLimit({
```
RobAndrewHurst, line 11: This is the configuration used for it. We can discuss what kind of  limits we want to implement. currenlty its limited to 100 requests per window every 15mins. It might not be enough. But we can discuss.

express.js
```diff
@@ -4,85 +4,88 @@ const express = require('express');
...
 
-app.use(`${process.env.DIR || ''}/tests`, express.static('tests'));
+app.use(env.DIR, express.static('public'));
 
-app.use(process.env.DIR || '', express.static('tests'));
+app.use(`${env.DIR}/tests`, express.static('tests'));
+
+app.use(env.DIR, express.static('tests'));
 
 app.use(cookieParser());
```
RobAndrewHurst, line 37: What is this cookieParser doing?

express.js
dbauszus-glx: I think this wants to be a seperate PR.

express.js
dbauszus-glx: I am not sure. It might be possible to remove this. Can you create a separate issue for this?

RobAndrewHurst: Yeah! I have dismissed the alert for now as the cookie gets handled in a different way anyways.

RobAndrewHurst: The rate limiter solved all the other security issues though :)

mod/utils/processEnv.js
```diff
@@ -0,0 +1,75 @@
...
+Object.entries(process.env).forEach(([key, value]) => {
+  if (!(key in xyzEnv)) {
+    xyzEnv[key] = value;
+  }
+});
+
+// Freeze to prevent modifications
+Object.freeze(xyzEnv);
+
+globalThis.xyzEnv = xyzEnv;
```
RobAndrewHurst, line 75: I have renamed the env object to xyzEnv for better context of reference aswell as avoiding a reserved word. The xyzEnv object is also assigned to the globalThis object and added in on the express.js file  so that we don't need to import it in every module.

mod/utils/processEnv.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Add remaining env vars
+Object.entries(process.env).forEach(([key, value]) => {
+  if (!(key in xyzEnv)) {
+    xyzEnv[key] = value;
+  }
+});
+
+// Freeze to prevent modifications
+Object.freeze(xyzEnv);
```
RobAndrewHurst, line 73: I have also frozen the object so that nothing can tamper with the object itself.

mod/utils/processEnv.js
```diff
@@ -0,0 +1,75 @@
...
+@property {String} [SAML_WANT_ASSERTIONS_SIGNED] - Require signed assertions (true/false)
+@property {String} [SAML_AUTHN_RESPONSE_SIGNED] - Require signed responses (true/false)
+@property {String} [SAML_SIGNATURE_ALGORITHM] - Algorithm for signing (e.g., 'sha256')
+@property {String} [SAML_IDENTIFIER_FORMAT] - Format for name identifiers
+@property {String} [SAML_ACCEPTED_CLOCK_SKEW] - Allowed time difference in ms
+@property {String} [SAML_PROVIDER_NAME] - Display name for your service
+@property {String} [SLO_CALLBACK] - URL for handling logout callbacks
+*/
+
+const xyzEnv = {
```
RobAndrewHurst, line 54: I have also added in on the object itself the default properties. This makes for better code completion in our editors.

RobAndrewHurst (commented):
I have made some changes that I think will be better suited to the use case of environment variable consumption.
This changes help us to avoid repeat require statements on all the different modules. And can instead import it in one place.

RobAndrewHurst:
This pr is dependent on #1819 being merged in to solve the security issues above.

cityremade (commented):
thanks, looks good.

dbauszus-glx (changes requested):
The process.env variables added from the limiter merge should be added to xyzEnv module. 

```js
const limiter = rateLimit({
  windowMs: process.env.RATE_LIMIT_WINDOW ?? 1 * 60 * 1000, // 1 min
  limit: process.env.RATE_LIMIT ?? 1000, //1000 requests per 1min
  standardHeaders: 'draft-8',
  legacyHeaders: false,
});
```

AlexanderGeere:
I have added the new env variables to xyzEnv.

AlexanderGeere (approved):
Looks good to me.

RobAndrewHurst:
> I understand that the globalThis is assigned by the require in the express module.
> 
> require('./mod/utils/processEnv.js');
> 
> image
> 
> How does that work with deployments to vercel which do not execute the express module?

image

the api.js module has a processEnv required at the top of it which should add in the global object on a deployed instance.

dbauszus-glx:
I added script documentation in regards to the processEnv script file.

Should the object be assigned conditionally? 

```js
globalThis.xyzEnv ??= xyzEnv;
```

RobAndrewHurst:
> I added script documentation in regards to the processEnv script file.
> 
> Should the object be assigned conditionally?
> 
> ```js
> globalThis.xyzEnv ??= xyzEnv;
> ```

++ for this!

dbauszus-glx:
Should the type assignment be done in the processEnv script?

This was one of the arguments for using a globalScript.

ie. COOKIE_TTL is assumed to be an integer and doesn't have to be parsed on every use.

```js
parseInt(xyzEnv.COOKIE_TTL)
```

RobAndrewHurst:
> Should the type assignment be done in the processEnv script?
> 
> This was one of the arguments for using a globalScript.
> 
> ie. COOKIE_TTL is assumed to be an integer and doesn't have to be parsed on every use.
> 
> ```js
> parseInt(xyzEnv.COOKIE_TTL)
> ```

Great point. Yeah I agree that it should happen in the processEnv script

RobAndrewHurst:
> I am ok for this to be merged as it is now to be tested in deployments with the next minor release.
> 
> Items such as the RETRY_LIMIT being defined as string in the environment variables should be tested independently.

I have created a sub-issue on the issue this is linked to

PR #1820, opened 2025-01-17
It should be possible to toggle graduated theme categories on the label.

dbauszus-glx:
Graduated theme category labels should behave the same as categorized theme labels. Clicking on the label will toggle the display of the category in the mapview.

[Screencast from 2025-01-17 16-51-36.webm](https://github.com/user-attachments/assets/587f6b69-87b5-45af-9158-ad3aafa29c1a)

### DIFF
--- a/lib/layer/themes/graduated.mjs
+++ b/lib/layer/themes/graduated.mjs
@@ -39,6 +39,12 @@ export default function (theme, feature) {
 
     const cat = theme.categories.at(index)
 
+    if (cat.style === null) {
+
+      feature.style = null;
+      return;
+    }
+    
     // Spread cat style to retain scale property
     feature.style = {
       ...feature.style,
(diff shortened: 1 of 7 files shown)


AlexanderGeere:
@dbauszus-glx should we not then also include this control?
Screenshot 2025-01-20 093330
 
The switch all 'button'.

dbauszus-glx:
@AlexanderGeere I added the switchAll control from the categorized legend.

dbauszus-glx:
A themeLegendSwitch method has been created since this element is used now for graduated and categorized theme legends.

AlexanderGeere (approved):
Looks good to me! can even toggle them on and off in custom views

simon-leech (changes requested):
I have turned off all options using the 'switch all' button - but I can still see the features drawn on the map and they are still selectable. They should be styled to null? 

image

dbauszus-glx:
> I have turned off all options using the 'switch all' button - but I can still see the features drawn on the map and they are still selectable. They should be styled to null?
> 
> image

Only if they match a category. If you values which do not match a category then they will be styled as default.

cityremade (approved):
All good. Interesting this was never a thing before.

simon-leech (approved):

PR #1836, opened 2025-01-21
Material icons replacing mask-icon logic.

cityremade:
Follow-up to css restructuring in order to make possible planned ui development. 
.mask-icon css class which sets stored svg as a div background is replaced by .material-symbols-outlined 
This makes all material icons resources available to use as inline and pseudo elements where needed.

Includes changes to ui elements html content where needed.
Includes changes to padding on elements where removing mask-icon resulted in unwanted space.

Edit: I've added a legacy stylesheet to keep all classes that need to be backwards compatible.
`.expander` class is in legacy while the expanding element in restructured css has been renamed to `.caret`


simon-leech:
Initial Comments: 
1. While we are doing this, please can we move the dark plugin into the core? 
2. We need to decide how we handle backwards compatibility of existing custom plugins, currently they will be affected. 
3. Dropdowns are not working at all currently, for example a filter dropdown or theme swapping.

simon-leech:
image

Type key entry text is not visible.
Type report has no icon anymore. 


image
Do we need the small line below the selected panel option? It feels like the active colour does that job. 

Most of the button hovers in the button panel are no longer showing the dictionary term but instead the dictionary label - e.g. zoom_in

cityremade:
Replaced by https://github.com/GEOLYTIX/xyz/pull/1836

PR #1837, opened 2025-01-21
UserSettings

simon-leech:
This PR updates the interface of the `userLocale` plugin, replacing it with the `userSettings` plugin. 

The `userSettings` plugin shows the user a `dialog` that holds the ability to Save or Delete the userLocale, 
and also for the user to update their own language and save this. 

The `settings` icon has also been tweaked to improve it. 

image

To Do 
- Mapview needs refreshing once language is altered, to ensure changes are immediately seen. 
- Button should remove `active` class when toggled / save or delete is applied.
- Button does not seem to create the `dialog` the second time you press the button
- Create a user settings object that's exported so additional settings can easily be added (each will have a function that returns the HTMLElement for that setting)


jfitzpatrick17:
Nice one, this looks good!

A couple of thoughts/points in regards to how this looks...
1) Do we think this is clear enough to end users? It might help to have a bit more of an explanation, something along the lines of the following:

image

2) I wonder whether we keep it simple and remove the words 'click to'

3) Can we add a bit more spacing between the layout / default view and language settings, it feels a bit squashed together to me

dbauszus-glx (changes requested):
Please don't remove the userLocale plugin which is required for testing functionality without the need of a custom plugin.

I'd recommend to create a custom plugin first for the user settings which will be much easier to test and can always be moved to the core plugins once set in stone.

We do not want to make patch/minor releases dependent on fixing layout / functionality issues in plugins core or custom.

PR #1839, opened 2025-01-22
Filter Visual Changes

simon-leech:
1. Added new filter icon
2. zoomToExtent renamed zoomToFilteredExtent
3. zoomToFilteredExtent only shown when filter current has a length
4. The zoomButton for zoom restricted layers has been moved to the left of the display toggle too, to ensure things stay in the same place across layers. 
5. The zoomButton has a new icon that includes a plus in the centre to distinguish between our search icon, and a zoom in icon - this was causing confusing for users. 

Filter Icon will show when the layer has any filters applied to it. To indicate this to the user.
User can click the button to zoom to the extent of the filtered dataset.
image

The zoomtoFilteredExtent button can be omitted by passing a `layer.viewConfig` object without it in. 
e.g. 
``` json
 "viewConfig": {
    "displayToggle": true,
    "zoomBtn": true
  },
```


AlexanderGeere:
Changed the zoom icon to use display instead of visibilty to avoid this gap:
Screenshot 2025-01-22 155241

Inbetween the filter icon and the layer toggle.

dbauszus-glx:
I simplified the visibility condition reducing duplicate code.

dbauszus-glx:
I removed the filter function. The style can be assigned to the element on creation. That related script is close together.

https://github.com/GEOLYTIX/xyz/pull/1839/commits/b7d0b3268cea0cfd3a70d3ddbf348f4d21c8cd9e

dbauszus-glx (approved):
Works well for me.

RobAndrewHurst (approved):

PR #1842, opened 2025-01-23
Login rate limit

dbauszus-glx:
Failed login attempts should stop sending mails once the maxFailedAttempts limit is reached.

The debounceRequest method has been elevated from the user registration to the user module to debounce any login and register method requests to this module. This prevents spamming the user login from the same request address.

---
  - https://app.asana.com/0/0/1209243153938140


RobAndrewHurst:
https://github.com/GEOLYTIX/xyz/pull/1842/commits/7af6d4602f637f9dffb600e313a374f6372d9143

Fixed an optional chaining preference issue flagged by sonarcloud

RobAndrewHurst (changes requested):
## Noted behaviour.

You can lock the ip address by providing incorrect credentials.

While the ip is locked I get a login screen.

I can then provide correct credentials to the login, and then it gives me the message 'Address unknown temporarily locked.'

I can then refresh the page a bunch of times and then it eventually logs me in.

Findings: 
- Cookie is created with correct credentials provided even though ip is locked.

Is this desired behaviour?

dbauszus-glx:
@RobAndrewHurst the login screen is a view; The view api is not rate limited. The rate limit should apply to the request sending a payload only. I have now changed the check to only debounce requests with a body.

dbauszus-glx:
@RobAndrewHurst The debounce check must be before the password check. Hence it doesn't matter whether you provide the correct or wrong password for the lock. Every debounce resets the clock. If you repeatedly resend the data you should never get through. You will get through once you wait 5+ seconds after sending credentials.

RobAndrewHurst:
> @RobAndrewHurst The debounce check must be before the password check. Hence it doesn't matter whether you provide the correct or wrong password for the lock. Every debounce resets the clock. If you repeatedly resend the data you should never get through. You will get through once you wait 5+ seconds after sending credentials.

Yeah that's expected and what I was getting

RobAndrewHurst:
> @RobAndrewHurst the login screen is a view; The view api is not rate limited. The rate limit should apply to the request sending a payload only. I have now changed the check to only debounce requests with a body.

Will give this a go now!

mod/user/_user.js
```diff
@@ -49,26 +51,69 @@ The route method assigns the host param from /utils/reqHost before the request a
...
+      : 'invalid';
   }
-}
\ No newline at end of file
+
+  // The remote_address has been previously used
+  if (
+    Object.hasOwn(previousAddress, req.params.remote_address) &&
+    // within 5 seconds or less.
+    new Date() - previousAddress[req.params.remote_address] < 5000
```
RobAndrewHurst, line 108: Do you think we should maybe increase the debounce time to 10s ?

RobAndrewHurst (approved):
Otherwise I am all happy for this to go in. 👍

We can discuss the debounce time in the comments and will be happy with whatever we all decide on.

@dbauszus-glx Good job fam 👍

mod/user/_user.js
dbauszus-glx: [Screencast from 2025-01-24 15-20-26.webm](https://github.com/user-attachments/assets/91c90830-fb4a-44ed-a65e-5eb9397640ab)

I was able to hit the lock by typing my password correctly manually in less than 10 seconds. But wasn't able to do so in 5 seconds.

AlexanderGeere (approved):
Im happy with this. Nice one.

cityremade (approved):

PR #1843, opened 2025-01-24
Assign admin flag from payload.

dbauszus-glx:
The default view requests a cookie if authentication uses a token.

A token must never have admin rights.

A cookie generated from a token should not have admin rights neither.

### DIFF
--- a/mod/user/_user.js
+++ b/mod/user/_user.js
@@ -21,7 +21,7 @@ The _user module exports the user method to route User API requests.
 @module /user
 */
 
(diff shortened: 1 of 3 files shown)


simon-leech (approved):
This is working as expected for me. 

Generated a token and used that to authenticate, the token login has no admin rights. 
I tested calling the `api/workspace/test?token=XX` route too as this endpoint requires admin rights, and correctly get a message saying I don't have admin credentials. 👏

PR #1847, opened 2025-01-27
Layer keyvalue dictionary

dbauszus-glx:
The keyvalue_dictionary argument will be the layer with a keyvalue_dictionary property if called from the layer decorator.


lib/plugins/keyvalue_dictionary.mjs
```diff
@@ -27,6 +27,11 @@ Replaces key-value pairs in the mapview locale object with dictionary entries.
 @property {Object} mapview.locale The locale object of the mapview.
 */
 export function keyvalue_dictionary(keyvalue_dictionary, mapview) {
+  if (!mapview && keyvalue_dictionary.keyvalue_dictionary) {
```
RobAndrewHurst, line 30: So is the purpose of the falsy check on the `mapview` to return the value for the provided `keyvalue_dictionary` key?

Should we maybe not rather put this into it's own named function so that we can better understand what the function call is doing from where it's called?

dbauszus-glx: I export now the parseObject method from the keyvalue_dictionary plugin module to be called from a layer.keyvalue_dictionary method.

The layer decorator will only call a layer method or plugin.

simon-leech (approved):
This is brill - going to be so useful! Thanks @dbauszus-glx :)

RobAndrewHurst (approved):
I am all happy with this!

PR #1848, opened 2025-01-27
check params to prevent cross site scripting

dbauszus-glx:
The {{redirect}} input parameter should not be required in the login view template.

Parameter provided by user input, eg language must be tested before being substituted in a template [eg. default view].


AlexanderGeere:
@dbauszus-glx i fixed what sonarcloud was complaining about

AlexanderGeere (approved):

RobAndrewHurst (approved):
Looks all good to me

PR #1857, opened 2025-01-30
jsoneditor element and userlayer plugin

dbauszus-glx:
The PR adds a jsoneditor element module which dynamically imports the [vanilla-jsoneditor](https://www.npmjs.com/package/vanilla-jsoneditor) as esm module.

The _elements module has been cleaned up to sort the exported modules alphabetically and show the requires in the docs which link to the module.

image

The json location entry module has been updated to use a jsoneditor control instead of an text area input with a self-rolled json validation.

image

The json dataview has been updated to use a jsoneditor control instead of just stringifying JSON into a text area.

https://github.com/user-attachments/assets/6d229264-1086-487a-8707-c1959957a53a

The userLayer developer plugin has been added. The plugin adds a button to the layer view control. The button opens a jsoneditor control in dialog. A JSON layer can be added to the mapview through the custom jsoneditor toolbar button. This is a development tool. There is no json nor template validation!

https://github.com/user-attachments/assets/e965a1e7-fe3c-4ea2-8d47-1305e862bdad

The _plugins module has been cleaned up to sort the exported plugin modules alphabetically and show the requires in the docs which link to the individual module documentation.

image


AlexanderGeere (approved):
Very cool functionality, nice!

simon-leech (approved):
I have pushed two minor changes to the `json.mjs` file:

1. I have added an additional filter to filter out the `text` button, as this does not do anything and is confusing to the user to see this as a button that doesn't change anything. 

image
image

2. I noticed that displaying a `json` entry `inline` causes the formatting to go quite odd. I don't think we should handle the CSS just for this use case so instead I put a warning in that you shouldn't display them inline, as this feels like a proper edge case use. I couldn't actually remove the `entry.inline` as this is applied in the `infoj.mjs` file, but hope a warning is enough to prevent people from displaying json data in this way.  

image
image



I really like the `userLayer` plugin, that's very cool - and the tree view within that is super helpful for easily seeing what's provided in the `json`


All happy with this :)

PR #1860, opened 2025-01-30
Store view params on mapview.locale.view

dbauszus-glx:
The current view params are now stored in the locale view allowing the parameter to be stored in a userLocale.

I noticed that view currently not correctly assigned.

I fixed this in the mapview module viewConfig method.

### DIFF
--- a/lib/mapview/_mapview.mjs
+++ b/lib/mapview/_mapview.mjs
@@ -210,16 +210,16 @@ function viewConfig(mapview) {
 
   mapview.view.projection = `EPSG:${mapview.srid}`;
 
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):
I can see a view parameter in `mapview.locale.view`, everything else working as expected.

RobAndrewHurst (approved):
LGTM!

simon-leech (approved):
This is working for me! 
1. Tested in default MAPP.
2. Tested with userLocale and correctly saves the `view`. 
3. Tested in a custom view with multiple mapviews and the `view` is still being applied correctly.

PR #1864, opened 2025-02-01
CSS - Material Symbols / Dark Mode plugin (Redux)

RobAndrewHurst:
Mask icon block vector now replaced with inline Material Icons.
Dark Mode plugin included in core.
HTML in ui elements simplified where possible.

Dark mode CSS in one stylesheet ordered by components. Can be removed when no longer needed.

This PR requires review along with v4.13.0 update to relevant plugins. 

*_**⅀  Summary of Changes**_

* 

Minor changes to dictionaries

**How Icons now work**

Custom made code should create elements with the following structure:
`<button data-id="zoom-to" class="material-symbols-outlined" title="Zoom to Extent">map_search</button>`

Icon elements can be buttons, divs, spans with the class **"material-symbols-outlined"** and the selected icon name as text content.

Inside hyperlinks or buttons `<span class="material-symbols-outlined">` can be used to contain an icon next to text.

**Alternative way to use Material Icons:**
Use a tag and class which will apply content to pseudo element:

```
/*css*/
.right-arrow:after {
   content: 'arrow_right_alt';
}
```

```
<!--html-->
<span class="material-symbols-outlined right-arrow">
```

**Core plugins**

* All icons in plugin containers replaced with Material Icons where mask icon was previously used
* Link button plugin: configurable symbols now automatically apply “material-symbols-outlined” css class, `css_class` and `css_style` parameters still in use. New parameter `icon_name` now supported - that way any symbol from Material Icons library can be used

```
{
    "href": "/my_url",
    "title": "My title",
    "css_style": "color: blue",
    "icon_name": "home"
  }
```

**UI Elements**

* expandable drawers and panels - new expanding arrows
  * in order to use expanding arrow anywhere use `<div class="material-symbols-outlined caret"/>`
  * similar case with .close, .toggle, .minimize, .fullscreen
* dropdown - fixed arrow icon - with dropdown open the arrow didn’t rotate previously
* checkbox - now using Material Icons checkbox style
* dialog - Material Icons now used in the toolbar for min/ax/imise, close

**Entries**

* cloudinary - file upload input now has a drag and drop zone
* link - new parameter `icon_name` for applying any Material Icons symbol

**Location view**

* reordered elements in location header
* save new changes icon now a green cloud to attract attention
* delete location now a red bin
* remove from selection now a subtle cross

**Colours**

* added 2 new colours to colour scheme - red no-colour equivalent for dark mode and green ok-colour for information 
* dark mode grouped in one stylesheet for further work

**CSS structure**

* stylesheets for ui elements and layout now separated
* separate stylesheet extracted where possible and relevant as of now
* some elements from default view that are referenced further within plugins (#mapButton etc.) also moved to their respective stylesheets. This reduced the style tag in the default view.
* hardcoded colours replaced with variables

These changes need to be reviewed against plugins v4.13.0 from pr#286 from plugins repository. Ideally plugins are merged in beforehand.


dbauszus-glx (changes requested):
I find the drawer elements very hard to read with dark_mode enabled.

image

simon-leech:
This is going to seem negative but not at all - I really like the new UI, it feels much fresher! I'm just pointing out the current issues, but that's not to say it's all bad as I really like it! 😅 I really like the new checkbox UI, that's much cleaner! 

**Suggestions**
This icon is unclear to me compared to the previous filter icon is there a reason we changed this one? It's not obvious this means that a filter is being applied?
image 

Is there a reason we changed this one too - I don't find this intuitive that is means you need to zoom in to view the data. 
image

Is there a reason we changed these two buttons from the plus / minus (the current zoom in one) was used for the zoom to extent previously? 
image

The modeller icon looks a little weird, like its being duplicated inside the element? 
image

**Dark Mode**
When you turn on dark mode, layers that are not in a group have no border around them, like they do in light mode.
 
image
image

cityremade:
Here's how to change icon when using link_button plugin:
`"link_button": [
  {
    "href": "/url/url",
    "title": "Link Title Here",
    "css_style": "color: blue;",
    "css_class": "bold",
    "icon_name": "grid_view"
  },
  ...
]`

css_class should no longer contain 'mask-icon', etc.
new parameter is `icon_name` which is the actual icon name from Material Icons library 
https://fonts.google.com/icons 
Screenshot 2025-02-04 at 15 23 59

dbauszus-glx:
Some icons look like they are disabled, even if they are not. The active icon is not reflected by an active colour. The icon should reflect if the measure_distance plugin is active.

image

AlexanderGeere:
I like the new look just had a few suggestions i thought i would float by. 

How would we feel about adding a background color to the controls to create a bit of a distinction:
Screenshot 2025-02-06 112805
Like that for example. 

Also making the toggles a bit bigger? 
Screenshot 2025-02-06 112912
This is `font-size: 1.8em` the width and height could also be adjusted following this change.

The overall panel would look like this:
Screenshot 2025-02-06 105325

simon-leech:
I pushed a change to the zoom in button to try using this one instead. I think its clearer to indicate you need to zoom in to a 'point' to show the data 
image

simon-leech:
In dark mode - i see a little white bar across the bottom of the layer panel. 

image

The gazzetteer searching colours are white. 
image

dbauszus-glx:
The control panel element should be sized by available space only and not contain a magic pixel value, eg. 50.

image

dbauszus-glx:
The feature_info plugin get's listed as Current Location, just like the locator plugin.

Most button use a span for the icon, but the feature_info plugin uses a div for the icon?

Does it matter? I'd prefer if we use one or the other consistently rather than like this and that.

dbauszus-glx:
Please retain the css rule for active mask-icon button. Otherwise older plugins become unusable.

image

dbauszus-glx:
I resolved the issue with locator and feature_info plugin icon and dictionary.

The active colour for legacy plugin icons has already been resolved.

dbauszus-glx:
I solved the height 100% - 50px issue with a flex column.

@cityremade Can you check this is working for you?

The #ctrl elements should be a flex column. With this neither the #ctrl-tabs nor #ctrl-panel should require a height property but will the space.

image

dbauszus-glx (approved):

AlexanderGeere:
I added a flex-grow: 1 to #ctrl-panels to avoid dropdowns at the bottom on the panel from going behind it: 
Screenshot 2025-02-17 155556

AlexanderGeere (approved):
This is looking good to me!

simon-leech (changes requested):
Pushed a few minor changes: 
1. Add Alt to Image for accessibility
2. Add aria-label to searchbox for accessibility
3. Removed overwride of infoj-key to ensure this goes orange when layer is active

@cityremade  
1. We may want to address a low-contrast error using Orange on white background for accessibility? 
2. Please can we have a warning from the link_button if you use a `mask-icon` inside the `css_style`?

cityremade (approved):
happy with all new changes.

cityremade:
in the header drawer white-space: nowrap replaced with normal;

simon-leech (approved):
image

I fixed an issue with the close buttons being greyed out by updating the button to use a `class` of "material-symbols-outlined" instead of "material-symbols-outlined-close". 


I am all happy with this PR now - great work @cityremade  🎉🎉

PR #1867, opened 2025-02-03
The temporal utility mod must not throw an error to take down infoj p…

dbauszus-glx:
The temporal utility module should be able to handle null, undefined, or string values (including empty).

The integer value required for the Date generation should be parsed from entry value (or newValue).

An error should be logged instead of thrown with the method returning and not crashing the infoj process.


RobAndrewHurst (approved):
Looks all good to me!

simon-leech (approved):
This is working for me :)

PR #1868, opened 2025-02-04
Use workspace key as db for userlocale

dbauszus-glx:
An object store must be referenced to an XYZ instance workspace which may contain multiple locales.

Locale keys in the workspace must not be modified as they are used as reference in queries to the XYZ host.

Any XYZ instance has a TITLE. If not specified in the environment variables the title will be set in the processEnv module.

The XYZ TITLE will be assigned as workspace.key property if not explicit set in the workspace object.

The workspace.key will be assigned as workspace property to locales returned from the XYZ workspace module.

A user [indexed] DB is unique by combining the user name with the locale.workspace property.

This allows to store multiple locale with the same key but different names in the same 'locales' [object] store in the user DB.

The test plugin interface has been updated to allow for input of a locale name before storing or removing a locale in the 'locales' object store.

A method to remove a locale has been introduced to supersede the method to delete a user DB.

A list method returns an array of locales [key, name] from the 'locales' object store.


cityremade (approved):
I am happy with how these methods work.

simon-leech:
Review in progress - please don't merge

simon-leech (approved):
I added a new method `putLocale` which calls the `put` method. 
The `put` method has been generalised to allow for any object to be stored to an objectstore. This in the future could be something like a userSettings object that is stored against a workspace, rather than against a locale. 

The methods are all working as expected for me :)

PR #1872, opened 2025-02-06
Jenks must only parseFloat valid numbers 

simon-leech:
Addresses issue as we must first filter out any features with a field value of null. This ensures that we don't crash the parse float process


AlexanderGeere (approved):

dbauszus-glx (approved):
image

The legend is correct with the applied fix.

image

I did a commit to pass the filter through the map method. Rather than filter to create an array and then iterate the array again to parseFloat on itself.

RobAndrewHurst (approved):

PR #1881, opened 2025-02-07
Display check for viewport param reload on changeEnd.

dbauszus-glx:
This PR fixes an issue where layers with a viewport parameter requested a reload on changeend even if the layer display was toggled off (through zoom level restrictions or otherwise).


simon-leech (approved):

cityremade (approved):

AlexanderGeere (approved):

RobAndrewHurst (approved):
All happy with this!

Issue #1875, opened 2025-02-07
Drawing fails on a layer with geoms


dbauszus-glx:
It is not possible to create new geometries on layers with a geoms configuration. The draw methods should warn and not offer the option to draw.

dbauszus-glx:
This has been resolved in a path PR. The drawing panel will not be shown if the layer does not have a geom field configured.

simon-leech:
An empty draw method is no longer assigned.

PR #1885, opened 2025-02-10
Return warning if drawing with no geom property

simon-leech:
## Description

Addresses the attached issue. 
Added a check on the property `geom` as this must be provided to draw new features to the layer. 
This check occurs after the `elements.length` check to ensure that you only see a warning for layers that have a `draw` property assigned to them. 

This should be tested with drawing to a layer and also drawing to a field on a location - as both use the same `draw` method. 

## GitHub Issue
https://github.com/GEOLYTIX/xyz/issues/1875

## Type of Change

## Testing Checklist 

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


cityremade (approved):
to clarify: this prevents drawing ui from being created on special non-vector layers.

AlexanderGeere (approved):

dbauszus-glx (changes requested):
The short circuit should happen earlier in the execution of the draw default method. I would do this before or after the hidden flag check. Definitally before the elements check.

The module is missing a documentation stub and default export function should be named and documented.

dbauszus-glx (approved):
I added the drawPanel function documentation and moved the geom check ahead of the draw config object iteration.

PR #1887, opened 2025-02-10
Improve parsing performance for keyvalue_dictionary execution.

dbauszus-glx:
The keyvalue dictionary can be a huge array of objects.

Filtering an array of tens of thousands of objects for the last match on key and value can take several minutes.

Using a weakMap with a [key, value] array as key [index] is a lot more performant. The same lookup should only take milliseconds.

The findLast is irrelevant since the weakMap creation will overwrite earlier [key, value] which are the same symbol.


simon-leech (approved):
This is faster for me than before - thanks @dbauszus-glx ! 💯

PR #1894, opened 2025-02-11
update layer.draw in decorator

dbauszus-glx:
Not every layer should get an empty draw object property assigned. The draw property indicates that it should be possible to draw on the layer.

The comment ` // Layer must have an empty draw config to allow for role-based assignment of drawing methods.` doesn't really mean anything to me. I would need to see an example configuration for this.

The `layer.edit` property has long been retired. There shouldn't been a check and warning on this anymore. If used you will not see the drawing elements and won't be able to draw on the layer. No need to warn on this and assign the draw property keeping outdated configuration in place if they even exist.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -55,18 +55,6 @@ export default async function decorate(layer) {
     zoomToExtent,
   });
 
-  // Warn if outdated layer.edit configuration is used.
-  // Set layer.draw to layer.edit if it exists.
-  if (layer.edit) {
-    console.warn(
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):
I seem to be able to assign role based drawing, so this looks good to me. Also gets rid of the warning appearing on every layer.

simon-leech (approved):
I also cannot work out what that comment meant - I can use roles for drawing or not fine. 
All good with me

RobAndrewHurst (approved):

PR #1895, opened 2025-02-11
Check layer format in layer decorator

dbauszus-glx:
Instead of checking for the format in the addLayer method before calling the layer decorator method the layer format should be checked in the layer decorator method itself as this method will also be called from the [location] layer entry method which will otherwise err.

With the check for the format a warning will be issued if the format is missing in the infoj layer entry method.

The layer entry decorate method should shortcircuit on the existence of the openlayers layer property `.L` not the show method instead of conditionally calling the decorator method on a nullish condition of the show [method] entry property.


simon-leech (approved):
This is working for me - tested providing a bogus layer entry and correctly see a warning from that

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #1904, opened 2025-02-17
Core test plugin & Integrity Tests

RobAndrewHurst:
# Core test plugin & Integrity Tests

The test plugin has now been included in the core plugins.

The Integrity tests have been changed to run on the latest version of codi with the core test plugin.

To run the different tests you can provide the `test` param as part of the url params.

eg.

`/?test=core` - run the core front end tests
`/?test=integrity` - run the integrity tests

Config required on a locale:

```json
{
  "test": {
  "options": { <-- Options passed to the runWebTestFunction
    "quiet": true, <-- will only show errors (Defaults to false)
    "showSummary": true, <-- will show a summary (Default to false)
    }
  }
}
```

> [!IMPORTANT]
> These tests are to be revised with a new strategy where keys in a layer will be used to execute specific tests.


simon-leech (changes requested):
@RobAndrewHurst  Please can you resolve the Merge Conflict for me before I test this out?

dbauszus-glx (changes requested):
I worked out that I need to add the `core=test` URL parameter to run the coreTest.

I am not entirely sure how the _mappTest global works. How is this assigned?

I am also unsure about the test plugin assigning a plugin. ie mapp.plugins.integrityTests How is this supposed to work?

Can the import for the codi framework by dynamic like we import maplibre, turf, and other libraries only if needed and not as default. 

I don't understand the options property of the plugin config. The plugin object itself is a config option. It feels a bit ya dawg to have config in config.

RobAndrewHurst:
> I worked out that I need to add the `core=test` URL parameter to run the coreTest.
>
> I am not entirely sure how the \_mappTest global works. How is this assigned?

> [!NOTE]
> This gets assigned to the globalThis in the /tests/\_mappTest.mjs module. It's follows the same pattern as the mapp.mjs module

> I am also unsure about the test plugin assigning a plugin. ie mapp.plugins.integrityTests How is this supposed to work?

> [!NOTE]
> With regards to this, I need the plugin to be added and execute after the mapview itself has loaded for the integrity tests. There are other options for this in the future, but this is just a work around for now.
> The test gets triggered in the default view if the hooks are present.
>
> ```javascript
> //Execute integration tests
> if (mapp.hooks.current.test === "integrity") {
>   await mapp.plugins.integrityTests(mapview);
> }
> ```

> Can the import for the codi framework by dynamic like we import maplibre, turf, and other libraries only if needed and not as default.

> [!NOTE]
> This has been done.

> I don't understand the options property of the plugin config. The plugin object itself is a config option. It feels a bit ya dawg to have config in config.

> [!NOTE]
> The test runner takes in a set of parameters, so I am thinking that we can just provide that as an object to the runner. I can then rename options to something like 'runnerOptions' or 'runOptions' as these are specific to codi. Things around it can be used to extend the plugin itself.

> Is the icon change in the default view related to this PR or does this come from a different PR?

> [!NOTE]
> This has been resolved. Bad merge on my end

> Should the test view be removed in this PR?

> [!NOTE]
> This has been done.

simon-leech (dismissed):
I pushed two minor commits: 
1. To remove the console log of the mapview.
2. To ensure that the `layer.filter.current` is passed to the `getLastLocation` - as without it its possible to be selecting locations that cannot be accessed on that layer. 

Issues I'm currently seeing: 
1. If I have access to multiple locales, the testview should loop through them and run the tests on all available locales for the user logged in. 
2. It doesn't seem to be removing layers from the display and URL when the layer test has ended.

RobAndrewHurst:
@simon-leech 
> I pushed two minor commits:
>
> 1. To remove the console log of the mapview.
> 2. To ensure that the `layer.filter.current` is passed to the `getLastLocation` - as without it its possible to be selecting locations that cannot be accessed on that layer.
>
> Issues I'm currently seeing:
>
> 1. If I have access to multiple locales, the testview should loop through them and run the tests on all available locales for the user logged in.

> [!IMPORTANT]
> This isn't a thing, and will only be implemented in another iteration.

> 2. It doesn't seem to be removing layers from the display and URL when the layer test has ended.

> [!NOTE]
> Is this important?

AlexanderGeere (approved):
This is looking good on my end.

simon-leech:
@RobAndrewHurst no issue with the locale looping, that can wait- apologies should have put that under a different request. 

I do wonder if the turning off display should be included though.
When you have lots of layers in an application, and it's constantly turning them on for the tests, you could potentially crash the application trying to turn on 20+ layers all at once?

RobAndrewHurst:
> @RobAndrewHurst no issue with the locale looping, that can wait- apologies should have put that under a different request.
> 
> I do wonder if the turning off display should be included though. When you have lots of layers in an application, and it's constantly turning them on for the tests, you could potentially crash the application trying to display 20+ layers all at once?

Are there any failures?

Because it should be turning them off, but if there is a failure the layer will remain on.

dbauszus-glx (approved):
This is working well for me now.

PR #1910, opened 2025-02-18
disabled nested elements in view changeend method.

dbauszus-glx:
Instead of just the nested drawer elements all elements but the .header class element in the drawer scope should be handled by the changeend event method to ensure that button in the drawer are disabled or enabled according to the layer being zoom range.


simon-leech (approved):
Working for me - custom plugin buttons are correctly disabled when the layer is disabled, and enabled when the layer is accessible.

RobAndrewHurst (approved):

PR #1916, opened 2025-02-19
Return error message on view template request

dbauszus-glx:
A language template request will return the template string for failed lookups. This ensures that an error message is provided where the lookup for an error message fails.

However a failing view template lookup should return the error message itself. This ensures that the reason for the failed lookup is returned.


simon-leech (approved):
Working for me - correctly see the error about incorrect file path rather than 'default_view'

AlexanderGeere (approved):
Seeing the error as expected.

PR #1920, opened 2025-02-21
Checkbox edited highlight;

cityremade:
* Checkbox with value changed now gets highlighted label background
* Font colour set to off black so that it is retained in dark mode - previously would flip to white;
* Input border colour moved to colour root variable.

### DIFF
--- a/public/css/_inputs.css
+++ b/public/css/_inputs.css
@@ -43,7 +43,7 @@ input[type='search'],
 input[type='date'],
 input[type='time'],
 input[type='datetime-local'] {
-  border: 1px solid #ccc;
+  border: 1px solid var(--color-input-border);
   padding: 5px;
 
   &:focus {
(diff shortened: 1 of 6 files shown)


simon-leech (changes requested):
@cityremade  On the issue I noticed that it was both checkboxes and input boxes. 
Checkbox is working ok for me now, but input text box still not. 

Looks like the CSS is being overwritten so its preventing the editing class from being shown? 
image

simon-leech (approved):
Looking good to me across dark and light mode  -  all input types are correctly highlighted when edited.

AlexanderGeere (approved):

dbauszus-glx (changes requested):
I think this would look cleaner with the checkbox element highlighted and not just the label part of the checkbox element.

There is an issue with the checkbox label not being displayed when a boolean entry is not editable.

[Screencast from 2025-02-24 16-18-34.webm](https://github.com/user-attachments/assets/7ce61d1f-9230-4022-9b4a-e3c1a03fd16d)

dbauszus-glx (approved):

PR #1925, opened 2025-02-21
Dialog - css changes;

cityremade:
1) header now sticky
2) padding now removed from parent container as this caused scrolled area visible over the header 3) added margin right on first child of header to prevent cramming text when minimized.


dbauszus-glx (approved):
Lovely work.


https://github.com/user-attachments/assets/701772ee-232d-404d-a77c-8c040cbacc6f

dbauszus-glx (changes requested):
The default padding looks rather heavy when tested with a plugin.

image

The container did not become scrollable after the content exceeded the dialog content.

image

cityremade:
Multi filter local css updated to fix this.

dbauszus-glx (approved):

PR #1956, opened 2025-03-03
alert-confirm doesn't need space for minimize/close button.

cityremade:
Header margin override for confirm and alert dialogs.

### DIFF
--- a/public/css/elements/_alert_confirm.css
+++ b/public/css/elements/_alert_confirm.css
@@ -12,6 +12,12 @@ dialog.alert-confirm {
     outline: none;
   }
 
+  & > header {
+    & > :first-child {
+      margin-right: 0;
+    }
+  }
+
   h4 {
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
Is this required?

I can see that this overrides the _dialog.css rule to set the margin-right to 5em for the first-child.

If this is needed then must be explained why this is needed otherthise the arbitrary 5em margin-right should be removed.

```css
  & > header {
    display: flex;
    justify-content: space-between;
    position: sticky;
    top: -1px;
    left: 0;
    background-color: inherit;

    & > :first-child {
      margin-right: 5em;
    }
```

simon-leech (approved):
Fine by me - alert confirm looks as expected for me 👍

dbauszus-glx (approved):

PR #1957, opened 2025-03-03
Locale without layers should not crash the process

dbauszus-glx:
An empty layers object should be assigned to a locale without layers.

This workspace would crash the _workspace module otherwise.

```js
{
  "locale": {}
}
```

The mapp library would also fall over on the layers length being undefined rather than 0.


AlexanderGeere (approved):
Working as expected.

cityremade (approved):

PR #1958, opened 2025-03-04
remove browser-sync

RobAndrewHurst:
# rm-browser-sync 🚚

## Why? 🤔

The browser sync dependency and scripts have been removed as there are security
vulnerabilities that are being flagged.

This isn't used that much as it as
it's not that seemless to be started and stopped. This will be re-addressed
with possible other solutions.

## Check ✅

- Documentation and scripts have been updated.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -116,7 +116,7 @@ This ensures that your test environment always has the latest changes without ma
 
 ### VSCode Tasks & Launch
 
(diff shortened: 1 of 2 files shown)


dbauszus-glx (approved):
I can confirm that I am now able to install, build, and run mapp without any vulnerabilities.

image

simon-leech (approved):

PR #1960, opened 2025-03-04
Review colour schema and update dark_mode plugin

dbauszus-glx:
The mapp colour schema needs be reviewed.

We should not use colour definitions like light/lighter, dark/darker, etc.

Colours should be hirachical/functional. ie primary, secondary, tertiary, on/off, active, etc.

This allows to defined colours in colour schemas, ie a dark schema, or a contrast schema.

Instead of assigning a dark colour variable to an element which had previously light colour variable we change the colour definition of the colour variable in the root.

There should not be more than a dozen colours which are almost the same. Ideally there is only primary, secondary, and tertiary, with the addition of functional colours.

The --dm colours need to be removed from the _colours.css sheet.

Any colour in the css sheets should be a variable.

The _dark_mode.css must be removed.

The dark_mode plugin needs to be updated to switch the values assigned to the colour variables.


public/css/base/_colours.css
```diff
@@ -1,95 +1,72 @@
...
-  --color-mid: #e3e3e3;
-
-  --color-on: #d57120;
-
-  --color-no: #a21309;
-
+  /* begin default colour palette */
+
+  /* Backgrounds */
+  --color-base: #f2f2f2;
```
RobAndrewHurst, line 5: Do we need to give all of these variables the color prefix? 

Do we feel it's necessary or a bit redundant?

RobAndrewHurst (changes requested):

public/css/base/_colours.css
cityremade: css can have other variables. These would be strictly colour related so I've left them.

public/css/base/_colours.css
```diff
@@ -1,95 +1,72 @@
...
 * {
   transition: 0.2s ease-in-out;
   transition-property: background-color;
 }
 
-/* light mode classes */
-
 .primary-colour {
-  color: var(--color-primary);
+  color: var(--color-accent);
```
RobAndrewHurst, line 37: This is confusing to me.

this class is 'primary-colour' but the variable referenced is 'accent'?

Same with the one below

dbauszus-glx:
@cityremade Did you build the mapp-docs pages? I find it hard to read the toggleDarkMode method description on the dark_mode module.

image

dbauszus-glx:
There should be a utility method which documents and overrides the colour theme with a provided theme eg from the dark_mode plugin.

I do not understand the base, base-secondary, base-tertiary, base-contrast.

These colours look almost the same to me.

image

dbauszus-glx:
Accent colours only make sense in combination with a colour.

image

dbauszus-glx:
The dark_mode plugin requires and calls the cssColourTheme utility.

image

dbauszus-glx:
`--color-base` and `--color-secondary` should not be the same colour.

`--color-base` should not be used for the `#ctrl` element in the default view. That element should not have a background. The base colour should be assigned to the body instead as this is the base colour.

`--color-base` is assigned to multiple elements in the _tabview.scss. Are elements painted in the same colour on top of each other? Is it possible to assign the base colour to the parent tabview element?

`--color-base` should be assigned to the elements which should have the base colour. eg. the dialog background. There is no need to have a secondary colour variable which is the same as the base colour.

https://github.com/GEOLYTIX/xyz/pull/1960/commits/3854f683ef96c08a8deb7154e43e6099a62cbe20

AlexanderGeere (commented):
I think i found a bug with the dark_mode plugin, not sure if we want to address it here. 
If I turn on dark_mode and then refresh i get this error: 
Screenshot 2025-04-01 162143

dbauszus-glx:
I changed base-border to border since the colour variable is only used for borders and has inherent connection to the base variables which are used for backgrounds only.

RobAndrewHurst:
> I have updated the colour description from previous commits.
> 
> @RobAndrewHurst Is there a way to show the colour in the rendered markup?
> 
> image

Not too sure. There is probably a plugin out there. But nothing I know off the top of my head!

dbauszus-glx:
`accent-hover` is now `hover` for the same reasons. The accent doesn't make sense without a reference colour.

All state colours are now described.

image

dbauszus-glx:
`text` has been changed to `font` since this applies to all font elements as default, including font icons, not just text characters.

dbauszus-glx:
Instead of primary, secondary, and tertiary font [text, icon] colours the font colours are arranged as default, contrast, and mid to sit inbetween.

image

dbauszus-glx:
I restored the primary colour. This does not make sense as accent. An accent only makes sense in regards to another colour. But the primary colour provides identity and should stand out from the base colour. Hence not an accent. It also doesn't make sense to have an accent colour assigned to the primary colour css classes.

dbauszus-glx:
I removed the `base-contrast` colour since it was not at all in contrast with the base colour. I assigned the tertiary colour instead which is slightly offset to the secondary colour which currently is exactly the same as the base colour.

dbauszus-glx:
I assigned the border colour where secondary colour was used for border or box-shadows. A box-shadow must work with a border colour. I renamed background to background-color where base-secondary is used as background-color.

dbauszus-glx:
Removed tabulator css.

dbauszus-glx:
I added the custom theme plugin.

```js
    "custom_theme": {
      "base": "#eadbcb"
    },
```

image

dbauszus-glx:
The setTheme method will now remove the current theme entries before assigning the colourTheme params as current Theme.

RobAndrewHurst (approved):

public/css/base/_colours.css
cityremade: naming resolved.

AlexanderGeere (approved):

simon-leech (changes requested):
Minor Comments: 
1. Selected options in a multi-select dropdown are impossible to see in light mode. Slightly easier to see in dark mode but still not ideal. 
2. In editing mode - the text, numeric and checkbox inputs are blue to indicate they are editable. But the dropdowns don't change even though they are editable too?
3. Gazetteer search has pills with grey on black text which is quite hard to read. In dark mode its grey on white which is a bit better.
4. The small cross for gazetteer searching is hard to see in dark mode.
image
5. When swapping between light and dark mode with a drawing dialog open- there's a small delay on the header of the drawing dialog updating. Is this expected? All the other CSS changes apply immediately so this looks a bit strange.

simon-leech (approved):

PR #1963, opened 2025-03-06
The string value of like filter must be decoded.

dbauszus-glx:
A string value provided as argument to the ILIKE SQL filter condition will not be matched with encoded values, eg. `%20` for spaces.


RobAndrewHurst (approved):
LGTM

Happy with this!

added in a new test, but will be relooked at in the minor

PR #1965, opened 2025-03-06
Dynamic modules

RobAndrewHurst:
# Dynamic module import

Found an issue with the modules being imported for module queries in the
`getTemplate` module.

It was using cjs instead of the newley adopted esm structure.

> [!NOTE]
> This has be fixed 🩹

## Config Changes

Query modules that are still using the old cjs syntax will need to be adjusted.

> [!NOTE]
> Something like this will need to change to...

```javascript
module.exports = (_) => {
  return "I am a bad module query fam";
};
```

> [!NOTE]
> THIS!

```javascript
export default (_) => {
  return "I am a module query fam";
};
```

## Aditional changes

The workspace src can now be assigned via the `force` flag so that we can
reload a workspace for cli tests.

The reason for this is for local testing.

The template.type check on a module has also been removed as this is legacy code.

## Tests

> [!WARNING]
> The mocks are going to need to be addressed to ensure that they are being
> mocked in one place. This will ensure that we only mock the implementation
> once going forward which will make things easier to manage. An issue will be
> created for this.


dbauszus-glx (changes requested):
I am ok with this in principle. Though I'd like to do some research in the future whether it might be feasible to import the module directly from the file system / url rather than from an data [url] string.

I tested a simple query module:
```
export default (_) => {
  console.log(_);
  return `SELECT '${JSON.stringify(_.workspace?.templates?.query_module)}'`;
};
```

With following templates configuration:
```js
    "query_module": {
      "dbs": "NEON",
      "src": "${RESOURCES}/bugs_testing/templates/query_module.js",
      "module": true,
      "value_only": true
    }
```

This worked as expected. But I still need to update the getTemplate module documentation in this PR.

What should happen with a bad module?

eg.
```js
export default /(_) => {};
```

The error message `Unable to parse template string.` is not meaningful in this regard. The query module should not attempt to parse this.

The getTemplate prints following error message:
image

Is there a use case for the `template.type === 'module'` check? I feel this maybe legacy code which isn't actually used and can probably be removed.

How can I test the following condition with a module which does not have a default export.
```js
template.render = importedModule.default || importedModule;
```

This PR also updates the provider/file module. The default export method from that module should be named and documented. Can you comment on the use of `import.meta.dirname` either in the script or the function docs.

This PR also makes changes to the cacheWorkspace(src) method in the cacheWorkspace module. The src param must be docoumented. Is there a test for this where I can see this working in debug?

There should be a ticket for this and the changes should be outlined in the PR.

RobAndrewHurst:
@dbauszus-glx 

> I am ok with this in principle. Though I'd like to do some research in the future whether it might be feasible to import the module directly from the file system / url rather than from an data [url] string.
>
> I tested a simple query module:
>
> ```
> export default (_) => {
>   console.log(_);
>   return `SELECT '${JSON.stringify(_.workspace?.templates?.query_module)}'`;
> };
> ```
>
> With following templates configuration:
>
> ```js
>     "query_module": {
>       "dbs": "NEON",
>       "src": "${RESOURCES}/bugs_testing/templates/query_module.js",
>       "module": true,
>       "value_only": true
>     }
> ```
>
> This worked as expected. But I still need to update the getTemplate module documentation in this PR.
>
> What should happen with a bad module?
>
> eg.
>
> ```js
> export default /(_) => {};
> ```

> [!NOTE]
> This isn't an example of a bad module.
> A bad module would be old syntax
>
> ```javascript
> module.exports = (_) => {
>   return "I am a bad module query fam";
> };
> ```

> The error message `Unable to parse template string.` is not meaningful in this regard. The query module should not attempt to parse this.

> [!NOTE]
> I will have a look into this!

> The getTemplate prints following error message: image
>
> Is there a use case for the `template.type === 'module'` check? I feel this maybe legacy code which isn't actually used and can probably be removed.

> [!NOTE]
> I am not too sure but the general syntax in config should be:
>
> ```json
> {
>   "query_mod": {
>     "src": "a-source-to-somewhere",
>     "module": true
>   }
> }
> ```

> How can I test the following condition with a module which does not have a default export.
>
> ```js
> template.render = importedModule.default || importedModule;
> ```

> [!NOTE]
> I have added in a test to the `/test/mod/workspace/getTemplate.test.mjs` module where it will test how to execute a module that get's added in with no default.
>
> ```javascript
> export function foo(_) {
>   return "I am a module query fam";
> }
> ```
>
> ```javascript
> await codi.it(
>   {
>     name: "get mod query",
>     parentId: "workspace_getTemplate",
>   },
>   async () => {
>     const template = "mod_query_no_default";
>
>     const { default: getTemplate } = await import(
>       "../../../mod/workspace/getTemplate.js"
>     );
>
>     //Get template from workspace
>     const result = await getTemplate(template);
>
>     //execute the named function from the render method.
>     const foo = result.render.foo();
>
>     codi.assertEqual(foo, "I am a module query fam");
>   },
> );
> ```

> This PR also updates the provider/file module. The default export method from that module should be named and documented. Can you comment on the use of `import.meta.dirname` either in the script or the function docs.

> [!NOTE]
> This is now done.

> This PR also makes changes to the cacheWorkspace(src) method in the cacheWorkspace module. The src param must be docoumented. Is there a test for this where I can see this working in debug?

> [!NOTE]
> I have now documented this in the module.
> There isn't currently a direct test for this, but you can see an example of it's usage in the `/test/mod/workspace/getTemplate.test.mjs` module.

> There should be a ticket for this and the changes should be outlined in the PR.

> [!NOTE]
> I have created an issue for this here: https://github.com/GEOLYTIX/xyz/issues/1967

AlexanderGeere (commented):
This is working as expected for me. I just had a thought with regards to loading bad modules i.e. cjs modules could we get a similar message to when like a layer template fails to load e.g.
Screenshot 2025-03-07 134642

dbauszus-glx:
I have updated the provider/file method documentation.

image

This method is not required to be async. The file system is accessed through the readFileSync method.

RobAndrewHurst:
> This is working as expected for me. I just had a thought with regards to loading bad modules i.e. cjs modules could we get a similar message to when like a layer template fails to load e.g. Screenshot 2025-03-07 134642

I think we are going to have to make a separate issue for this, as this isn't that straight forward.

dbauszus-glx:
The src for the workspace cache is only used in the cacheWorkspace method.

Instead of assigning this in the module closure it is much easier to just assign this from global xyzEnv in the cacheWorkspace method. This way it will always be updated if the xyzEnv workspace variable changed.

https://github.com/GEOLYTIX/xyz/pull/1965/commits/e9f9ed673ab3d733a6809b755aad2293c7d7ebe1

RobAndrewHurst:
> The src for the workspace cache is only used in the cacheWorkspace method.
> 
> Instead of assigning this in the module closure it is much easier to just assign this from global xyzEnv in the cacheWorkspace method. This way it will always be updated if the xyzEnv workspace variable changed.
> 
> [e9f9ed6](https://github.com/GEOLYTIX/xyz/commit/e9f9ed673ab3d733a6809b755aad2293c7d7ebe1)

This is exactly what I had before..

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #1972, opened 2025-03-10
Check mapbox style and accessToken

dbauszus-glx:
This PR adds a check for a valid Mapbox accessToken to retrieve the style from the mapbox api for use in a maplibregl format layer.


dbauszus-glx:
@RobAndrewHurst This should be a patch PR once 4.13 is released.

lib/layer/format/maplibre.mjs
```diff
@@ -46,10 +46,22 @@ export default async (layer) => {
 
   layer.mapview.Map.getTargetElement().prepend(layer.container);
 
+  // Check mapbox style URL and accessToken
+  if (layer.style?.URL && layer.accessToken) {
```
RobAndrewHurst, line 50: Is it always required to have an access token?

dbauszus-glx: Yes. Because we only use this mapbox styles. We never tested this with a different style.

RobAndrewHurst: should the `layer.style.object` property be the only property being set on the
MapLibreGL libary?

Because if the accessToken is required then whats the point of using either or?

dbauszus-glx: Happy for you to change this. I am ok with this as long as current configurations check the key before loading, and it is possible to expand this in the future with configurations that have styles which do not require an access token.

dbauszus-glx:
@RobAndrewHurst I added a check on the style.object.

RobAndrewHurst (approved):
Nice very happy for this to go in!

PR #1975, opened 2025-03-10
Fix for dark mode - media queries for toolbar on mobile;

cityremade:
Quick fix for v4.13 before colour scheme is unified.
Dark mode stylesheet gets media queries to make map toolbar background transparent and buttons and links dark.


dbauszus-glx (changes requested):
This should be in the default view as the #mapbutton element is only relevant to the default view. And the mobile display [width] is specific to the default view as well.

cityremade:
yes, in theory. In practice we have redundant rules for dark mode. this dark mode stylesheet is supposed to go once colours are unified.

dbauszus-glx (approved):

RobAndrewHurst (approved):

PR #1986, opened 2025-03-11
Remove cloudinary provider module

dbauszus-glx:
The cloudinary provider was used pre signer.

The cloudinary provider module isn't even imported into the provider module.


RobAndrewHurst (approved):
LGTM!

Issue #1719, opened 2024-11-18
Ability to turn `draw` methods without the drawer panel


simon-leech:
**Issue Template**

## Description

It should be possible to avoid returning a `drawer` when you only have a single `draw` method defined. In the config below we should allow removing the panel. 
e.g. 

``` json
"draw": {
 "point": true
}
```

Maybe it could be something like this to signify to ignore the panel. 
If you had multiple draw methods then you would just return the buttons rather than being within a panel.
``` json
"draw": {
 "point": true,
 "BtnOnly": true
}
```

Open for discussion this one as need to be clear on requirements before we start it!

## Type of Issue 
Please delete options that are not relevant, and select all options that apply. 

- ✅ New Feature

dbauszus-glx:
This should be more general for all layer view panel. Whether to draw the panel content in a drawer or just add the content to the layer view.

dbauszus-glx:
I linked a draft PR for the drawer being created for circle drawing config and draw panel.

@simon-leech The flag cannot be BtnOnly since the content may be anything, even more drawer.

Better suggestions in regards to `contentOnly` flag key are welcome.

AlexanderGeere:
As per previous conversations I have changed it to drawer false, it applies to all panels or situations where a drawer is called in the framework. I have also ammended it to include the header of the element. 

So a group thats has group_drawer: false and a layer that has drawer: false would look like this:

![Image](https://github.com/user-attachments/assets/4d117a4a-c150-473d-8318-91600a7adbf7)

in the circle example specifying 
```JSON
   "draw": {
    "circle": {
      "drawer": false
     },
     "point": true
    }
```

would lead to:

![Image](https://github.com/user-attachments/assets/616aef00-88df-4acc-86e8-71159fad72fa)

PR #1992, opened 2025-03-12
Control drawer creation with flag

dbauszus-glx:
### Description
Whenever a drawer is created it is possible to supress the drawer creation in the config.

This is done in most instances by supplying `drawer: false` as in the below example. 

For example the draw config for a layer may include a circle config. It should be possible to flag for both config blocks to supress the creation of the drawer with the contentOnly flag.
```json
"draw": {
  "drawer": false,
  "point": {
    "label": "Add Point"
  },
  "line": {
    "label": "Add Line Yeah"
  },
  "polygon": {
    "label": "Add Polygon Now"
  },
  "rectangle": {
    "label": "Add Rectangle Here"
  },
  "circle_2pt": {
    "label": "Add Circle 2pt Go",
    "tooltip": {
      "metric": "area",
      "units": "meter",
      "convertTo": "yard"
    }
  },
  "circle": {
    "drawer": false
    "label": "Add Circle On Me"
  }
},
```

image

For layers:
Layers and entries can be in groups so to disable the group drawer `group_drawer: false` can be in any member of the group.

For Locations:
To disable the location view drawer for a layer `location_drawer: false` should be passed leading to:
Screenshot 2025-04-25 110445

### Issue
[1719](https://github.com/GEOLYTIX/xyz/issues/1719)

### Testing
This can be tested in `bugs_testing/drawer/workspace.json`.


AlexanderGeere (commented):
This is still working how i expect it to after the merge conflicts were resolved.

simon-leech (changes requested):
Love it - such a nice feature to reduce drawer numbers! 👏 

I pushed a minor change to change `group_drawer` to `groupDrawer` as we already use `groupClassList` so we should match the case. 

One thing i noticed, 
if you remove the group from a set of grouped layers, the styling looks different for the layer headers. I think maybe this is due to CSS nesting on the drawer > header ?

dbauszus-glx:
I remove the layer.location_drawer. This should be handled on the application level. We don't want location from one layer to look different than location from a different layer. This could quickly become messy.

I also remove the groupDrawer as this is counter intuitive. A group by definition is a drawer having. If the group is not desired it shouldn't be set rather than being set and then unset.

dbauszus-glx:
@simon-leech The groupDrawer logic has been removed as this is too complex and not the intention of this PR.

The intention is to prevent drawer which are automatically created to be not created.

A group is an optional element which is only created if configured; There should not be an option for an option to not show.

simon-leech (approved):
All good with me - I really like this 👏

RobAndrewHurst (approved):

PR #1999, opened 2025-03-13
_default view - restored search icon when location tab is inactive;

cityremade:
Added css rule for active locations button - on load will be search for gazetteer.
When active will change to location pin.


dbauszus-glx (approved):
lgtm

RobAndrewHurst (approved):

PR #2003, opened 2025-03-13
Map toolbar button text - css alignment

cityremade:
Buttons and hyperlinks inside map button get centered text;
In media queries for mobile material symbol line height set to `normal`.


AlexanderGeere (approved):

dbauszus-glx (approved):
I added the relevant css to the _toolbars where the .btn-column class is defined.

The btn-column class may be defined in an application view without the mapButton id.

The center of the text [icon] should always be applied and not just on mobile view restriction. Just because the issue is more prominent in a mobile view doesn't mean it shouldn't be addressed more  generally.

I removed .btn-column specificity from the _button css. Having this twice in _toolbars, and for the id in _map_toolbars is enough.

This is exactly the reason why I would have prefered to have the id mapButton in the view css which has the ID and not in the general css sheets which are applied to views which either do not have the id element or have the id assigned to a different element.

PR #2004, opened 2025-03-14
Warn in express module if node version is below 22.

dbauszus-glx:
image

Just a warning in the express script if the node version [on local] is below the required 22.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Dynamic module imports reduce the need to bundle 3rd party libraries such as [Ta
 
 ## XYZ Dependencies
 
-Node.js v18+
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
I also updated the readme dependencies.

image

RobAndrewHurst:
do we maybe not also want this warning to be on vercel too?

AlexanderGeere (approved):

dbauszus-glx:
> do we maybe not also want this warning to be on vercel too?

We would have to move the warning to the api module to checked and logged on every API call if we want this for serverless deployments which do not use express.

cityremade (approved):

PR #2005, opened 2025-03-14
Remove simple-statistics from dependencies and provide dynamic import

dbauszus-glx:
This PR removes the simple-statistics dependency.

An esmImport utility module is added.

An simpleStatistics utility module is added.

The simpleStatistics method will return an import promise from the esmImport module method.

The simpleStatistics import must be awaited to call the jenks method from the imported module in the jenks featureFields module method.

The featureFields process method which calls the jenks distribution method must be async in order to await the jenks method which await the import of the simple-statistics module from esm.


AlexanderGeere:
Are we going to address the biome issues in this PR?

AlexanderGeere (approved):
Looks all good to me.

simon-leech (approved):
Working all good for me - tested in default MAPP and Reports

RobAndrewHurst (approved):

Issue #1971, opened 2025-03-10
Material icons - prevent delay in loading the stylesheet.


cityremade:
At the moment Material Symbols are imported inside ui stylesheet which may result in delay in rendering the symbols.

Proposal: take out material symbols stylesheet and load before the framework stylesheets.

Here's the guide:
https://developers.google.com/fonts/docs/material_symbols

cityremade:
atm we have this in import:
https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200

100...700 is probably font weight range where we use 300 only as of now so this could be restricted.

I don't think restricting icon names is a good idea as we will need pr to include more.

RobAndrewHurst:
<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/61992098-7d18-452a-8c50-004cdb8fdc9f" />

currently its 3.3mb

dbauszus-glx:
@RobAndrewHurst Can you limit the axis `opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200` to see how this affects the size.

Starting with grad @0 which is the default that we use.

Weight [wght] @300.

Size [opsz] @24.

Do we use fill at all or can this be limited to 0?

cityremade:
we don't use fill

RobAndrewHurst:
```javascript
@import url(https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,300,0..1,0);
```

dbauszus-glx:
@RobAndrewHurst And with fill 0 as well? eg. `@import url(https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,300,0,0);`

dbauszus-glx:
Looks like 293kb if I read the content-length correctly.

![Image](https://github.com/user-attachments/assets/c35e2a85-fc47-4e35-8ee3-dc209742693b)

RobAndrewHurst:
<img width="890" alt="Image" src="https://github.com/user-attachments/assets/61584a03-48ae-4337-8fe9-92630160c8d9" />

@dbauszus-glx Yeah that's about right

dbauszus-glx:
Sounds very reasonable in exchange for not having svg icons in the repo.

PR #2011, opened 2025-03-18
Align MapButton panel CSS with mobile view

simon-leech:
The mobile view and larger screen view had different `mapButton` styling. 
The mobile view looked nicer as it had buttons that were individual not a panel that could contain just one or two things., yet still spanned the whole screen space.

**Changes**
1. mapButton has no background colour
2. individual buttons have background colour
3. Dark Mode fixes for this too

image


dbauszus-glx:
I don't understand the issue this PR addresses and the #mapButton / .btn-column style config is all over the shop in markup and different stylesheets. Is there a ticket which outlines the issues addressed in this PR?

simon-leech:
@dbauszus-glx Linked the issue to the PR, sorry should have done that earlier! 

I don't understand the import comment, I didn't change the material icons import.
All I changed was the CSS to move the styling from the mobile view only to be applied across both views.

dbauszus-glx (changes requested):
The import statement is changed in this PR to import all styles for all icons which results in a much larger import which delays the styling being applied. This is explained in #1971 

image

dbauszus-glx:
Also the screenshots in the PR do not look like that the button icons are centered which was addressed here. https://github.com/GEOLYTIX/xyz/pull/2003

simon-leech:
@dbauszus-glx  Corrected the issue with the PR - the changes are now all updated against latest minor and only the changes for this PR should be seen ! :)

RobAndrewHurst (approved):
Happy with this, we are probably going to want to rethink this long line of icons anyways.
But that's for another time.

dbauszus-glx (changes requested):
Without the background the label should be alligned to the edge of the mapview.

This looks weird.
image

Either the .btn-column class or #mapButton specificity must go. I'd say to keep the class rather than the id. Otherwise the class would be pointless because we want to add items with the same class to multiple mapviews in an application.

Class specificity will be overwritten by id specificity!

image

dbauszus-glx:
There is still a lot of override... I have now renamed to #mapButton specificity to .map-button class specificty. This allows you to see that same css is set twice in different [style] sheets.

Having assign id style rules on top of class style rules in the same compiled sheet is bad practice.

image

dbauszus-glx:
I have now removed the _map_toolbar as it confusing that the same element is defined twice and depending on the import of the sheet is defined which rules is applied first.

A sheet with the name toolbars should either contain all the toolbars or there should be individual sheets for each type of toolbar.

No more round robbin in the assignment of style rules. 
image

dbauszus-glx (approved):

PR #2012, opened 2025-03-18
Create assignTargetFromString method to reduce complexity

dbauszus-glx:
The dataview entry method has been cleaned up with the chkbox element logic being moved to the ui.Dataview decorator method.

The required chkbox element method has now been documented.

The [string] target property will now be consolidated by the assignTargetFromString method in the dataview entry module.


RobAndrewHurst (approved):
Nice, reducing the complexity and passes the current Dataview test.

AlexanderGeere (approved):
Working as expected, checked in a report as well.

PR #2013, opened 2025-03-18
Conditional field template lookup in query module

dbauszus-glx:
Only use template lookup for field value if template is defined in infoj entry.


simon-leech (approved):
Working for me :)

RobAndrewHurst (approved):

PR #2018, opened 2025-03-19
Calling API with parameters in square brackets 

### LINKED ISSUE #2017: match is not defined Error in API Query Parsing
When making a request to the API with query parameters like the following:
``` js
/api/query?template=whatever&ids=[1,2,3]
```
The server throws an error:
``` js
ReferenceError: match is not defined
```

This bug affects query parameters that use array-like values enclosed in square brackets.
The fix should ensure that match(...) is always called on a string (params[key]).

**Steps to Reproduce**
Pass a query string that includes an array inside square brackets (e.g., ids=[1480000656,1480000655]).
The API tries to parse the params[key] value but fails when calling match(...) without a string context.

simon-leech:
Fixes #2017 by ensuring we call match on a string.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -342,7 +342,8 @@ function validateRequestParams(req) {
     // Check whether the params value begins and ends with square braces.
     if (params[key].match(/^\[.*\]$/)) {
       // Match the string between square brackets and split into an array with undefined array values filtered out.
-      params[key] = params[key].match(/^\[(.*)\]$/)[1]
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
Working as expected with this test template.

```
"select_array": {
  "template": "select array[%{arr}]",
  "value_only": true
},
```

image

RobAndrewHurst (approved):

PR #2019, opened 2025-03-19
Reduce cloudinary entry module complexity

dbauszus-glx:
An imgOnload method for the image element onload method is created to reduce the complexity of the upload method which has been renamed to newImage.

The documentation has been updated.

The location_update query template called from the cloudinary module method incorrectly set the array template to be a string  `{a,b,c}`. The query module then attempts to parse this string as JSON which will throw an error exception.

Array values should be provided as array values for the `%{imagess}` field to be sent as SQL params.

The `entry.file ??= input.files[0]` assignment was deadly. The same file would be uploaded every time on the same images entry.

Changing the file in quick succession could potentially cause the previous fiule to be referenced in the onload method. This is much cleaner by passing the file as an aditional argument to the onload methods.

Dejunked CSS rules.

Instead of creating nested media print queries for classes in classes not to be displayed on mobile the dedicated `.mobile-display-none` class should be assigned to the element.

```css
@media print {
  .image,
  .images,
  .documents {
    .drag_and_drop_zone {
      display: none;
    }
  }
}
```

The `.images-grid` is only ever used within the cloudinary type entry. This should not sit in the `_locationview.css` sheet witth duplicate specificity to `_cloudinary.css` sheet.

The `.trash` specificity nested in the images-grid isn't used anywhere and can be removed.

With the input definition wasn't that good that it needed to be defined twice. If happening only in the context of the dropzone then it should be only defined in that context and well. Meaning that the input covers the whole of the area.

Replace self closing span with content after from css with a proper span element. **A span is never self closing!**

Delete/close is always top / right, not left/bottom.

The link-with-img is not that good that it needs to be defined multiple times.

```css
  .link-with-img {
    .delete {
      font-size: 1.5em;
    }
  }
```

The drag and drop didn't work and has meaningless prevent default garble. The view should be disabled after checking for a file. Not disable and return without a file never to be enable again since there is no upload.

None of the prevent default where needed and the weird uncommented spread did not work.

```js
      ondrop=${(e) => {
        e.preventDefault();
        upload(e, { ...entry, ...{ file: e.dataTransfer.files?.[0] } });
      }}
```

https://github.com/GEOLYTIX/xyz/pull/2019/commits/e001e5f6e279a463ba472bff6ab7f588b366e002


RobAndrewHurst:
https://github.com/user-attachments/assets/d57de0dc-af6b-4931-8e54-ea987f6c8e78

When dragging and dropping an image in cloudinary you I get two images being added to the cloudinary folder.

See the screen recording above.

RobAndrewHurst (changes requested):
See comment above!

dbauszus-glx:
@RobAndrewHurst 
I added the preventDefault with comments in regards to the input element onchange events being triggered when nested in an element with an ondrop event.
https://github.com/GEOLYTIX/xyz/pull/2019/commits/594cceceb2b7c202b0842efd99ed27e8148cecca

AlexanderGeere (approved):
I amended the cloudinary entry tests. The methods are working as expected, tested in a report as well.

RobAndrewHurst (approved):

simon-leech (changes requested):
Everything is working as expected for me - tested using single image and images in both MAPP and custom views and working as expected.

Only thing I noticed, and not sure if its in the scope of this PR but its related so let me know if you want to do it in another PR or this one. But if you have no `env.CLOUDINARY` key present, you don't get any warnings when saving. You just get an error 
image

RobAndrewHurst:
> Everything is working as expected for me - tested using single image and images in both MAPP and custom views and working as expected.
> 
> Only thing I noticed, and not sure if its in the scope of this PR but its related so let me know if you want to do it in another PR or this one. But if you have no `env.CLOUDINARY` key present, you don't get any warnings when saving. You just get an error image

This won't be returned to the front end but warns on the backend

simon-leech (approved):

PR #2020, opened 2025-03-19
Mock preload 📦

RobAndrewHurst:
I have implemented preload functionality into Codi 🐶 where we can define a
directory that will be executed/loaded before the actual tests begin.

This can be done by adding the `preload` property in the `codi.json` config
file in the root of xyz.

```json
{
  "excludeDirectories": ["lib", "browser", "_mapp.test.mjs"],
  "preload": "tests/__preload"
}
```

The current `preload` uses the directory `tests/__preload`. Currently this
implements the `fs` and the `provider/file` mocks in a global scope.

## Pros

The benefit of this is that you can now refer to the different mocks once
instead of redeclaring them all the time.

## Cons

I found this isn't as easy to implement with modules that require keys at
runtime. So there is only certain circumstances in which we can use this. Other
alternatives will be thought of for other modules.


dbauszus-glx (changes requested):
I think the file name in the respective directory should either be the same as the file[module] being mocked or get a mock extension [eg __preload/mod/provider/file.mock.mjs].

I assume there will be multiple module which need to have mock files.

These shouldn't all be called fs.mjs in different folder.

image

RobAndrewHurst:
> I think the file name in the respective directory should either be the same as the file[module] being mocked or get a mock extension [eg __preload/mod/provider/file.mock.mjs].
> 
> I assume there will be multiple module which need to have mock files.
> 
> These shouldn't all be called fs.mjs in different folder.
> 
> image

Ah yes! Sorry I used the wrong file name for the file provider, and forgot to change it!

The mocks will replicate the structure as much as possible

dbauszus-glx (changes requested):
I still can't get this to work.

Even with the identical file.mjs and fs.mjs files in the test provider directory I get an error that the fileFn function is not defined.

image

RobAndrewHurst:
@dbauszus-glx I am not too sure what was happening but there were a bunch of commits that were out of sync. I did a force push, so please delete the branch and switch to it again

dbauszus-glx:
@RobAndrewHurst How can we complete the file.test.mjs ?

The file method will return json only if the provided file string ends in json eg `file('../../dir/tests/thing.json')`.

Without a file ending or any other ending the result returned should be string.

```js
await codi.describe(
  { name: 'file:', id: 'provider_file', parentId: 'provider' },
  async () => {
    const { default: file } = await import('../../../mod/provider/file.js');
    await codi.it(
      { name: 'Get File test', parentId: 'provider_file' },
      async () => {
        const fileContent = { text: 'I am a file' };

        readFileSyncFn.mock.mockImplementationOnce(() => {
          return JSON.stringify(fileContent);
        });

        const results = file('../../dir/tests/thing.json');

        codi.assertEqual(results, fileContent);
      },
    );
  },
);
```

RobAndrewHurst:
> @RobAndrewHurst How can we complete the file.test.mjs ?
> 
> The file method will return json only if the provided file string ends in json eg `file('../../dir/tests/thing.json')`.
> 
> Without a file ending or any other ending the result returned should be string.
> 
> ```js
> await codi.describe(
>   { name: 'file:', id: 'provider_file', parentId: 'provider' },
>   async () => {
>     const { default: file } = await import('../../../mod/provider/file.js');
>     await codi.it(
>       { name: 'Get File test', parentId: 'provider_file' },
>       async () => {
>         const fileContent = { text: 'I am a file' };
> 
>         readFileSyncFn.mock.mockImplementationOnce(() => {
>           return JSON.stringify(fileContent);
>         });
> 
>         const results = file('../../dir/tests/thing.json');
> 
>         codi.assertEqual(results, fileContent);
>       },
>     );
>   },
> );
> ```

I can add another test for this!

RobAndrewHurst:
@dbauszus-glx I have updated the file.test.mjs with a test with a file that doesn't have an extension.

dbauszus-glx:
@RobAndrewHurst Is the base provider test file expected to fail?

image

RobAndrewHurst:
> @RobAndrewHurst Is the base provider test file expected to fail?
> 
> image

No. this is a bug to do with codi on windows

RobAndrewHurst:
Closing this as it needs a different approach. 

thanks all for having a look

Issue #1896, opened 2025-02-11
Merge templates in [layer] object


dbauszus-glx:
A template can be defined as an object with a key in a layer.

The mergeTemplates method should retrieve this template from the workspace and merge the [layer] object into the template.

```json
      "scratch": {
        "format": "wkt",
        "table": "scratch",
        "qID": "id",
        "template": {
          "key": "layer_template",
          "src": "${RESOURCES}/bugs_testing/templates/scratch.json"
        }
      }
```

The src, _type, and the template object itself should be removed before the merged obj is returned from the workspace API.

dbauszus-glx:
This issue has already been resolved.

PR #2026, opened 2025-03-24
Nested locales

dbauszus-glx:
This PR introduces the nested locale logic and array keys. Locales are no longer required to be listed in the locales object of the workspace. A locale can be stored and accessed as a template. A nested locale or array locale is merged from multiple locales.

eg `locale=[europe,uk,london]` Will retrieve the locales, left to right, and deep merge each retrieved locale into its parent.

### locales / getNestedLocales

A nestedLocales method was added. The method is called from the [get] locales method if a locale parameter has been provided. The nestedLocales method will be returned instead.

```
/api/workspace/locales?locale=us
```

Any locale can be nested. The getNestedLocales interface is simply to provide a user interface for which locales are intended to be nested.

### locale [key array]

The [get] locale from workspace API has been extended to allow requesting multiple locales as nested. The locale property can be provided as an array.

```
/api/workspace/locale?locale=[us,brand_a_locale]&layers=true
```

The getLocale module method will return a locale with the array key which merges the locales left to right.

The locale workspace API provides the locale as optional parameter to the getLayer method to prevent having to call the getLocale module for every layer requested in the context of a [nested] locale.

## getLocale

The getLocale module method will call itself iteratively for nested locales with the parentLocale as optional parameter. The locale key is shifted from the locale array property to allow for locales defined in array to be merged.

## getLayer

A locale object can be provided to the getLayer method. If a locale is provided [eg from the getLocale] module method then it is not required to get the locale again from the getLocale method itself. This removes an additional round trip when requesting a locale with layers.

## getTemplate

The getTemplate method receives the template key as string parameter. The locale will always be an object in the context of the getTemplate module method. The requesting key is assigned to each template object within the getTemplate method.

## mergeTemplates

The check whether a template of the same name exists has been removed. Only the obj.template string property value will be used to lookup template keys in the workspace.templates.

## _default.html

A list of nested locales will be requested from the default view script if a locale has a locales array property. The nested locales returned from the locales workspace API will be spread into the locales array of locales available from the workspace.locales.


dbauszus-glx:
A test workspace would look like this:

```json
{
  "templates": {
    "query": {
      "roles": {}
    },
    "common_layer": {},
    "finance_layer": {
      "roles": {}
    },
    "brand_a_layer": {
      "roles": {}
    },
    "brand_b_layer": {
      "roles": {}
    },
    "brand_c_layer": {
      "roles": {}
    },
    "brand_a_locale": {
      "roles": {
        "brand_a": {}
      },
      "layers": {
        "brand_a_layer": {
          "template": "scratch_geojson"
        }
      }
    },
    "brand_b_locale": {
      "roles": {
        "brand_b": {}
      },
      "layers": {
        "brand_b_layer": {
          "template": "scratch_geojson"
        }
      }
    },
    "brand_c_locale": {
      "roles": {
        "brand_c": {}
      },
      "layers": {
        "brand_c_layer": {}
      }
    },
    "UK_locale": {
      "roles": {
        "UK": {}
      },
      "extent": {},
      "layers": {
        "brand_c_layer": {}
      }
    },
    "OSM": {
      "display": true,
      "format": "tiles",
      "attribution": {
      }
    },
    "scratch_geojson": {
      "format": "geojson",
      "dbs": "NEON",
      "table": "scratch",
      "srid": "3857",
      "geom": "geom_3857",
      "qID": "id",
      "style": {
        "highlight": {
          "scale": 1.3,
          "strokeColor": "#090"
        }
      },
      "infoj": [
        {
          "type": "geometry",
          "display": true,
          "field": "geom_3857",
          "fieldfx": "ST_asGeoJSON(geom_3857)"
        },
        {
          "type": "pin",
          "label": "ST_PointOnSurface",
          "field": "pin",
          "fieldfx": "ARRAY[ST_X(ST_PointOnSurface(geom_3857)),ST_Y(ST_PointOnSurface(geom_3857))]"
        },
        {
          "title": "ID",
          "field": "id",
          "inline": true
        }
      ]
    }
  },
  "locale": {
    "layers": {
      "OSM": {
        "template": "OSM"
      }
    }
  },
  "locales": {
    "us": {
      "_roles": {
        "us": {}
      },
      "extent": {},
      "layers": {
      },
      "locales": [
        "brand_a_locale",
        "brand_b_locale"
      ]
    },
    "europe": {
      "_roles": {
        "europe": {}
      },
      "extent": {},
      "layers": {
      },
      "locales": [
        "UK_locale",
        "brand_b_locale",
        "brand_c_locale"
      ]
    }
  }
}
```

mod/workspace/getLocale.js
```diff
@@ -25,34 +30,51 @@ The getLocale method requests the workspace from cache and checks whether the re
...
-  let locale = Object.hasOwn(workspace.locales, params.locale)
-    ? workspace.locales[params.locale]
-    : workspace.locale;
+  const localeKey = Array.isArray(params.locale)
+    ? params.locale.shift()
+    : params.locale;
+
+  let locale;
+
+  if (!localeKey) {
```
RobAndrewHurst, line 71: What happens if we provide a locale that doesn't exist?

RobAndrewHurst: @dbauszus-glx I have added in a failing test where I provide locales that do not exist to the _workspace.js module.

How do we want this to be handled?

RobAndrewHurst (changes requested):

mod/workspace/getLocale.js
dbauszus-glx: The error message should be returned. eg.
image

AlexanderGeere (changes requested):
I noticed a this error on an instance i setup that is very similar to how some client instances currently work:

Screenshot 2025-04-11 112945

It has a lot of locales and a bit of complicated permission structure. I'm not entirely sure whats causing it, you can replicate it on `bugs_testing\locales\workspace.json`.

This works fine on current patch.

dbauszus-glx:
A template object may be defined inside a template [eg layer]. This template will be added to the workspace.templates.

A template property which is not a string must not be gotten from the getTemplates method to prefent a circular reference issue in the deep merge.

A template which errs must return.

simon-leech (changes requested):
@dbauszus-glx  Please can you resolve the merge conflict on this PR with the _default.html view?

dbauszus-glx:
> @dbauszus-glx Please can you resolve the merge conflict on this PR with the _default.html view?

The conflicts have been removed.

AlexanderGeere (changes requested):
I have noticed two things:
1. Template errors are not coming through. on `bugs_testing/locales/workspace.json` i see this error, on current patch, in console 
```Javascript 
Array [ "ENOENT: no such file or directory, open 'C:\\Users\\Admin\\Documents\\GEO\\xyz_resources\\bugs_testing\\site.json'" ] 
``` 
I don't see this on the PR.
 
2. Upon fixing this the sites layer still doesnt show in the layers panel, configured as such
``` JSON
      "sites": {
        "table": "public.scratch_sites",
        "template": {
          "key": "assess_my_site",
          "src": "${CLIENT}/site.json"
        },
```

RobAndrewHurst:
> I have noticed two things:
> 
> 1. Template errors are not coming through. on `bugs_testing/locales/workspace.json` i see this error, on current patch, in console
> 
> ```js
> Array [ "ENOENT: no such file or directory, open 'C:\\Users\\Admin\\Documents\\GEO\\xyz_resources\\bugs_testing\\site.json'" ] 
> ```
> 
> I don't see this on the PR.
> 
> 2. Upon fixing this the sites layer still doesnt show in the layers panel, configured as such
> 
> ```json
>       "sites": {
>         "table": "public.scratch_sites",
>         "template": {
>           "key": "assess_my_site",
>           "src": "${CLIENT}/site.json"
>         },
> ```

Can you please write a test for this?

dbauszus-glx:
@AlexanderGeere This is issue was first outlined here: https://github.com/GEOLYTIX/xyz/issues/1896

I have now addressed this in the last commit.

dbauszus-glx:
@RobAndrewHurst We need to work out tests which request workspace objects with different user roles from the same cached workspace.

dbauszus-glx:
I am not sure how feasible it is to write tests for everything locale/layer templates / roles for this PR. I can see this could be taking weeks dragging the implementation of this PR into the far future.

I think we should focus on layer template tests first. I creates a WIKI page for this. https://github.com/GEOLYTIX/xyz/wiki/Templates-%5BWorkspace%5D

```json
{
  "templates": {
    "OSM_tile_layer": {
      "format": "tiles",
      "URI": "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png"
    },
    "OSM_tile_layer_src": {
      "src": "${RESOURCES}/templates/osm_layer.json"
    },
    "osm_attribution": {
      "attribution": {
      }
    },
    "osm_bw_style": {
      "style": {
        "contextFilter": "grayscale(100%)"
      }
    }
  },
  "locale": {
    "layer": {
      "attribution": {
        "repo": "https://github.com/GEOLYTIX/xyz"
      }
    },
    "layers": {
      "OSM": {
        "display": true,
        "name": "OSM",
        "format": "tiles",
        "URI": "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png"
      }
    }
  }
}
```

The next test would be locale and locales tests. I will add to the page once the layer tests are ready to get some feedback how feasible these tests are.

Roles tests will be a magnitude more complicated and probably require their own wiki page which I am planning to create next week.

dbauszus-glx:
I merged the patch which had a conflict in regards to the locale.layer merge.

RobAndrewHurst (approved):

dbauszus-glx:
The src property will now be removed when a template is fetched from src to enable caching. The src property will not be removed if the src starts with `file:`, hence this resources will always be fetched from their src which is immediate for the file system.

RobAndrewHurst (changes requested):
I noticed that if I edit my workspace locally and then refresh the browser I don't see changes.

I thought all local resources would not be cached when running  locally?

RobAndrewHurst:
> @RobAndrewHurst This is only for templates for the time being not the worksapce itself.

Thanks for the clarification

RobAndrewHurst (approved):
Looks good

simon-leech (changes requested):
Review underway - please don't merge :)

simon-leech:
@dbauszus-glx  Please can you have a look at my latest commit. I fixed a few spelling mistakes in the documentation  and added to TODO comments in `getLayer.js` and `getLocale.js` if you could please add the reason for why we delete all the property keys just for future reference :)

cityremade (approved):

dbauszus-glx:
@simon-leech works for me.

I'd need to see more of the configuration to assess what's going wrong.

image

image

simon-leech (approved):
Ran through a scenario and understand how to set this up! :)

PR #2028, opened 2025-03-25
uhtml static module

RobAndrewHurst:
# uhtml static module

I have added in the esm uhtml v3.1.1 module into the `lib/utils`.

I have removed the npm package and added this into the ignore for biome.


dbauszus-glx:
@RobAndrewHurst The esmImport utility is only meant for dynamic imports which are not always required.

uhtml is always required. If esm is down or unavailable we would not be able to use the mapp library at all.

I am worried that the top level uhtml import from esm is not conditional. It is not possible to provide the module through a link like we do with openlayers which is always required.

It should be possible to compile the mapp library with uhtml. The uhtml methods can be compiled from a module file in the public directory.

RobAndrewHurst:
@de

> @RobAndrewHurst The esmImport utility is only meant for dynamic imports which are not always required.
> 
> uhtml is always required. If esm is down or unavailable we would not be able to use the mapp library at all.
> 
> I am worried that the top level uhtml import from esm is not conditional. It is not possible to provide the module through a link like we do with openlayers which is always required.
> 
> It should be possible to compile the mapp library with uhtml. The uhtml methods can be compiled from a module file in the public directory.

@dbauszus-glx Happy for this!

dbauszus-glx (approved):

PR #2040, opened 2025-04-01
mailer username auth & servername

RobAndrewHurst:
> [!NOTE]
> I have added in two keys after looking at the nodemailer docs to ensure a bit
> more control of different params required for the createTransport function call.

```bash
TRANSPORT_USERNAME: username
TRANSPORT_NAME: server_name
```

The reason for the `TRANSPORT_USERNAME` introduction is that not all providers
authenticate with an email address but also a normal username.

I also added in the `TRANSPORT_NAME` so that we can control the optional
hostname of the client, used for identifying to the server. This will default
to the local part or username of the `TRANSPORT_EMAIL` variable.

Additional changes I made were to clean up the docs.


AlexanderGeere (approved):
I added some extra checks. If you supplied `TRANSPORT_USERNAME` and not `TRANSPORT_EMAIL` the function would just return. If you don't supply `TRANSPORT_NAME` or `TRANSPORT_EMAIL` it would crash. The checks I added should fix these.

RobAndrewHurst:
> I added some extra checks. If you supplied `TRANSPORT_USERNAME` and not `TRANSPORT_EMAIL` the function would just return. If you don't supply `TRANSPORT_NAME` or `TRANSPORT_EMAIL` it would crash. The checks I added should fix these.

So `TRANSPORT_EMAIL` is not optional as it's needed to send the email because its used as the from property.

AlexanderGeere:
> I added some extra checks. If you supplied `TRANSPORT_USERNAME` and not `TRANSPORT_EMAIL` the function would just return. If you don't supply `TRANSPORT_NAME` or `TRANSPORT_EMAIL` it would crash. The checks I added should fix these.

These are not necesary as emai is still required.

simon-leech (approved):

dbauszus-glx (approved):
I have updated the TRANSPORT env check.

This shouldn't be a condition for import to warn and crash.

https://github.com/GEOLYTIX/xyz/pull/2040/commits/0cd8ab3ba9006d406d4e1edb9d71bbbe136b7d09

PR #2045, opened 2025-04-02
check entry.value type in date module

dbauszus-glx:
The temporal.date and temporal.datetime method should receive an integer as specified in the documentation.

The params [entry] object for the newValue check is unnecessary.

Date and datetime entry fields will work with string fields but an integer stored in a string field can not work with date[time] range filter.

This PR adds a check on the timestamp type which is either the entry.newValue or entry.value property in the entry.edit condition.

### DIFF
--- a/lib/ui/locations/entries/date.mjs
+++ b/lib/ui/locations/entries/date.mjs
@@ -43,7 +43,7 @@ function date(entry) {
       console.warn(`${entry.type} field: ${entry.field} should be an integer.`);
     }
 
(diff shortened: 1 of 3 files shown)


cityremade (approved):
ok so now unix timestamp is parsed inside the date entry, not temporal functions.
This fixes my problem.

AlexanderGeere (approved):

RobAndrewHurst (changes requested):
image

These failing tests need to be addressed first before merging in

RobAndrewHurst (approved):

PR #2046, opened 2025-04-03
Svg icons cleanup

cityremade:
Retiring all svg icons as they're no longer in use.


dbauszus-glx (approved):
I replaced the remaining svg data url and removed the svg data url loader from the css build command.

RobAndrewHurst (approved):

Issue #1620, opened 2024-10-29
the mighty `z-index` clean up 🧼 


RobAndrewHurst:
## Description

The `z-index` css property is used and abused in the current codebase trying to get different elements to fit in different contexts. We can use the `isolation` property to correct these based on the context of where the elements are being used. 

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

```css
/* Keyword values */
isolation: auto;
isolation: isolate;

/* Global values */
isolation: inherit;
isolation: initial;
isolation: revert;
isolation: revert-layer;
isolation: unset;

```

We want to avoid using z-index values of `99999` as I don't know what that magic number is doing compared to `999998`. It's very hard to read and extremely hard to manage different elements.

cityremade:
fixed by linked pr.

PR #2047, opened 2025-04-03
z-index eradication;

cityremade:
Multiple figure z-index replaced with isolation property. 

isolation: isolate for elements that should come up on top.

Image preview (.interface-mask) now in its own stylesheet.

Removed default focus outline from dialogs to prevent browser/system default styling.

https://github.com/GEOLYTIX/xyz/issues/1620


RobAndrewHurst (changes requested):
image

There is an issue with drop downs and other elements.

They seem to go behind other elements. 

I have noticed this with numeric fields & other drop downs.

image

cityremade:
Dropdown and searchbox now have z-index set at 4 so that they have a higher rank in stacking context than the slider.
Slider has z-index as of now as it is part of the logic how the element was designed.
Slider element may need review some time soon as the marking dot is an svg background with fixed orange colour in it.

RobAndrewHurst (changes requested):
https://github.com/user-attachments/assets/56212d96-400e-4638-ac28-805ac702660a

@cityremade Nice! The dropdown and a-b slider work better now. 

Dropdowns with other dropdowns seem to be an issue. Where the header of a dropdown seems to appears above the dropdown elements

Let me know if the video isn't clear enough

cityremade:
Updates:

position: relative removed from class dropdown. This caused new stacking context and nested `ul` appeared underneath next container.
`.dropdown > ul` gets z-index at 4 to come on top slider elements.
`.dropdown` container is now a **div**, no longer a button as button shouldn't contain any block html elements. This changes disabled behaviour as `.dropdown.disabled` is now a class, not property
pills - fix for styling, close button was previously invisible.

lib/ui/elements/dropdown.mjs
```diff
@@ -65,7 +65,7 @@ export default (params) => {
 
   params.node = mapp.utils.html.node`
     ${params.pills?.container}
-    <div
+    <button
```
RobAndrewHurst, line 68: Why are either of these even a thing?

Shouldn't this be a select tag?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

RobAndrewHurst (changes requested):

AlexanderGeere (approved):
Tested a few dialogs and some custom views. all looks well to me.

dbauszus-glx:
Should the dialog be on top of these buttons?

image

jfitzpatrick17:
Do we know why the multi-filter dialog positions to the left of the window like that sometimes? (Should be on the right of the map view). I occasionally get this happening too and have to drag it back.

dbauszus-glx:
@jfitzpatrick17 Can you add a seperate ticket for the dialog position?

dbauszus-glx (approved):

simon-leech (approved):

PR #2048, opened 2025-04-03
CSS colour classes - cleanup;

cityremade:
**Removed classes:**
- light-background ➡️  unused
- light-colour ➡️  unused
- off-black-background ➡️  unused
- primary-background ➡️  unused
- primary-colour 
⬆️
This styling is now inherited by button component.
This class has been used with buttons.

- off-black ➡️ Unused. **_Plugins use only the colour variable which can have a fallback colour if necessary._**

**Renamed classes:**

no-color ➡️ danger
ok-color ➡️ info

**Changed classes**

* on
⬆️
Eradicated .on class which was used in 2 cases and was impossible to find.
Now:
on for display toggle is toggle-on
on for location edit is edit-toggle-on

In plugins on-colour was used only once and not in line with new colour scheme so no loss here.

Also: text input and textarea text colour now basic. Otherwise it's a spiral of making everything 'primary' and therefore at risk of poor readability.

### DIFF
--- a/lib/plugins/login.mjs
+++ b/lib/plugins/login.mjs
@@ -29,7 +29,7 @@ export function login(plugin, mapview) {
   const iconName = `${mapp.user ? 'logout' : 'lock_open'}`;
 
   const iconClass =
-    'material-symbols-outlined' + (mapp.user ? ' no-colour' : '');
+    'material-symbols-outlined' + (mapp.user ? ' danger' : '');
 
   btnColumn.appendChild(mapp.utils.html.node`
     <a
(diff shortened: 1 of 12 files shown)


dbauszus-glx (changes requested):
The danger class which defines the color: danger variable should be called color-danger to be clear what it does. With that you don't need the .danger class in the _material-icons.css.

https://github.com/GEOLYTIX/xyz/pull/2048/commits/c6067b994274d6f6975c8d699972dd4d31699a19

The same should be done for the info reducing the required definition of the color variable in classes in the style sheets.

cityremade:
Color classes in _colours can be applied to text within framework components but they can't override material symbol colour therefore we need them in both stylesheets.
I'd rather question if they are needed in _colours - I'd rather bind colours as private properties to specific ui components.

RobAndrewHurst (approved):

AlexanderGeere (approved):

simon-leech (approved):

PR #2049, opened 2025-04-09
add a user filter to the sqlfilter module

dbauszus-glx:
A default filter can add SQL filter statement for the user.email of the request.

```json
"filter": {
  "default": {
    "char_field": {
      "user": true
    }
  }
}
```

The default filter can not be modified by the client.

The documentation for the sqlfilter module has been updated with typedef for filterTypes and typeCheckers objects.

The deeply nested array methods in the mapFilterEntries method has been replaced with an easier to read `for (const [field, value] of Object.entries(filter))` loop.

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -182,10 +182,10 @@ async function layerQuery(req, res) {
   // Create filter condition for SQL query.
   req.params.filter = [
     (req.params.layer.filter?.default &&
-      `AND ${sqlFilter(req.params.layer.filter.default, req.params.SQL)}`) ||
+      `AND ${sqlFilter(req.params.layer.filter.default, req)}`) ||
       '',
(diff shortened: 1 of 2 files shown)


simon-leech (commented):
This is working as expected for me when I apply it to the layer. 

One question - I wonder if we should change the 
`user`: `true`  to be `email`: `true`. 
This is as we use `email` in the `queryparams` module so it would be consistent to use the same parameter value for getting the email for either a filter or to be passed to a query?

dbauszus-glx:
Sure. I am happy to change this.

dbauszus-glx:
@simon-leech Should now work with email instead of user.

simon-leech (approved):

RobAndrewHurst (approved):

PR #2053, opened 2025-04-11
Do not debounce admin user requests

dbauszus-glx:
Admin user requests with a post body should not be debounced.


cityremade (approved):
tested locally no longer fails on frequent updates.

simon-leech (approved):
Working fine for me

Issue #1968, opened 2025-03-07
Ping utility


dbauszus-glx:
There should be a utility method which sends a ping request. A ping request will be repeated after a set time until fulfilled.

PR #2054, opened 2025-04-11
Ping Util

simon-leech:
# Ping Utility

## Description

  This function will ping the database running a specific query.
  The query will return a response, and this response will be checked against the successCondition and errorCondition if provided.
A `responseType` is provided to ensure the condition is evaluated properly. 
If the value of the response is the successCondition, the promise resolves.
If the value of the response is failCondition the promise rejects. 
If a callback is provided, this is completed.
If callback is set to false, the ping utility will return before evaluating the condition (ie stop). 

``` js
mapp.utils.ping({
  query: 'location_field_value',
  responseType: 'text',
  successCondition: 'ok',
  failCondition: 'failed',
  queryparams: {
    id: true,
    field: 'model_status',
    table: 'location_table',
  },
  interval: 4000,
  callback: (response) => {
    console.log(response);
  },
});
```
## GitHub Issue

[Provide the link to the relevant GitHub issue it addresses.](https://github.com/GEOLYTIX/xyz/issues/1968)


dbauszus-glx:
@simon-leech 

params.query is required. The method should shortcircuit with a warning if the query property is undefined.

params.field and any property required for the query should be added to queryparams.

I do not understand the stages. The ping should terminate with an error thrown if the ping query errs.

I would just provide a property success condition. The response is compared to the success condition value. If the successCondition matches the response then the callback is called with the response as param.

cityremade:
Most of the time we expect the status information to be updated by the model procedure. 
However at times it might be needed to update the field based on status value returned from ping. 
Could it be included here? updating the field based on returned status?
Or could it be a separate utility?

simon-leech:
> Most of the time we expect the status information to be updated by the model procedure. However at times it might be needed to update the field based on status value returned from ping. Could it be included here? updating the field based on returned status? Or could it be a separate utility?

@cityremade I think this should be a separate utility if we think it's widely used enough to facilitate this.
The callback in the ping utility can return you the status, so you'll have access to it once the ping completes. So you could then pass this to a query for example in a custom plugin.

cityremade:
Screenshot 2025-04-22 at 16 27 23

I cannot get this working with a button where a click would ping a table. I am getting the above if I use my own query or the default one.

dbauszus-glx:
The ping utility method does not require to be async and return a method.

The async pingQuery method will be called in a timeout with the provided interval property.

The method will call the callback which can be used to cancel the execution.

The params object is provided to the XHR utility which automatically cancel a running query.

I created a simple ping entry method which shows how to cancel the ping execution when a location is removed.

tests/lib/utils/ping.test.mjs
```diff
@@ -1,273 +0,0 @@
-//Assigning mapp.utils.xhr to a property to restore original function with.
```
RobAndrewHurst: Why were these tests removed?

dbauszus-glx: Because they didn't test the ping utility logic.

AlexanderGeere (approved):
This is working as expected.

RobAndrewHurst (approved):
All happy with this!

simon-leech:
I can't review this as its my PR - but i'm happy with the changes that have been made so all good to go! 🥳

PR #2056, opened 2025-04-14
user select property for some elements;

cityremade:
user-select: none now set for these elements:

button
hyperlink
drawer header
dialog
checkbox
.legend
.meta (inside layer view)
.link-with-img

any more?

### DIFF
--- a/public/css/base/_button.css
+++ b/public/css/base/_button.css
@@ -3,8 +3,12 @@ button {
   outline: none;
   background: none;
   text-align: center;
-  user-select: none;
   color: var(--color-primary);
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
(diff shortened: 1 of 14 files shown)


simon-leech (approved):
This seems fine to me - tested and working as expected.

AlexanderGeere (commented):
Seems it is a firefox quirk, not sure if want to look at fixing it on firefox. Adding user-select: none to .drawer does fix it on firefox, so its maybe an engine difference in the inheritance of the property. For safari we could add -webkit-user-select as well.

AlexanderGeere (commented):
Seems it is a firefox quirk, not sure if want to look at fixing it on firefox. Adding user-select: none to .drawer does fix it on firefox, so its maybe an engine difference in the inheritance of the property.

dbauszus-glx (approved):
working for me as expected.

PR #2059, opened 2025-04-14
Remove vector_layer entry module

dbauszus-glx:
The `legacy_layer` entry method is handling `mvt_clone` and `vector_layer` entry configurations.

The featureData method in the layer entry module handles the featureSet or featureLookup entries.

Otherwise the data is assumed to be features for the setSource layer method.


AlexanderGeere (approved):
Tests all pass, checked some custom views as well. Looks good to me.

cityremade (approved):

PR #2060, opened 2025-04-15
The getLayer module method must merge the locale.layer object

dbauszus-glx:
The getLayer module method should merge the layer object into the default locale.layer object if this exists.

### DIFF
--- a/mod/workspace/getLayer.js
+++ b/mod/workspace/getLayer.js
@@ -3,6 +3,7 @@
 The getLayer module exports the getLayer method which is required by the query and workspace modules.
 
 @requires /utils/roles
+@requires /utils/merge
 @requires /workspace/mergeTemplates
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Tested using a locale.layer and merging onto layers and working correctly using a `default` filter. 
Then applied a filter on a field on this layer and it was correctly only returning the min/max including that filter applied. 

Also tested using roles as follows as working as expected. 
``` json
"roles": {
"Brazil": {
  "layer": {
    "filter": {
      "default": {
        "ctrycd": {
          "match": "BRA"
        }
      }
    }
  }
},
"Paraguay": {
  "layer": {
    "filter": {
      "default": {
        "ctrycd": {
          "match": "PRY"
        }
      }
    }
  }
}
```

PR #2064, opened 2025-04-15
Fix saml import logic

RobAndrewHurst:
To resolve this issue, I made two key changes for correct SAML usage:

1. Used destructuring to import the `SAML` class directly from the module.
2. Ensured the `getModule` function is properly awaited.

```diff
- const SAML = await import("@node-saml/node-saml");
+ const { SAML } = await import("@node-saml/node-saml");

// ...rest of the SAML config

samlStrat = new SAML(samlConfig);
```

```diff
- const exportedModule = getModule();
+ const exportedModule = await getModule();
```

These changes ensure that `SAML` is correctly imported as a class and that asynchronous module loading works as expected.

### DIFF
--- a/mod/user/saml.js
+++ b/mod/user/saml.js
@@ -165,7 +165,6 @@ const getModule = async () => {
 The SAML service has not been configured correctly.
 */
 function saml_not_configured(req, res) {
-
   res.status(405).send('SAML not configured');
 }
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
All good with me - thanks @RobAndrewHurst !

dbauszus-glx (changes requested):
The saml/metadata request fails for me.

This should return an error if configured poorly.

image

dbauszus-glx:
I create a method to be returned if the SAML init fails.

This is a 405. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405

image

dbauszus-glx (approved):

PR #2065, opened 2025-04-15
pill - inner close button always contrasting.

cityremade:
Fix for pill close button which should always get contrasting font colour.
This should fix pill problem in admin panel where close cross colour is set to primary without mapp css stylesheet.

### DIFF
--- a/public/css/base/_material_icons.css
+++ b/public/css/base/_material_icons.css
@@ -28,7 +28,6 @@
 
   &.close::after {
     content: 'close';
-    color: var(--color-font-mid);
   }
 }
 
--- a/public/css/elements/_pills.css
+++ b/public/css/elements/_pills.css
@@ -20,8 +20,14 @@
 
     button {
       cursor: pointer;
+      font-size: 1em;
       padding: 0 2px;
-      color: inherit;
+
+      &.close:after {
+        content: 'close';
+        color: var(--color-font-contrast);
+        vertical-align: middle;
+      }
     }
   }
 }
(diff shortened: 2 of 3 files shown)


simon-leech (changes requested):
@cityremade  Can you provide some more info on how to test this? Specifically the bit about the admin panel? 

I tested in default mapp windows and dropdown_pills still have a close button as expected. Just not quite sure how to test the custom admin panel ?

dbauszus-glx (changes requested):
I am happy with the colour.

image


image


Can we adjust the height? Looks like the close button is sitting low.

image

cityremade:
@simon-leech admin panel shouldn't need mapp css stylesheet as there's no map. 
if there's anything that won't work in this view without mapp.css it should be reviewed.

AlexanderGeere (approved):
Looks good to me

dbauszus-glx (changes requested):
This is broken again. Please review previous comments.

What the nature of the close class? How is the :after used?

If the color of nested elements is the same please use inherit.

Please ensure that icon is the same size as the text. Font-size: 1em is not required. 1em is literally the default size for font. https://en.wikipedia.org/wiki/Em_(typography)

AlexanderGeere (approved):
This looks good to me, tested in dark mode as well.

dbauszus-glx:
Alignment is still off on ubuntu / chrome. 
image

RobAndrewHurst (approved):

RobAndrewHurst:
@cityremade Have a look at the changes and let us know when we can squash and merge!

PR #2066, opened 2025-04-16
remove locale.layer merge from addlayer method

dbauszus-glx:
The locale.layer default object is already merged in the getLayer module of the Workspace API.

The locale.layer, layer.template, and layer.templates properties will no longer be returned in Workspace API responses in the future.

The iteration of layers can be simplified with a for of loop.

The layers array can be cloned as a whole, the decoratedLayers object is not required with this.

zIndex can increase itself with ++

Layer objects will always have a key, this is provided by the Workspace API methods.

The decorate layer method checks for the format anyways, no additional check is required.


AlexanderGeere (approved):
Working as expected, The tests still pass. Tested in a couple reports as well.

PR #2068, opened 2025-04-17
CSV Download

simon-leech:
# CSV Download Character Fix

## Description

This PR updates the csvDownload function to correctly handle special characters (e.g., Ł, ę, ż) in CSV exports by adding a UTF-8 Byte Order Mark (BOM). 
This ensures compatibility with Microsoft Excel and other programs that expect a BOM for proper UTF-8 detection.

Previously, the function attempted to export CSV files using:
``` js
type: 'text/csv;charset=windows-1250'
```

The Blob API does not actually encode content as windows-1250.
As a result, the file was still UTF-8 encoded, but mislabeled.
This caused character mismatch (e.g., MÄ™Å¼czyzna instead of Mężczyzna).

## Type of Change

## Code Quality Checklist

### DIFF
--- a/lib/utils/csvDownload.mjs
+++ b/lib/utils/csvDownload.mjs
@@ -33,7 +33,12 @@ export default function csvDownload(data, params = {}) {
     rows.unshift(params.fields.map((field) => field.title || field.field));
   }
 
-  const blob = new Blob([rows.join(params.join)], { type: params.type });
+  // Prepend UTF-8 BOM
+  // This is done to ensure that language-specific characters are displayed correctly.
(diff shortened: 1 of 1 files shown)


AlexanderGeere:
I don't think this will work. We would need to encode the data into windows-1250 which i believe is not possible.

AlexanderGeere (changes requested):
I don't think this will work as we would need to encode the text into the specified charset, which is not possible as far as I understand.

cityremade:
how useful this could be https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

PR #2069, opened 2025-04-17
Histogram template.

cityremade:
Added histogram query template.
The query is designed for continuous data such as disposable income, households etc.

The query requires 2 parameters: table and field.

Optional parameter is a number of buckets which defaults to 7.

Output format is initial, to be agreed how it should be structured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.13.2**
+**v4.14.1**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 ![Codi Unit Tests](https://github.com/GEOLYTIX/xyz/actions/workflows/unit_tests.yml/badge.svg)
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.13.1**
+**v4.13.2**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 ![Codi Unit Tests](https://github.com/GEOLYTIX/xyz/actions/workflows/unit_tests.yml/badge.svg)
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.14.1**
+**v4.15.0-alpha**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 ![Codi Unit Tests](https://github.com/GEOLYTIX/xyz/actions/workflows/unit_tests.yml/badge.svg)
--- a/lib/dictionaries/de.mjs
+++ b/lib/dictionaries/de.mjs
@@ -49,6 +49,7 @@ export default {
   layer_add_new_location: 'Erstelle neue Lage',
   layer_filter_header: 'Filter',
   layer_filter_select: 'Filter Auswahl',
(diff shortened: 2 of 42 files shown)


simon-leech (changes requested):
The query must have the `filter` parameter, and optionally the `viewport` parameter. 

As the query should not attempt to run on all data on a table, but only the data selected. 
e.g. if I search country = UK, the histogram should be built for the UK only, not all countries in the table.

AlexanderGeere (approved):
I added the `filter` and `viewport` to the query otherwise this looks good to me.

cityremade:
in bugs and testing I've added a plugin which adds a button to log histogram data based on layer, filter and viewport with an example layer configuration.

bugs-and-testing: histogram/

dbauszus-glx (changes requested):
I can't make that query work.

This line doesn't feel right. `from \${table} true \${filter} \${viewport}` Is there a `WHERE` missing?

I log my template string with table and field provided as:

```
\n    with percentiles as (\n        select\n        percentile_cont(0.98) within group(order by pop_20_km2) p98,\n        percentile_cont(0.02) within group(order by pop_20_km2) p2\n        from ${table} \n        where true ${filter} ${viewport}\n    ),\n    buckets as (\n        select \n        case \n        when pop_20_km2 <=p2 then 0\n        when pop_20_km2 < p2 + 1*(p98-p2)/7 then 1\nwhen pop_20_km2 < p2 + 2*(p98-p2)/7 then 2\nwhen pop_20_km2 < p2 + 3*(p98-p2)/7 then 3\nwhen pop_20_km2 < p2 + 4*(p98-p2)/7 then 4\nwhen pop_20_km2 < p2 + 5*(p98-p2)/7 then 5\nwhen pop_20_km2 < p2 + 6*(p98-p2)/7 then 6\nwhen pop_20_km2 < p2 + 7*(p98-p2)/7 then 7\n        when pop_20_km2 >= p98 then 8 \n        end as bucket\n        from ${table} true ${filter} ${viewport}\n        cross join percentiles\n    )\n    select \n    count(*)::integer as count, \n    bucket\n    from buckets\n    group by bucket\n    order by bucket\n 
```

With an error near true
`error: syntax error at or near "true"`

dbauszus-glx:
After the last commit the query times out on a reasonable sized table.

dbauszus-glx:
This is working with an integer field on a reasonable sized dataset.

dbauszus-glx (changes requested):
This doesn't work when I set the viewport in my dataview.

```json
        "dataviews": {
          "histogram": {
            "display": true,
            "dataview": "Json",
            "query": "histogram",
            "viewport": true,
            "queryparams": {
              "table": true,
              "field": "pop_20"
            }
          }
        }
```

AlexanderGeere (approved):
made some minor tweaks. Looks good to me!

dbauszus-glx:
i cleanup the loop to make and variable assignment to make this easier to read.

```js
  const decimals = parseInt(_.decimals) || 0;
  const buckets = parseInt(_.buckets) || 7;

  let bucketCases = '';
  for (const i of Array(buckets).keys()) {
    bucketCases += `WHEN ${_.field} < p2 + ${i + 1} * (p98 - p2) / ${buckets} THEN ${i + 1} \n`;
  }

```

can someone help me with a chartjs config?

dbauszus-glx (approved):

PR #2070, opened 2025-04-17
Radio ui element and radiogroup entry.

cityremade:
Initial use case: users switch between model types in order to provide relevant parameters.

Creates `mapp.ui.elements.radio` module with corresponding stylesheet.

Radio is a round interactive element that allows users to choose only one option from a predefined set of mutually exclusive choices.
Radio buttons can be organised into groups identified by their common name and function as interdependent boolean flags.

Creates `radiogroup` entry type which is a set of mutually exclusive options. 

Most of the code is based on checkbox element but this component is designed to update multiple fields at once.
Each radiogroup entry requires its own boolean field.

Radiogroup entry configuration pattern:

```json
{
              "field": "radiogroup",
              "title": "Test radiogroup",
              "label": "Choose me",
              "type": "radiogroup",
              "radiogroup": {
                "name": "my_radiogroup",
                "caption": "this is radio group caption"
              }
            },
            {
              "field": "radiogroup2",
              "label": "No choose me",
              "type": "radiogroup",
              "radiogroup": {
                "name": "my_radiogroup"
              }
            },
            {
              "field": "radiogroup3",
              "label": "No No choose me",
              "type": "radiogroup",
              "radiogroup": {
                "name": "my_radiogroup"
              }
            }
```


RobAndrewHurst:
No issue associated with this

PR #2071, opened 2025-04-17
entryQuery method in mapp/infoj module; add runOnce flag, cancel currently running query in xhr util.

dbauszus-glx:
An entry with a query and the `run` / `queryCheck` flag will always be run when the entryQuery method is executed.

An entry with a query and the `runOnce` flag should only be run once, not everytime the entryQuery method is executed.

The callback method is now assigned as entry.queryCallback(response, entry)

The url is assigned as entry.url with the entry provided to the XHR util as object in order to ensure that the query is cancelled if not resolved.

I tested this with following entry which delays the response by 4 seconds allowing to toggle the location edit which will execute the query again if not set to runOnce.

```json
{
  "type": "json",
  "query": "json_query",
  "queryparams": {
    "table": true,
    "id": true
  },
  "template": {
    "key": "json_query",
    "template": "WITH DELAY AS (SELECT pg_sleep(4)) SELECT ST_asGeoJSON(geom_3857) FROM ${table}, DELAY WHERE glx_id = %{id}",
    "value_only": true
  }
}
```

The hasRan flag has been removed as it wasn't clear what this is supposed to be doing.


AlexanderGeere (dismissed):
```Javascript
   delete entry.runOnce
```

This line causes the warning to trigger when a reload is triggered as the the entry no longer has `run` , `queryCheck` or `runOnce`. 

we could maybe do something like this? 
```Javascript
    entry.runOnce &&
    entry.runOnce !== 'complete' &&
    mapp.utils
      .xhr(entry)
      .then((response) => entry.queryCallback(response, entry));

    entry.runOnce = entry.runOnce ? 'complete' : undefined;
```

dbauszus-glx:
@AlexanderGeere I added a check for the entry.xhr which will be assigned by the xhr util. Can you check whether this does the trick?

AlexanderGeere (approved):
I changed it to `&&` and this now prevents the warning while maintaining the intended functionality. Looks good to me!

PR #2077, opened 2025-04-23
Toast

cityremade:
Toast:

- new ui element module ui.elements.toast
- takes 2 parameters: html / text content to display and optional flag `"information": true` in order to show only OK button. Skip information flag to show accept/reject buttons. Should it be the other way round?
- new stylesheet for toast layout

In order to test:

add the following in any local plugin:

```js
mapp.ui.elements.toast({
     content: 'Hello from ui toast!',
     information: true
});
```
Content can be either html or plain text.
Comment out information flag to show accept/reject.


dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2077/commits/1c7a6a03c1511410ef5a0e4710d86185f078289c

The logo should be optional. We don't want to hardcode branding in the mapp library.

The actions should be defined as an array with optional callback methods.

The ease in animation can be defined in the .toast css specificity. No need to set the display property. The animation will be called when the html element is appended.

The method does not return a HTMLElement but a promise which resolves to the value of the action button.

I have updated the jsdocs.

dbauszus-glx (changes requested):
Please see other comments in regards to the jsdocs and actions.

The button classes should be defined in the buttons css.

We don't want button definitions in contexts. All buttons in the mapp ui should look the same.

Otherwise we get a mess once we override buttons in and out off context.

cityremade:
Example of buttons on youtube:
youtube buttons

Different buttons for different purposes.

cityremade:
Needed css for buttons moved to button stylesheet.
Removed code no longer needed.
Included classlist parameter for actions.
Updated example in docs.

dbauszus-glx:
@cityremade 
The keyvalue_dictionary is used for translations. This allows to translate the content text as well which can not be added to the dictionary.

dbauszus-glx:
@cityremade 

We can add shorthands for the accept and cancel buttons like so:
https://github.com/GEOLYTIX/xyz/pull/2077/commits/98a882b52c4921d0658d9eb84a0c1d261f7a840b

Dictionary is not needed since the content must and can be translated with the keyvalue_dictionary utility.

dbauszus-glx (approved):

RobAndrewHurst (changes requested):
If you create a toast with no actions the toast does not disappear on it's own.

Usually if a toast has nothing to action on it will disappear after a given amount of time.

cityremade:
@RobAndrewHurst exactly. I am thinking to put there default 'ok' or another option - add a cross in top right corner.

RobAndrewHurst:
> @RobAndrewHurst exactly. I am thinking to put there default 'ok' or another option - add a cross in top right corner.

I think you are not misunderstanding me.

If you provide no actions it should just a toast with 0 controls. It would appear and should disappear.

Are you familiar with Toasts/Snack bars in general?

cityremade:
yes I am familiar with that. I close then multiple times a day on different website as none of that stuff disappears.

RobAndrewHurst (changes requested):
When providing 0 actions to the toast element the toast should resolve (disappear) on it's own after a given amount of time. As per Toast/Snack bar notification standards

dbauszus-glx:
The button to close should be optional as not every toast element shuld be dismissed without confirming an action.

I added a 3 second timeout which can be configured to close to selfclose the timeout without action items.

RobAndrewHurst (approved):
LGTM!

PR #2091, opened 2025-04-28
Action buttons with inline icon and text.

cityremade:
Adds .action class in button stylesheet which controls icon sizing.

Add locations panel - drawing buttons now with symbol and text.

### DIFF
--- a/lib/ui/elements/drawing.mjs
+++ b/lib/ui/elements/drawing.mjs
@@ -115,8 +115,9 @@ function point(layer) {
   // Create the button
   layer.draw.point.btn = mapp.utils.html.node`
     <button
-      class="flat wide bold"
+      class="action wide"
       onclick=${(e) => drawOnclick(e, layer, layer.draw.point)}>
+      <span class="material-symbols-outlined">add_location_alt</span>
       ${layer.draw.point.label}`;
 
   return layer.draw.point.btn;
@@ -148,8 +149,9 @@ function line(layer) {
   // Create the button
   layer.draw.line.btn = mapp.utils.html.node`
     <button
-      class="flat wide bold"
+      class="action wide"
       onclick=${(e) => drawOnclick(e, layer, layer.draw.line)}>
+      <span class="material-symbols-outlined">timeline</span>
       ${layer.draw.line.label}`;
 
   return layer.draw.line.btn;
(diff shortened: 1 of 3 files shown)


lib/ui/elements/confirm.mjs
```diff
@@ -55,15 +55,15 @@ export default function confirm(params) {
         e.target.closest('dialog').close();
         resolve(true);
       }}
-      class="raised primary-colour bold">
+      class="raised old">
```
RobAndrewHurst, line 58: Spelling issue

RobAndrewHurst:
@cityremade There is a merge conflict, can you resolve when you get the chance?

lib/ui/elements/confirm.mjs
dbauszus-glx: resolved in commit.

simon-leech (changes requested):
image

The Circle From Centre label should be within the Circle Configuration group I think. Otherwise its unclear that that group has anything to do with the button action. 

Is there a reason we removed the bold class from the drawing buttons? I think we should retain that as custom plugins use bold and we want to make sure things align? 

I really like the addition of the icon next to the text - really helps explain what's its for! Nice! 👏

dbauszus-glx (changes requested):
image

Working as expected.

The only thing I don't understand is the easeout transition for button.

What is this required for?

image

cityremade:
ease is for softer property change on the button. Without it we have this hard unpleasant glitch.

This pr is not about the layer editing.

I can remove the changes to the drawing panel as I have put them for testing. This layout is a separate problem to address.

Once you are happy with the button itself I can restore classes on edit buttons.

simon-leech:
> ease is for softer property change on the button. Without it we have this hard unpleasant glitch.
> 
> This pr is not about the layer editing.
> 
> I can remove the changes to the drawing panel as I have put them for testing. This layout is a separate problem to address.
> 
> Once you are happy with the button itself I can restore classes on edit buttons - that is make the button `.flat` again instead of `.action`. Also I have removed bold on text as it is not readable when every element is bold.

I wouldn't remove the changes on the drawing panel - I think they look really good so no benefit to undoing that. 

I'm testing again now - understood on the bold, good point! :)

simon-leech (approved):
image

I added the icon to the Modify Geometry while testing this. 
All looks good to me 👍

PR #2095, opened 2025-04-29
make btn-column scrollable

dbauszus-glx:
The btn-column element should be nested inside the map element with a max-height of that element.

The content should be scrollable if the max-height is exceeded.

The scrollbar should not be displayed.

The same logic can apply for mobile view restrictions on screen width.

The display:none element at the bottom is no longer required.


AlexanderGeere (approved):
I added the firefox scrollbar removal, Looks good to me otherwise.

cityremade (approved):

RobAndrewHurst (approved):

PR #2096, opened 2025-04-29
Cookie module should return null without an ACL configured.

dbauszus-glx:
This prevents an Error on a public instance without an ACL.

You will still see a login button which is a default syncPlugin.

Either set the syncPlugins to empty array or just the zoomBtn to not see the login.

```json
"syncPlugins": [],
```

image


RobAndrewHurst (approved):
LGTM

AlexanderGeere (approved):

simon-leech (approved):

PR #2112, opened 2025-05-02
userIDB plugin

dbauszus-glx:
The userIDB plugin provides an interface to store the user object in the userIndexDB 'user' store.

image

The userObject returned from the store by name will be assigned the mapp.user object.

image

'anonymous' will be used as email value if the mapp.user is falsy.


AlexanderGeere (changes requested):
When saving the object is stored with a strange name:
Screenshot from 2025-05-06 15-50-34

This can be viewed if you inspect and navigate to the storage tab.

dbauszus-glx:
@AlexanderGeere thanks for finding this. This was actually a huge logical issue which I hopefully have addressed in the last commit.

The user store should be within the MAPP indexedDB.

While the locales store should be in a workspace indexedDB. The workspace being derived from the process.env TITLE.

AlexanderGeere (approved):
Nice one! looks good to me.

RobAndrewHurst (approved):

PR #2114, opened 2025-05-02
Check roles in template merge

dbauszus-glx:
roles will be provided as optional argument to the mergeTemplates method.

The objTemplate method will check whether role restriction permit access to a template before the template is merged into the object [eg locale, layer].


simon-leech (approved):
Working for me - I correctly see role-restricted nested locales ignore when i don't have the role.

AlexanderGeere (approved):
Looks good to me.

cityremade (approved):

PR #2117, opened 2025-05-07
Return dropdown node without pills container.

dbauszus-glx:
An undefined [`params.pills?.container`] ahead of the dropdown button [node] will return a document fragment.

The params.node should be assigned as a node with the button element or input group.

The node should be composed with the `params.pills?.container` if available. This will return a document fragment.

With this PR, a regular dropdown button without pills will return a node [button] which can be disabled like so.

```js
        const localesDropdown = mapp.ui.elements.dropdown({
          data_id: 'locales-dropdown',
          span: locale.name || locale.key,
          entries: locales.map((locale) => ({
            title: locale.name || locale.key,
            key: locale.key,
          })),
          callback: (e, entry) => {
            window.location.assign(`${mapp.host}?locale=${entry.key}`);
          },
        });

        localesDropdown.disabled = true;

        layersTab.appendChild(localesDropdown);
```

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -92,8 +92,7 @@ export default function dropdown(params) {
   }
 
   if (params.pills?.container) {
-
-    params.node = mapp.utils.html.node`${params.pills.container}${params.node}`
+    params.node = mapp.utils.html.node`${params.pills.container}${params.node}`;
   }
 
   return params.node;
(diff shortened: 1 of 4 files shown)


cityremade (approved):
It fixes my problem, I can now access the element.

RobAndrewHurst (approved):
Happy with this to go in!

simon-leech (approved):

PR #2119, opened 2025-05-08
max height property and blur event.

cityremade:
Removed: 

programmatic setting of ul max height based on height difference vs viewport.

Added: 

parameter maxHeight for dropdown which sets `max-height` property in pixels, on dropdown options list, number.
*It was used in filter panel dropdown while not supported by the dropdown element

`blur` event on dropdown button - button emits blur event when another element is clicked or goes into focus. 
*On blur the result list will close.

As before:
`<ul>` gets max height of 500px in the dropdown stylesheet.

---
  - https://app.asana.com/0/0/1209927918503567


cityremade:
found on Mac OS: when dropdown is open in a scrollable window parent window doesn't scroll anymore. 
It scrolls again after dropdown is closed.
So open dropdown comes into focus and other events are suppressed.
Dropdown out of focus/closed - parent window scrollable again.

dbauszus-glx:
The dropdown will now collapse in an intersection observer callback.

https://github.com/user-attachments/assets/3e39f2df-2dde-4ab1-b1bc-7c30c17c5e5c

RobAndrewHurst:
PR needs to be based on the patch branch

PR #2120, opened 2025-05-09
Allow configuration to filter the available options for circle drawing

simon-leech:
Previously it was not possible to say which units you allow for circle configuration. they were all always included. 
This PR introduces the `layer.draw.circle.unitsOptions` array which can be assigned to reduce the options. If it is not included, then they will all be included as normal. 
``` "unitsOptions": ["meter", "km"]``` will subset to just allow the meter and km options.

### DIFF
--- a/lib/ui/elements/drawing.mjs
+++ b/lib/ui/elements/drawing.mjs
@@ -346,17 +346,21 @@ function circle(layer) {
   layer.draw.circle.unitsOptions ??= ['meter', 'km', 'miles', 'meter2', 'km2'];
 
   const optionsGiven = layer.draw.circle.unitsOptions.join(', ');
-  const unitsAvailable = units.map(unit => unit.option).join(', ');
-  
-  if (!units.some(unit => layer.draw.circle.unitsOptions.includes(unit.option))) {
-    console.log(`layer.draw.circle.unitsOptions - ${optionsGiven}: not found in available options. They are ${unitsAvailable}`);
-    
-    // Set the layer.draw.circle.unitsOption to just meter 
+  const unitsAvailable = units.map((unit) => unit.option).join(', ');
+
+  if (
+    !units.some((unit) => layer.draw.circle.unitsOptions.includes(unit.option))
+  ) {
+    console.log(
+      `layer.draw.circle.unitsOptions - ${optionsGiven}: not found in available options. They are ${unitsAvailable}`,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by AlexanderGeere — CHANGES_REQUESTED** [review]
A warning for if you put an invalid string in the config would be nice. 

Im also getting this error:
Screenshot from 2025-05-09 16-12-51

with:
Screenshot from 2025-05-09 16-13-30

The `find` is presumably not finding anything cause the error above.

**Comment by simon-leech** [comment]
@AlexanderGeere  Added a warning if the option is not valid and default to just meters. Ensure the unit is set to the first one in the configured units provided too.

**Review by AlexanderGeere — APPROVED** [review]
I rearranged things a bit so we can assign the radiusMax/Min and radius from the options provided, as when it initially loaded i could select a circle that was 100s of KMs.

I also changed the log to a warn for the misconfiguration.

**Review by dbauszus-glx — APPROVED** [review]

**Review by AlexanderGeere — APPROVED** [review]

**Comment by dbauszus-glx** [comment]
@AlexanderGeere why is this needed?

The unitsConfig is only needed for lookup of values.

The unitsOptions array will be filtered.

**Comment by AlexanderGeere** [comment]
@dbauszus-glx  without it i was still seeing all the options even when trying to only have km2 for example. 
Screenshot from 2025-05-12 16-33-08

The dropdown pulls its entries from `unitsConfig`:
``` Javascript
    entries: Object.values(circle.unitsConfig),
```

dbauszus-glx:
@AlexanderGeere Ok, valid in that case. The forEach doesn't return anything though. 

I think the circle.units assignment from the keys array should happen inside the check if the current units is not in the unitsConfig.
```js
  circle.units = Object.keys(circle.unitsConfig)[0];

  if (!Object.hasOwn(circle.unitsConfig, circle.units)) {
    console.warn(`Unkown circle units conversion ${circle.units}`);
    circle.units = 'meter';
  }
```

dbauszus-glx (approved):

jfitzpatrick17:
Don't want to distract from the review of this, but with the new set-up could this now support users being able to choose their default metrics (in the user settings area for example - e.g. km or miles), or would that be a separate issue we'd want to create?

dbauszus-glx:
@jfitzpatrick17 Seperate issue and not easily doable.

PR #2122, opened 2025-05-09
Ensure layer always has layer.params.fields

simon-leech:
A `type:layer` entry may not have a `params.fields` value if not required. 
This will result in an error that `layer.params.fields is not iterable`. 
This PR nullish assigns to resolve this on the vector layer itself.

### DIFF
--- a/lib/layer/format/vector.mjs
+++ b/lib/layer/format/vector.mjs
@@ -56,7 +56,8 @@ export default function vector(layer) {
   }
 
   // Set default layer params if nullish.
-  layer.params ??= { fields: [] };
+  layer.params ??= {};
+  layer.params.fields ??= [];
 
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
I think this might better be assigned when and where required. I'll have a look.

My concern being that there are checks on whether the featureFields are reset inb regards to this param. This should not be set on every layer just because but only on layers which require the param.

```js
export function reset(layer) {
  if (!layer.params.fields) return;

  // Create featureFields object if nullish.
  layer.featureFields ??= {};

  layer.params.fields.forEach((field) => {
    // Set empty values array.
    layer.featureFields[field] = {
      values: [],
    };
  });
}
```

dbauszus-glx (approved):

PR #2124, opened 2025-05-12
Expose Dark Mode Theme

simon-leech:


AlexanderGeere (changes requested):
The dark mode plugin doesn't work on this PR

simon-leech:
@AlexanderGeere  Good spot - should be corrected now!

dbauszus-glx:
The theme object should not be defined on the root of the utils since this is not a utils method and the setTheme utils method is dependent on the configuration object.

I moved the themes inside utils.cssColour.

dbauszus-glx (approved):
please not change of structure in utils.

AlexanderGeere (approved):
Working as expected!

PR #2127, opened 2025-05-14
Bump jsoneditor import from esmImport util

dbauszus-glx:
The jsonEditor should imported through the esmImport util.

The jsonEditor dependency has been bumped to 3.3.1 which includes a ton of patches. https://github.com/josdejong/svelte-jsoneditor/releases


RobAndrewHurst (approved):
Nice one! 

Looks good to me!

PR #2129, opened 2025-05-14
Add mapboxStyle layer format

dbauszus-glx:
This PR adds the mapboxStyle vector tile format to the layers format.

The ol-mapbox-style plugin is imported from unpkg. https://github.com/openlayers/ol-mapbox-style

The mapboxStyle format works the same as the maplibre format for mapbox style layer but is a much smaller import, support v12+ street styles, and is native styled in Openlayers. This should perform better but may have slight differences in style display.

```json
      "style_new_ol": {
        "format": "mapboxStyle",
        "style": {
          "URL": "mapbox://styles/dbauszus/cm6hs2100003c01sdclwv2o88"
        },
        "accessToken": "***"
      }  
```


simon-leech (commented):
If the format works the same as the maplibre but its faster. Is there any benefit to maintaining both? Could we not just push legacy `format: maplibre` to use the `mapboxStyle` format?

simon-leech:
Its working fine for me though - not sure if its because you said it and its confirmation bias but pan and zoom with mapboxStyle layer does seem faster for me!

RobAndrewHurst (approved):
Happy with this!

AlexanderGeere (approved):
I like the look of it, nice one.

Issue #1882, opened 2025-02-07
Dataview in dialog


dbauszus-glx:
In order to better test the dialog element it would be nice to have an option to open a dataview with dialog as target.

Or perhaps provide a dataview toolbar utility that opens the data in a jsoneditor in a dialog.

PR #2130, opened 2025-05-14
feat(dataview): target dataview/s in dialogs

AlexanderGeere:
## Description
This PR enables a new option for dataview targets. Setting `target: dialog` will spawn the dataview in a dialog. 

A new module has been added `dataviewDialog.mjs` which is used to create the dialog. You can pass in 
```JSON
   "dialog_css":{
      "width": 400,
      "height": 400
    }
```

To control the size of the dialog.

The dialog is also resizable, by dragging the resize indicator - which should be in the bottom right corner of the dialog. 

## GitHub Issue
[#1882](https://github.com/GEOLYTIX/xyz/issues/1882)

## Type of Change
- ✅ Documentation

## How have you tested this?
This has been tested locally using `bugs_testing/dataviews/dataviews.json`.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


simon-leech (changes requested):
This is working well for me so far on initial testing! 👍  

Two comments: 
1. Can we make the resize handler slightly larger - its really tiny on chrome so the user might not know its possible? 
2. Using `queryCheck` with a `type:dataview` when the data returned is null should disable the checkbox. This isn't working at the moment. Not sure if we want to separate this into another patch PR? 
``` json 
"queryCheck": true,
 "query": "query_here",
      "template": {
        "key": "query_here",
        "template": "select null"
      }
```

AlexanderGeere:
@simon-leech as far as the resizer, its system based so its not easily changed. We could do something like this
```html
 dialog::-webkit-resizer{background-image: url(/path/to/image);}
```

But I dont know if we want to go down that road.

dbauszus-glx:
We should stick to the system elements. Different systems provide native accesibility tools to enhance the user experience if required.

simon-leech (approved):
I approve this as its working for me as expected. 

Understood on using the native resize and the `queryCheck` issue is an issue off the PR, so its a bug that should be addressed in a patch instead.

dbauszus-glx (dismissed):
The dataviewDialog method should be in the ui.utils instead of the ui.elements since the method does not return an element.

You could use a check `if (dataview.target !== 'dialog') return;` at the beginning of the dataviewDialog method. This way you don't have conditionally call the method.

You can also add this to the dataviewDialog method on the condition that the dataview has a location with removeCallbacks.

```js
    entry.location.removeCallbacks.push(() =>
      entry.dataview_dialog.dialog.close(),
    );
```

This one liner should not be a seperate method. This makes it just more difficult to read and debug.
```js
function assignTarget(dataview) {
  dataview.target =
    dataview.target instanceof HTMLElement
      ? dataview.target
      : mapp.utils.html.node`
        <div class="dataview-target">`;
}
```

I added todo for the content_panel and toolbar part in the dataviewDialog method which I do not completely understand.

AlexanderGeere:
@dbauszus-glx  I had to reduce the complexity of the module so i moved the tabview stuff into its own function. 

The toolbar can only be made to display in the show function as this does not exist until `mapp.utils.Dataview` is called.

lib/ui/utils/dataviewDialog.mjs
```diff
@@ -0,0 +1,98 @@
...
+  mapp.ui.elements.dialog(dataview.dataview_dialog);
+
+  //Mimic the css of the tabview to get similar user experience.
+  const content_panel = dataview.dataview_dialog.node.querySelector('.content');
+
+  content_panel.classList.add('panel');
+
+  dataview.target = content_panel;
+
+  //dataview.target.appendChild()
```
simon-leech, line 63: Do we need this commented out code?

lib/ui/utils/dataviewDialog.mjs
```diff
@@ -0,0 +1,98 @@
...
+  dataview.hide = () => {
+    dataview.dataview_dialog.dialog.close();
+  };
+
+  //Close the dialog if the location is removed.
+  dataview.location?.removeCallbacks?.push?.(() =>
+    dataview.dataview_dialog.dialog.close(),
+  );
+
+  //Charts need to be recreated when the dialog os resized.
```
simon-leech, line 92: Minor spelling mistake in comment

simon-leech (dismissed):

dbauszus-glx:
@simon-leech @AlexanderGeere A dialog target must have a checkbox, like a tabview target.

simon-leech:
I'd like to re-review this only once the querycheck PR is in and merged, as would like to check that's working against the dialog target too

dbauszus-glx (changes requested):
There should be a reasonable min size if undefined. It should also be possible to set a min size eg. min width and min-height.

image

AlexanderGeere:
@dbauszus-glx i set the minimum to 310x300 and you can set the minWidth and minHeight using:
```JSON
"dialog_css": {
  "minHeight": 100,
  "minWidth": 150,
}
```

dbauszus-glx (changes requested):
Almost there.

I commented out the resize observer. The container panel doesn't resize with the dialog.

image

Hence the chart can not auto resize which is the default.

https://www.chartjs.org/docs/latest/configuration/responsive.html

AlexanderGeere:
@dbauszus-glx to avoid recreating the chart, I changed the resize observer so that it instead changes the size of the panel which allows the chart to auto resize, alternatively we could make it so that charts are not resizable. Without the resize observer you can make the chart bigger but you could not make it smaller without reloading.

cityremade (approved):
looks good to me!

RobAndrewHurst (approved):

dbauszus-glx (approved):
the resize is now working without an observer

PR #2131, opened 2025-05-14
Close dropdown in intersection observer event method

dbauszus-glx:
This PR introduces the maxHeight property for dropdown elements.

This PR also introduces an intersectionObserver for the dropdown.node. The oberserver element will close the dropdown element if the position of the parent is changed [eg. from a scroll event].

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -23,6 +23,7 @@ The dropdown method returns a dropdown element created from the params argument.
 @property {boolean} [params.pills] Applies pills component to show selected values.Available for mutliple choice.
 @property {Object} [params.entries] Array of option elements. Expected format: [{title: 'Title for Option 1', option: 'option1'}, ...]. Add property `selected: true` for entry selected by default. `title` will appear as the description, `option` is a value passed as selected.
 @property {function} [params.headerOnClick] Callback to execute when a header is clicked. Overrides default function.
+@property {Number} [params.maxHeight] Optional max height property on results unordered list <ul> in pixels. Defaults to 500px in dropdown stylesheet.
 
 @returns {HTMLElement} HTML dropdown element
 */
@@ -82,6 +83,10 @@ export default function dropdown(params) {
   } else {
     params.node = mapp.utils.html.node`
     <button
+      onblur=${(e) => {
+        // hide results list when button out of focus
+        e.target.classList.remove('active');
+      }}
       data-id=${params.data_id || 'dropdown'}
       class="dropdown">
       <div class="head" onclick=${params.headerOnClick}>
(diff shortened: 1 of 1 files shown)


simon-leech (dismissed):
If I open up a dropdown e.g. a filter near the bottom of the layer panel, previously the dropdown would open upwards so it wasn't lost off the page. 
On this PR, the arrow changes to indicate a click has been registered, but the dropdown cannot be seen. If i scroll so the dropdown is no longer at the bottom of the page, it can be seen. This should work as it did previously where a dropdown that's too large opens upwards. 

Dropdowns shown in a tabulator table are no longer visible either, the dropdown arrow changes but the content is not seen.

dbauszus-glx:
I fixed the issue in regards to the postioning and sizing of the dropdown.

https://github.com/user-attachments/assets/6aa12da3-4776-4b41-b31f-7b4b9ac169ba

I need to see the table dataview issue.

dbauszus-glx:
@simon-leech The positioning should be fixed for use in tabulator tables. The positioning must respect the mobile display where the offset may differ from the clientBoundingRect.top.

AlexanderGeere (changes requested):
I can still scroll the behind the dropdown in the multi-filter: 
Screenshot from 2025-05-19 10-02-29

You can see this on `bugs_testing/filter/all_filters.json` the multi-filter is on the scratch layer.

dbauszus-glx:
Using a custom dropdown element is impossible difficult to position without an anchor css property which isn't yet fully formalized or supported.

The select / option element is positioned correctly but not fully stylable. The position is more important than the styling.

This PR introduces a select / option element for dropdowns.

We can adjust the styling in order to make this look like current with the limitation that the styling will not be fully supported by all platforms.

@AlexanderGeere The search will not work with a select/option element. Can you try to implement a native search element with the entries used to populate the <datalist> in a similar fashion like the select <options>. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/search

AlexanderGeere:
@dbauszus-glx I have added a search element.

dbauszus-glx:
@AlexanderGeere This is looking pretty good already. I moved the searchInput method from it's own module into the dropdown module and unnested the onInput method as onSearchInput with the options assigned to the params object as searchOptions.

Is it possible to remove the dropdown icon which shows on hover over the search input?

Is it also possible to remove the duplicate item with the tick? Ideally a class should be set on the item which can be styled with a background colour.

image

I removed the the dropdown copy now and also the dropdown style, and the multi_dropdown which is no longer used anywhere.

simon-leech (changes requested):
This isn't working properly for me. 
The styling looks different which I think will be something we should override. 

The main issue I'm seeing is that when I select an option from a dropdown list, the option is not highlighted in blue until I select the next option?

simon-leech:
image

Dark mode styling needs correcting too

simon-leech:
image

Locale dropdown is empty on load too

dbauszus-glx:
I updated the background-color and color for the select element to be applied for dark_mode.

dbauszus-glx:
The locale dropdown uses the lagacy span property for the placeholder. I have updated this.

simon-leech:
@AlexanderGeere  If I type into a dropdown_pills entry I can end up with an empty pill being built. 
image

To do this I just type letters and then delete them, and it adds an empty pill which shouldn't happen.

simon-leech (approved):
Good with me - think I've tested all the dropdown locations 🤞😅

AlexanderGeere (approved):
This looks good to me.

PR #2134, opened 2025-05-15
userLocale export to allow for custom plugins to use it

simon-leech:
WIP

### DIFF
--- a/lib/ui/utils/userLocale.mjs
+++ b/lib/ui/utils/userLocale.mjs
@@ -1,0 +1,60 @@
+export default function userLocale(params, mapview) {
+  params.localeInput = mapp.utils.html.node`<input
+    type="text" value="${mapview.locale.name || mapview.locale.key}">`;
+
+  params.ulLocales = mapp.utils.html.node`<ul>`;
+
+  params.panel = mapp.utils.html.node`
+    <div 
+      class="drawer"
+      style="border: 1px solid #333;">${mapp.user.email || 'anonymous'}|${mapview.locale.workspace}
+      ${params.ulLocales}
+      ${params.localeInput}
+      <button class="raised"
+        onclick=${saveLocale}>Save</button>     
+      <button class="raised" 
+        onclick=${removeLocale}>Remove</button>`;
+
+  // Call the listLocales method to render the list of userlocales.
+  listLocales(params, mapview);
+
+  // Return the panel as content to be displayed.
+  return params.panel;
+}
+
+async function removeLocale() {
+  params.panel.classList.add('disabled');
+  mapview.locale.name = params.localeInput.value;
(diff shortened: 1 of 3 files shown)


lib/dictionaries/en.mjs
```diff
@@ -119,5 +119,9 @@ export default {
   toolbar_zoom_out: 'Zoom out',
   toolbar_zoom_to_area: 'Zoom to area',
   units: 'Units',
+  user_locale_projects: 'Below is a list of your current saved projects:',
+  user_locale_specific: 'Your projects are specific to you.',
```
RobAndrewHurst, line 123: The wording doesn't make sense for the name of the functionality.

should it be `userLocale` or `userProject` ie 

`Below is a list of your current saved locales:`,
`Below is a list of your current saved projects:`

lib/ui/utils/userLocale.mjs
```diff
@@ -0,0 +1,100 @@
...
+The list includes a delete button for each userLocale, which when clicked will remove the userLocale from the mapview and update the list of userlocales.
+*/
+async function listLocales(params, mapview) {
+  const locales = await mapp.utils.userLocale.list(mapview.locale.workspace);
+
+  const list = locales.map((locale) => {
+    const href = `${mapp.host}?locale=${locale.key}&userlocale=${locale.name}`;
+
+    const removeBtn = mapp.utils.html`<button
+       onclick=${async () => await removeLocale(params, mapview)}>
```
RobAndrewHurst, line 92: This won't delete the relevant locale in the list but the current as it's the current params when the list gets loaded

RobAndrewHurst (dismissed):

lib/dictionaries/en.mjs
simon-leech: Changed to use locales wording

simon-leech:
@RobAndrewHurst  this should be ready for rereview now :)

AlexanderGeere (approved):
This is working as expected.

dbauszus-glx (changes requested):
The userLocale element should be added to ui.elements instead of ui.utils to avoid confusion with the utils.userLocale which calls the userLocale elements method.

There is an inline max-height which isn't working.

There shouldn't be a scrollbar if the maxheight isn't reached.

The ul should element should have a class name which defines the max-height and scroll behaviour. Scroll should be on y axis only.

The scrollbar should be auto and now show if not required.

image

dbauszus-glx:
I changed the location of the userLocale to be an element method.

simon-leech (commented):
All looks good to me - thanks @AlexanderGeere ! :)

dbauszus-glx (approved):
Looking good now. I made the scroll auto to not show the scroll bar when not needed.

RobAndrewHurst (approved):

PR #2136, opened 2025-05-15
fix(theme): Categorized Theme missing nowrap property

simon-leech:
The graduated thematic allows for a flex legend with no wrap. 
We implemented this on the categorized theme too but for some reason we never implemented the nowrap. 
This PR simply adds the nowrap property to fix it.


dbauszus-glx (changes requested):
The legend must be scrollable for the overflow.

image

dbauszus-glx (approved):

PR #2139, opened 2025-05-15
feat(mapview): set zIndex in addLayer as mapview property

dbauszus-glx:
This PR ensures that additional layers provided to the mapview addLayer method will get an increasing zIndex.


AlexanderGeere (approved):

simon-leech (dismissed):
I'm seeing an issue on this so will add more info later on 👍

simon-leech:
image

I'm seeing layers without a zIndex at all. 

The code being ran is as follows: 

``` js
 await retailPlaceMapview.addLayer(
          params.locale.layers
            .filter((layer) => new Set(rpLayers).has(layer.key))
            .map((layer) => {
              console.log('Add layer:',layer.key, layer.zIndex);
              layer.display = true;
              return layer;
            })
        );

        // Log all the layer zIndex for layers on retailPlaceMapview
        Object.values(retailPlaceMapview.layers).forEach((layer) => {
          console.log('After layer added:',layer.key, layer.zIndex);
        });
```

cityremade:
and layers get this by defined order in the workspace? does grouping matter?
Is there impact on additional geometries? like catchment displayed under location pin ?

dbauszus-glx:
@cityremade Layers added to the mapview will get an increasing zIndex.

Eg. I add layers with zIndex 10, undefined, 20, undefined. Then first undefined should get a zIndex of 1, and the second undefined 2. I add more layers then the next undefined zIndex will be assigned as 3.

simon-leech:
Question too- should this be minor or patch PR? It seems to be a bug to me that layers are being given a zIndex of 0 even when other layers on the mapview have a value?

dbauszus-glx:
I think this is a minor because the module had changed in the current minor.

dbauszus-glx:
It's not technically a bug since no error is thrown. The zIndex is counted from 0 on every add layer and the PR changes this to store the last zIndex on the mapview.

dbauszus-glx:
@simon-leech  the mapview.zindex is not initiated at 1, not 0.

simon-leech (approved):

RobAndrewHurst (approved):

PR #2141, opened 2025-05-16
fix(entries.layer): Assignment of layer clone to entry object 

dbauszus-glx:
Remove vector layer in minor.

The assignment of the layer clone to the entry object breaks successive iteration of the layer entry method.

However layer properties should not everwrite entry methods. Let's say a type:layer entry has a style object. Then the style object of the entry should be used. Not the style object from the layer clone.

This PR filters the layer properties and only assigns properties to the entry which are not own properties.

```js
    // Filter out clone properties owned by the entry.
    const filtered = Object.keys(layer)
      .filter((key) => !Object.hasOwn(entry, key))
      .reduce((obj, key) => {
        obj[key] = layer[key];
        return obj;
      }, {});

    // Assign layer object to entry overriding entry properties.
    Object.assign(entry, filtered);
```

Changed the key to use the entry.location.id not the OL UID as its then linked to the location
Tweaked the decorateLayer method so that if the layer exists and is currently displayed, call the showLayer to potentially refresh the data


simon-leech:
@dbauszus-glx I pushed two minor changes: 

1. Change the key to use the entry.location.id not the OL UID as its then linked to the location
2. Tweaked the decorateLayer method so that if the layer exists and is currently displayed, call the showLayer to potentially refresh the data. This is required for if the location is updated, we need to ensure that associated layers are reloaded correctly.

simon-leech (approved):

RobAndrewHurst (approved):
Nice one!

Happy for this to go in

PR #2146, opened 2025-05-20
User Locale UAT

simon-leech:
Changes from User Locale User Acceptance Testing:
1. Dictionary terms updated
2. Added timestamp for when a locale is saved and shown in the UI when available
3. Removed check for lcoale.userLocale plugin as this isn't required as we export the method now for use elsewhere (this still isn't working though)


dbauszus-glx:
I'll close this in favour of this more concise PR which does just focus on the userLocale load in the default script.
https://github.com/GEOLYTIX/xyz/pull/2157

PR #2151, opened 2025-05-21
Assign mapButton to mapview object

dbauszus-glx:
The mapButton element should be assigned to the mapview object in the decorator call.

This way plugins can check for the mapview.mapButton instead of requiring a getElementById lookup.

This is backwards compatible since the element still has the same id.


cityremade (approved):

RobAndrewHurst (approved):
LGTM

PR #2157, opened 2025-05-21
Check userLocale in default script

dbauszus-glx:
The check for the userLocale should not be dependent on the userLocale plugin.

### DIFF
--- a/lib/utils/userLocale.mjs
+++ b/lib/utils/userLocale.mjs
@@ -27,14 +27,11 @@ A decorated locale object must be parsed with the jsonParser utility method in o
 The get method requests a locale stored in a userIndexedDB.
 
 @param {Object} locale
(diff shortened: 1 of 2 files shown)


RobAndrewHurst:
image

I get some weird behavior where I can save a userLocale and then when i select it from the list and it reloads into it, and then select the settings button again I get the modal mask appearing on screen but no modal.

You can see in the second screenshot I have clicked the userSettings button, bu nothing appears and I don't get any errors in the browser.

image

dbauszus-glx:
That issue should be addressed in the userSettings plugin. This PR should work without any plugin used.

simon-leech:
Working for me - will open another PR in a minute once this is merged to add the language translations and timestamp back in for release-ready.

PR #2162, opened 2025-05-22
Add description of csvDownload module methods.

dbauszus-glx:
This PR adds documentation to the csvDownload module methods.

The type param is to be tested for exports of strings with a different charset, eg latin.


AlexanderGeere (approved):
The documentation looks good. I think the text encoding will require a different PR

PR #2168, opened 2025-05-27
(bug): Encode userLocale Naming 

simon-leech:
UserLocale naming input must be encoded to allow special characters.
This must then be decoded when displaying the user locale href.


AlexanderGeere (approved):
Working as expected.

PR #2173, opened 2025-05-28
Clean up Openlayers Icon Style creation

dbauszus-glx:
The iconStyle method should not be defined in an forEach iteration.

### DIFF
--- a/lib/utils/olStyle.mjs
+++ b/lib/utils/olStyle.mjs
@@ -31,7 +31,6 @@ The olStyle method takes a mapp-style JSON representation to create an Openlayer
 
 @returns {Object} An Openlayers feature style object.
 */
-
 export default function olStyle(style, feature) {
   if (!style) return null;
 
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
Those are working now, I'm now hitting issues with using a `template_pin` `svgTemplates` though. 

``` json
{"icon":[{"type":"template","template":"template_pin","substitute":{"#FF69B4":"#000000"}}]}
```

dbauszus-glx:
The issue with svg lookup in icon object and icon array should now be resolved.

dbauszus-glx:
@simon-leech array of 1 icon is now turned into an icon object.

icon in theme.style is now checked.

simon-leech:
@dbauszus-glx One small thing left - the cluster style isn't being parsed I don't think. I'm ending up with white clusters when I have an icon of length 1 in the cluster style.

dbauszus-glx:
@simon-leech Can you give us a sample cluster style to check?

AlexanderGeere:
This configuration:
image
 works off the PR but not on it, is this not supported anymore? Should this then not be on minor as this is a breaking change.

simon-leech (changes requested):
Please don't merge until i've reviewed again

simon-leech:
> This configuration: image works off the PR but not on it, is this not supported anymore? Should this then not be on minor as this is a breaking change.

This is still not working on this PR - this needs to be fixed before this PR can be merged in.

dbauszus-glx:
@simon-leech  Array icons need to be reviewed. The configuration with an svg in an icon array should still work.

dbauszus-glx:
@simon-leech 
The issue with the svg property in icon arrays should be resolved in this commit. https://github.com/GEOLYTIX/xyz/pull/2173/commits/9eb8540c1a62e70e6f37149e0ce55cb5a463be81

simon-leech (approved):
Happy with this now - I tested on a range of different configurations and seems to be working. 
Would be good if everyone else can give it another once over though on a range of client configurations!

cityremade (commented):
distribution: count theme - is it expected to have just cluster in legend?

dbauszus-glx:
@cityremade Dependent on the configuration it is possible that only the cluster is shown. The behaviour though should not be different in this PR. Can you check against the minor whether the legend looks the same or different for this layer?

cityremade:
yes it's the same.

RobAndrewHurst (approved):

PR #2176, opened 2025-05-28
feature(roles): Return roles.tree from workspace, support roles with dot separator.

dbauszus-glx:
This PR adds the option to define a role as a dot seperated string. eg. `Europe.UK.brand_a`

The workspace roles method will pop the last part into the array of roles returned from the workspace.

A JSON object will be returned instead of an array with the `?tree=true` parameter.

eg.
```
{
  Europe: {
    UK: {
      brand_a: {}
    }
  }
}
```

The method to parse a Set of roles from workspace has been moved to the roles utility module.

Any current configurations should work as expected but it should be possible to define a nested structure in order to build an interface to define nested roles.

### DIFF
--- a/lib/layer/format/maplibre.mjs
+++ b/lib/layer/format/maplibre.mjs
@@ -75,7 +75,9 @@ export default async function maplibre(layer) {
     keyboard: false,
     pitchWithRotate: false,
     pixelRatio: 1,
-    preserveDrawingBuffer: layer.preserveDrawingBuffer,
+    canvasContextAttributes: {
+      preserveDrawingBuffer: layer.preserveDrawingBuffer,
+    },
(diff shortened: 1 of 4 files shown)


cityremade (approved):
working as expected.

AlexanderGeere:
So setting up roles like this: 
```Json
"x.y.z": {
            "draw": {
              "circle": {
                "popout": true
              }
            }
          }
```

Yields this: 
image

Just making sure that extra z is expected.

This is done for backwards compatibility.

AlexanderGeere (approved):
This looks good to me. I tried some complicated permission structures, working as expected.

RobAndrewHurst (approved):

PR #2179, opened 2025-05-28
Dictionaries From Public

simon-leech:
1. Load en as default in _dictionaries.mjs
2. Make all language files json not mjs
3. Add utility to mergeLanguage
4. Move language translations into public/dictionaries
**To do** 
- Test if this works
- Check how to use the proxy to load languages correctly 
- Any updates to the default view needed?

### DIFF
--- a/lib/layer/format/mapboxStyle.mjs
+++ b/lib/layer/format/mapboxStyle.mjs
@@ -26,10 +26,7 @@ The mapboxStyle method creates an Openlayers VectorTile layer and assigns a mapb
 @returns {layer} layer decorated with format methods.
 */
 export default async function mapboxStyle(layer) {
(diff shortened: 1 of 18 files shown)


dbauszus-glx:
The get method of the dictionary proxy isn't required to change. This should work the same as it does now and should not be an async method.

The dictionaries directory isn't required. The english dictionary is always required and can be assigned to the mapp.dictionaries object in the dictionary module.

The mergeLanguage isn't required to check whether a dictionary has already been merged. A user may want to merge a different dictionary for the same language.

The mergeLanguage method should be called in the default script. I reckon after the userLocale is merged.

dbauszus-glx:
Will close in favour of this PR.

https://github.com/GEOLYTIX/xyz/pull/2213

PR #2180, opened 2025-05-29
(bug): Non cached file templates should not be merged on each fetch

dbauszus-glx:
The cacheTemplate method should not merge the response into a template but assign the object.

This prevents arrays to increase in length when fetched multiple times from a file src.


simon-leech (approved):
This seems to be fine for me across a couple of instances 👍

PR #2184, opened 2025-05-29
(feat): Histogram utility

cityremade:
Adds visual element for histogram layer data. 
Documentation in the module ui/utils/histogram

Screenshot 2025-05-29 at 15 16 17


lib/ui/utils/histogram.mjs
```diff
@@ -0,0 +1,138 @@
...
+    return;
+  }
+
+  const queryparams = mapp.utils.queryParams(params);
+
+  const paramString = mapp.utils.paramString(queryparams);
+
+  params.host ??= params.layer?.mapview?.host;
+
+  const response = await mapp.utils.xhr(
```
simon-leech, line 133: Can we add a check on if response is null or instanceOf error. We need to handle if the query fails or returns null and perhaps alert or console log?

lib/ui/utils/histogram.mjs
```diff
@@ -0,0 +1,138 @@
...
+  if (params.queryparams.field) {
+    console.warn(`Histogram not created: field missing`);
+    return;
+  }
+
+  const queryparams = mapp.utils.queryParams(params);
+
+  const paramString = mapp.utils.paramString(queryparams);
+
+  params.host ??= params.layer?.mapview?.host;
```
simon-leech, line 131: Can `params.host` not be nullish assigned to just `mapp.host` instead of params.layer. `mapp.host` should always be present and its possible `params.layer` may not be if calling from an entry or custom view.

lib/ui/utils/histogram.mjs
```diff
@@ -0,0 +1,138 @@
...
+  @param {Object} params Histogram configuration object.
+  @property {string} params.dataview The type of dataview, expected value is `histogram`.
+  @property {HTMLElement} params.target Target element in which histogram is rendered.
+  @property {String} params.query Histogram query. Defaults to `histogram` if not defined. Histogram query requires `table` and `field` passed in queryparams object.
+  @property {Boolean} params.tooltip - defaults to false. If enabled each bucket will have a tooltip with bucket properties.
+  */
+async function create(params) {
+  const response = await requestData(params);
+
+  if (!response) return;
```
simon-leech, line 46: I think you perhaps could handle the `!response` within the `requestData` function - it would avoid duplication in the create and update method?

lib/ui/utils/histogram.mjs
```diff
@@ -0,0 +1,138 @@
...
+
+const histogram = {
+  create,
+  update,
+};
+
+export default histogram;
+
+/*
+// Definition of a histogram dataview within layer
```
simon-leech, line 17: Can you add to the docs here how to call it using the utility too ie and what needs to be provided such as where to output it etc?
``` js 
mapp.ui.utils.histogram(xx) 
```

simon-leech (changes requested):
Looks good from an initial pass over it!  🎉

This PR should be to merge into `minor` not `patch` though - as this is a new feature not a bug. 

I added a few comments on my first pass of the code.

RobAndrewHurst:
@cityremade I think we are going to want to reopen this pr from a branch off the minor branch

Issue #1993, opened 2025-03-12
Dialogs and drawer elements


dbauszus-glx:
Dialog and drawer elements are very similar in structure.

A drawer must have a header; A dialog can have a header.

A drawer can be collapsed into its header. A dialog with a header can be collapsed into this [minimised].

While a drawer is anchored to a fixed location, a dialog can be positioned within the context of it's parent.

However a dialog must be called from somewhere. If not through the API [eg alert] there is usually a button which will reflect the status of the dialog [eg active when shown].

It should therefore be possible to represent a drawer as a button. For example the drawing panel could be shown as a button which opens the draw content in a dialog.

It should also be possible to add a dialog button to the header of a drawer. This would allow to turn a drawer into a dialog. If closed this dialog would be rendered as a drawer back into it's original location.

PR #2185, opened 2025-05-30
feat(drawer): Drawer dialog popout

AlexanderGeere:
## Description
drawers and dialogs are very similar, this PR adds a new option to the drawer element `popout`. 
The drawers that can be popped out are:
1. Layer
2. Filter
3. Dataviews
4. Drawing
5. Circle drawer
6. Reports
7. Style

This is achieved by specifying `popout: true` in the configuration, leading to:
Screenshot from 2025-05-30 11-22-03

## GitHub Issue
[#1993](https://github.com/GEOLYTIX/xyz/issues/1993)

## Type of Change
- ✅ Documentation

## How have you tested this?
Tested using `bugs_testing/dataviews/dataviews.json` on the scratch and supermarkets layer.
[Localhost Link](http://localhost:3000/latest/?layers=OSM,scratch&z=8.69&lat=51.2353&lng=-0.01562)

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/tests/lib/ui/layers/panels/dataviews.test.mjs
+++ b/tests/lib/ui/layers/panels/dataviews.test.mjs
@@ -2,7 +2,7 @@ import dataviews_panel from '../../../../assets/dataviews/dataviews_panel.json';
 export function dataviews(mapview) {
   codi.describe(
     {
-      name: 'Panel Filter test:',
+      name: 'dataview panel test:',
       id: 'ui_layers_panel_dataviews',
       parentId: 'ui_layers',
     },
@@ -19,6 +19,7 @@ export function dataviews(mapview) {
             mapview: mapview,
             dataviews: dataviews.dataviews,
             showCallbacks: [],
+            hideCallbacks: [],
           };
 
           const drawer = ui.layers.panels.dataviews(layer);
--- a/tests/lib/ui/layers/panels/filter.test.mjs
+++ b/tests/lib/ui/layers/panels/filter.test.mjs
@@ -55,25 +55,17 @@ export function filter() {
           const filterPanelDropDown = filterPanel.querySelector(
             '[data-id=panel_test-filter-dropdown]',
           );
-          const drop_down_elements = filterPanelDropDown.querySelector('ul');
+
+          const drop_down_elements =
+            filterPanelDropDown.querySelectorAll('option');
 
           codi.assertEqual(
-            drop_down_elements.children.length,
-            2,
-            'We expect two entries into the dropdown from the infoj',
+            drop_down_elements.length,
+            3,
+            'We expect three entries into the dropdown from the infoj',
           );
 
-          filterPanel
-            .querySelector('ul')
-            .children[0].dispatchEvent(new Event('click'));
-
-          codi.assertEqual(
-            filterPanel
-              .querySelector('ul')
-              .children[0].classList.contains('selected'),
-            true,
-            'Expect an element to be selected',
-          );
+          filterPanelDropDown[1].dispatchEvent(new Event('click'));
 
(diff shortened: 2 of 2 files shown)


cityremade:
Some questions from me
- so at the moment multiple popped out containers overlap. Should they cascade? 
- it's unclear where the floater belongs when it's popped out. Elements belonging to the same layer or different layers - how to tell that?
- should they be grouped together by some some sort of attribute that would handle them as belonging to the same layer?
- do we want to keep crowding headers? Maybe some options would be better off within the container.

RobAndrewHurst:
> Some questions from me
> 
> * so at the moment multiple popped out containers overlap. Should they cascade?
> * it's unclear where the floater belongs when it's popped out. Elements belonging to the same layer or different layers - how to tell that?
> * should they be grouped together by some some sort of attribute that would handle them as belonging to the same layer?
> * do we want to keep crowding headers? Maybe some options would be better off within the container.

Great points, yeah we need to figure out a better way to layout the dialogs

RobAndrewHurst:
> > Some questions from me
> > 
> > * so at the moment multiple popped out containers overlap. Should they cascade?
> > * it's unclear where the floater belongs when it's popped out. Elements belonging to the same layer or different layers - how to tell that?
> > * should they be grouped together by some some sort of attribute that would handle them as belonging to the same layer?
> > * do we want to keep crowding headers? Maybe some options would be better off within the container.
> 
> Great points, yeah we need to figure out a better way to layout the dialogs

Re this don't do this yet. This can come in a future change

dbauszus-glx (dismissed):
The popout flag should be part of the viewConfig.

```js
    layer.drawer = mapp.ui.elements.drawer({
      data_id: layer.key,
      class: layer.viewConfig.classList,
      header,
      content: layer.viewConfig.content,
      popout: layer.popout,
    });
```

eg

```json
        "viewConfig": {
          "popout": true
        },
```

The popoutDialog method is only ever called from the ui/elements/drawer module [method]. The popoutDialog should be added to that module.

https://github.com/GEOLYTIX/xyz/pull/2185/commits/6ee1219835672a161bc680f31a20ca605f705576

Can the target for the dialog be the document instead of just the mapview? This would prevent this kind of issue?

image

Otherwise the extent of the dialog will need to be respected not just the centre.

@cityremade The positioning of multiple dialogs is out of scope for this PR. This should be addressed in a separate ticket for the dialog method.

_it's unclear where the floater belongs when it's popped out. Elements belonging to the same layer or different layers - how to tell that?_

We could define a string as header instead of a true flag for the popout property.

_should they be grouped together by some some sort of attribute that would handle 
them as belonging to the same layer?_

No. This is too complex.

_do we want to keep crowding headers? Maybe some options would be better off within the container._

No. There is a limited screen property. We not gonna start moving header button into drawer, and then stuff from the drawer/dialog somewhere else because this is getting crowded. This is an optional functionality, not default. This shouldn't be made available if to crowded or not required. Adding header icons to the content will look weird.

AlexanderGeere:
@dbauszus-glx  didnt need to change the target, was just missing the contained parameter.

dbauszus-glx (changes requested):
We are almost there...

image

The drawer in the layer view dialog should have the same css rules as the drawers in drawer layerview.

image


I think all that is required is the layer view class to be applied on the container.

AlexanderGeere:
@dbauszus-glx adding the classes didnt work, so i added some css to the `.popout` instead to match it to the layer-view. The other popouts look good to me.

dbauszus-glx (approved):
I am happy with the way this looks and works now.

PR #2186, opened 2025-06-02
Histogram utility minor

cityremade:
Adds visual element for histogram layer data. 
Documentation in the module ui/utils/histogram

Screenshot 2025-05-29 at 15 16 17

### DIFF
--- a/lib/ui/utils/_utils.mjs
+++ b/lib/ui/utils/_utils.mjs
@@ -49,6 +49,7 @@ function Chart(_this) {
 import * as cssColour from './cssColourTheme.mjs';
 import dataview from './dataview.mjs';
 import dataviewDialog from './dataviewDialog.mjs';
+import histogram from './histogram.mjs';
 import idleLogout from './idleLogout.mjs';
 import imagePreview from './imagePreview.mjs';
 
@@ -62,6 +63,7 @@ export default {
   cssColour,
   dataview,
   dataviewDialog,
+  histogram,
   idleLogout,
   imagePreview,
   Json: jsonDataview,
--- a/lib/ui/utils/histogram.mjs
+++ b/lib/ui/utils/histogram.mjs
@@ -124,12 +124,12 @@ function createHistogram(params, data) {
 async function requestData(params) {
   params.query ??= 'histogram';
 
-  if (params.queryparams.table) {
+  if (!params.queryparams.table) {
     console.warn(`Histogram not created: table missing`);
     return;
   }
 
-  if (params.queryparams.field) {
+  if (!params.queryparams.field) {
     console.warn(`Histogram not created: field missing`);
     return;
   }
--- a/lib/ui/utils/histogram.mjs
+++ b/lib/ui/utils/histogram.mjs
@@ -1,0 +1,148 @@
+/**
+## /ui/utils/histogram
+
+The histogram ui utility
+
+@module /ui/utils/histogram
+
+Accessible through 
+```js
+mapp.ui.utils.histogram.create(params);
+mapp.ui.utils.histogram.update(params);
+```
+*/
+
+const histogram = {
+  create,
+  update,
+};
+
+export default histogram;
+
+/*
+// Definition of a histogram dataview within layer
+```json
+"dataviews": {
+    "Histogram": {
+        "dataview": "histogram",
+        "tooltip": true,
+        "queryparams": {
+            "field": "population",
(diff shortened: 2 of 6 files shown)


AlexanderGeere:
@cityremade 
This is looking good, i just had a couple of questions.
1. How difficult would it be to label the axis and maybe title the histogram as well? 
2. Could we adjust the css for the tabview, it spreads out really wide:
Screenshot from 2025-06-03 09-54-41

I do appreciate if these are out of scope for this.

cityremade:
New layout options: title, caption, width, height, xlabel, ylabel, tooltip now in options too.

Screenshot 2025-06-03 at 12 20 43
Screenshot 2025-06-03 at 14 49 34

AlexanderGeere:
@cityremade I added a nullish asssign to the options, as you can provide no options just get a histogram.

lib/ui/utils/_utils.mjs
```diff
@@ -49,6 +49,7 @@ function Chart(_this) {
 import * as cssColour from './cssColourTheme.mjs';
 import dataview from './dataview.mjs';
 import dataviewDialog from './dataviewDialog.mjs';
+import histogram from './histogram.mjs';
```
RobAndrewHurst, line 52: This confuses me a bit. Shouldn't this be part of a dataview not something separate?

dbauszus-glx: This is correct. The histogram should work like the jsonDataview which assigned as a ui.utils. However I wasn't able to make this work.

dbauszus-glx (changes requested):
The histogram module should work like a dataview module. eg. the jsonDataview or a chartjs dataview.

This sample configuration should work.

```json
        "dataviews": {
          "histogram": {
            "dataview": "Json",
            "query": "histogram",
            "target": "tabview",
            "viewport": true,
            "mapChange": true,
            "queryparams": {
              "table": true,
              "field": "pop_20",
              "filter": true,
              "buckets": 5
            }
          },
          "histogram_chart": {
            "dataview": "chartjs",
            "query": "histogram",
            "target": "tabview",
            "viewport": true,
            "mapChange": true,
            "queryparams": {
              "table": true,
              "field": "pop_20",
              "filter": true,
              "buckets": 5,
              "chartjs": true
            },
            "chart": {
              "type": "bar"
            }
          },
          "histogram_dataview": {
            "dataview": "histogram",
            "query": "histogram",
            "target": "tabview",
            "viewport": true,
            "mapChange": true,
            "queryparams": {
              "table": true,
              "field": "pop_20",
              "filter": true,
              "buckets": 5,
              "chartjs": true
            }
          }
        }
```

The histogram.create should not request the data but assign a setData method which will rerender the histogram element.

The dataview update method will request the data and call the setData method.

cityremade:
requestData and update methods removed.
setData now re/creates visual element.

dbauszus-glx (approved):
I updated the documentation to include a sample data array.

image

I also refactor the createHistogramBuckets method.

Replacing the children and then creating multiple nodes, appending each individually are a lot of unnecessary calls to the web api. It is more efficient to map an array of html fragments and then render these into the node, which replaces the children in the node. This is one call to the  web api, no matter the number of buckets.

```js
function createHistogramBuckets(params) {
  const maxBucketHeight = Math.max(...params.data.map((item) => item.count));

  params.options.height ??= 100;

  const bucketElements = params.data.map((bucket) => {
    const style = `height: ${Math.round((params.options.height * bucket.count) / maxBucketHeight)}px`;

    const tooltip = params.options.tooltip
      ? mapp.utils.html`
          <div class="bucket-info">
          <span><b>Count</b> ${bucket.count}</span><br>
          <span><b>Min</b> ${bucket.bucket_min}</span><br>
          <span><b>Max</b> ${bucket.bucket_max}</span>`
      : ``;

    return mapp.utils.html`<div 
      class="bucket"
      data-bucket=${bucket.bucket}
      data-count=${bucket.count}
      data-bucket-min=${bucket.bucket_min}
      data-bucket-max=${bucket.bucket_max}
      style=${style}>
      ${tooltip}`;
  });

  const ylabel = params.options.ylabel
    ? mapp.utils.html`<div class="ylabel">${params.options.ylabel}`
    : '';

  mapp.utils.render(
    params.node,
    mapp.utils.html.node`${bucketElements}${ylabel}`,
  );
}
```

RobAndrewHurst (approved):

PR #2196, opened 2025-06-11
fix(vector): Feature count returns when array is empty

cityremade:
In vector.mjs when feature count needed for clustering is called some feature arrays are empty and the first array entry features[0] doesn't exist.

Adds an earlier check on array length.


simon-leech (approved):
This looks fine to me - just an early return to prevent a crash later down

dbauszus-glx:
This is a more serious issue which I want to address with an OL test case.

PR #2198, opened 2025-06-11
fix(plugins): Missing data-ids

simon-leech:
Added data-ids for a few buttons that were missing them for ease of selection / analytics.


AlexanderGeere (approved):

dbauszus-glx (changes requested):
The dataid should also be added to the zoomBtn and the id removed. The id shouldn't be required on the zoomBtn as they are unique to a mapview but not necessarily to a document view like any other plugin / mapview control element.

simon-leech:
@dbauszus-glx  Swapped the zoom in and zoom out buttons to data-id too. Ready for rereview

dbauszus-glx (approved):

PR #2200, opened 2025-06-12
fix(vector): Cluster source change event method fails with empty features

dbauszus-glx:
This PR creates the clusterSourceChange event method which is called by changes to the cluster source. Previously the logic was tied to the layer change event which caused execution on every render, eg when hovering over a cluster item. This also caused an issue where the method would be called when a layer is added to the map without any features added to the source.

The method will shortcircuit if the cluster source has no features. This should only happen if [all] features are removed from the source, eg. the source is cleared.

The cluster features map is throwing an error if a cluster feature has no features. This is likely to indicate invalid geometries in the vector source, eg. mixed projections.

### DIFF
--- a/lib/layer/featureFormats.mjs
+++ b/lib/layer/featureFormats.mjs
@@ -149,24 +149,25 @@ The cluster function processes an array of feature data and returns an array of 
 
 It creates Point geometries from coordinate pairs, extracts count and ID information, and assigns additional properties based on the layer's field definitions.
 
-The function also tracks the maximum count across all features in the layer.max_size property.
+The function also tracks the maximum count across all features in the layer.cluster.max_size property.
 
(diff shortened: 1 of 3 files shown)


simon-leech:
@dbauszus-glx Should this PR be merging into minor not main?

RobAndrewHurst:
> @dbauszus-glx Should this PR be merging into minor not main?

Not patch? Because it a bug fix?

dbauszus-glx:
@RobAndrewHurst @simon-leech Yeah. This should be a patch. Odd.

cityremade (approved):
looks good, I like async feature request separated from source definition.

simon-leech (approved):
All seems to be working fine for me!

AlexanderGeere (approved):

PR #2203, opened 2025-06-16
feat(util): scriptElement utils module and mapboxStyle format

dbauszus-glx:
The PR adds a utility method to load a script by adding a script element tag to the document head.

The load process will be assigned to a promises object to prevent the same script [src] to be loaded multiple times.

The utility is required to load the mapboxStyle plugin.

The import of the mapboxStyle plugin will throw an exception if Openlayers is not loaded. This will make it impossible to use the mapp library without Openlayers.


cityremade (approved):
no issues spotted.

PR #2209, opened 2025-06-18
fix(layer.entries) - Queryparams from layer and locale must be assigned to entry

simon-leech:
Queryparams from the `layer` and `locale` must be assigned to the `type:layer` entry. This is the identical format to how entry dataviews work and prevents issue where a parameter is defined just once in the `locale` - but is not accessible for `layer` entry queries. 

To test this - simply add queryparams to the `locale` and/or `layer`. 
E.g. 
``` json
"locale": {
"queryparams": {
  "testme": 123
 }
}
```
And check that your type layer entry query contains them in the paramstring

### DIFF
--- a/lib/ui/locations/entries/layer.mjs
+++ b/lib/ui/locations/entries/layer.mjs
@@ -227,6 +227,13 @@ async function showLayer() {
   const entry = this;
 
   if (entry.query) {
+    // Assign queryparams from layer, locale.
+    entry.queryparams = {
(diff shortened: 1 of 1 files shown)


simon-leech:
> The issue of where the locale.queryparams are assigned should be addressed first before adding the assignment of locale.queryparams in more places.

Discussed this one and we want to address the locale.queryparams in a separate PR. This PR is simply to address the bug first.

PR #2213, opened 2025-06-18
Move dictionaries to public; Loaded with utils.

dbauszus-glx:
tbc


simon-leech:
@dbauszus-glx  Just one quick comment before I test properly - is there a reason the english is still an `.mjs` file while all the rest are `.json` ?

dbauszus-glx:
@simon-leech Good the question. The reason being that the english dictionary is imported into the bundle.

dbauszus-glx:
I added a check to the mapview method to load a dictionary if not already loaded.

AlexanderGeere (approved):
Checked a few plugins, default view. Working as before

RobAndrewHurst (approved):

PR #2216, opened 2025-06-19
security(api): remove user controlled data

RobAndrewHurst:
# api user controlled data

```js
if (req.params instanceof Error) {
  return res.status(400).send(req.params.message);
}
```

This current code reflects the error message directly to the client, which could potentially expose internal implementation details.

```js
if (req.params instanceof Error) {
  return res.status(400).send("Invalid request parameters.");
}
```

**What does this change do?**:

- **Prevents Information Disclosure**: The original code sent the actual error message from `validateRequestParams()`, which could expose internal validation logic details to potential attackers.
- **Eliminates Reflection Attacks**: By not reflecting user-controlled data back to the client, we will prevent potential XSS or other injection attacks.
- **Consistent Error Handling**: Users now receive a consistent, generic error message that doesn't reveal the specific validation that failed.
- **Maintains Functionality**: The HTTP status code (400) is preserved, so legitimate clients can still handle the error appropriately.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -113,7 +113,10 @@ export default function api(req, res) {
   req.params = validateRequestParams(req);
 
   if (req.params instanceof Error) {
-    return res.status(400).set('Content-Type', 'text/plain').send(req.params.message);
+    return res
+      .status(400)
+      .set('Content-Type', 'text/plain')
+      .send(req.params.message);
   }
 
   if (req.params.logout) {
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):

dbauszus-glx:
I'd like to retain the error message. I believe the issue being that the response is not sanitized. This can be done by setting the content-type.

@RobAndrewHurst How de we know whether this resolves the sonar cloud flag?

PR #2218, opened 2025-06-23
Bump Openlayers; Scriptelement type param.

dbauszus-glx:
This PR bumps openlayers to from version 10.3.1 to version 10.6 which has recently been released.

https://github.com/openlayers/openlayers/releases/tag/v10.6.0

The olScript should be a single line function using the scriptElement utility module introduced in this minor release.

The scriptElement method should log which script is loaded and accept an optional param to specifiy the script tag. `application/javascript' is required for the OL script.


dbauszus-glx:
@cityremade The version is not set but checked in the mapp module. The assignment of the _ol const as ol property is weird. I changed this and updated the module documentation to clarify the intent and purpose.

cityremade:
and this ?
`if (parseFloat(ol?.util.VERSION) < 10.6) {`

dbauszus-glx:
@cityremade This is required for custom views. Some functionality maybe dependent on the current tested version which is loaded by the default view. A custom view may reference an openlayers version which is a couple of years old. The mapp library will load and work but all kinds of weird may happen. Hence we warn in the mapp module if the loaded OL version is below the current version against which we test.

simon-leech:
Please can we keep the console warning that would output both the current version loaded in the custom view, and also the current MAPP OL version. This is really helpful for configuration, as it makes it easy to see what version you need to update to. We don't want to lose this.

dbauszus-glx:
@simon-leech current version is logged with loaded version.

AlexanderGeere (changes requested):
this check 
```javascript
   else if (parseFloat(ol?.util.VERSION) < ol_current) {
          console.warn(`Openlayers v${ol?.util.VERSION} below current v${ol_current}.`);
    }
```

Fails, because `parseFloat('10.6.1')` yields 10.6 which is less than 10.61. 

we could do something like this:
```javascript
let ol_version = ol?.util.VERSION;
ol_version = ol_version
  ? parseFloat(
      ol_version.substring(0, ol_version.indexOf('.') + 1) +
        ol_version
          .substring(ol_version.indexOf('.') + 1)
          .split('.')
          .join(''),
    )
  : ol_version;
```

which will change x.x.x into x.xx or x.x.x.x into x.xxx, etc.

dbauszus-glx:
@AlexanderGeere I truncate the patch version. This shouldn't matter since there will never be a feature compatibility issue in a patch release.

simon-leech (approved):
Fine by me and see the warning correctly in custom view

PR #2228, opened 2025-06-24
feature(layer): JSON Editor layer panel and element

KieronFerrey:
This PR adds a JSON editor panel method which calls a layer json editor element


AlexanderGeere:
we probably need to move renderMenu out into a util that can be used where its needed to fix the code duplication

dbauszus-glx:
This PR adds the `layer.panels.jsonEditor` method which allows to update a layer with the json content of the JSE element in the dialog which can be toggled from the button returned from the panel method.

The `ui.utils.layerJSE` method has been added to reduce duplicate code. This method is called from the `userLayer` plugin to add new layers.

`update` and `remove` methods have been added to the layer decorator. The layer.remove() method will be called from the `mapview.removeLayer` method.

A jse css has been to allow for text button in custom JSE menus.

image

RobAndrewHurst (changes requested):
How is this supposed to handle a user giving a duplicate 'key'?

Given the nature of the `jsonEditor` it takes pretty much anything but what could we do to safe guard against this??

dbauszus-glx:
@RobAndrewHurst There are no safeguards as this is meant to be a development tool to test such things as duplicate keys, changed keys, removed or added infoj entries, changed or added themes, etc.

Issue #1469, opened 2024-09-06
location.flyTo extents, tests, and documentation


cityremade:
Spotted behaviour:

Onload: selected location has a **geojson catchment which is turned off**

`zoom to feature bounds` flies to the point - location extent is a point at this time  - as expected

Catchment is displayed from checkbox 

`zoom to feature bounds` flies to the extent of the catchment - as expected

Catchment is hidden from checkbox 

`zoom to feature bounds` still flies to the extent of the catchment - surprising

This means that once a geojson catchment is displayed it alters location extent and its visibility doesn't update location extent anymore. 

When catchment is an mvt_clone it doesn't change location extent - `zoom to feature bounds` flies to the point at all times, regardless of initial value for display flag  - again surprising. Should fly to mvt clone extent.

Is this the desired behaviour?

dbauszus-glx:
MVT_clones are about to be retired due to a multitude of issues.

Can I get a JSON location to replicate the issue with the location.flyTo() method?

dbauszus-glx:
The issue to resolve the layer extent must be completed first. https://github.com/GEOLYTIX/xyz/issues/1473

dbauszus-glx:
The vector_layer entry must be consolidated with the layer entry first before the location.flyTo extent can be updated. https://github.com/GEOLYTIX/xyz/issues/1586

AlexanderGeere:
We are going to add a getExtent function to the Pin, geometry, layer and location. These functions will be added to an array which will be iterated by flyTo to get the largest extent.

PR #2235, opened 2025-06-26
feature(flyTo): Fly to the extent of mvt entry type layers with featureSet/Lookup

AlexanderGeere:
## Description
This PR makes it possible to use the `flyTo` to fly to the greater extent of a location which uses an entry type layer, and featureSet or featureLookup. 

Previously `flyTo` would zoom in only on the location itself:
Screenshot from 2025-06-27 11-47-48

On the PR it should now flyTo the outer extent:
Screenshot from 2025-06-27 11-49-06

This is done by adding a `getExtent` function to the pin, geometry and layer entry. These functions are added to an Extents object on the location, when the entry is shown. The new `createExtent` function will loop over this object and create an extent which contains all the extens present in location.Extents.

`flyTo` calls `createExtent` with the location.Extents object. 

## GitHub Issue
#1469

## Type of Change
- ✅ Documentation

## How have you tested this?
Tested locally using `bugs_testing/infoj/layer_entry.json` on the points layer.
[Test Location](http://localhost:3000/latest/?layers=OSM,uk_retail_places,points&locations=points!4&z=10.03&lat=51.54777&lng=-0.05503)

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine
- ✅ Updated existing tests
- ✅ Added new tests

## Code Quality Checklist


dbauszus-glx (dismissed):
I don't understand the change to the feature_extent query template. Why is this required? Doesn't this mess zoomToExtent method in the layer decorate module?

Having a getFeatureExtent method with duplicate code to the zoomToExtent method in the location layer entry should be avoided.

The location layer entry method uses the layer decorator. The getExtent method should be added to the layer decorator. The zoomToExtent method can then call the internal getExtent method.

I'd probably split the location flyTo method. Have a location getExtent method called from the flyTo method.

What is the `values_` property of the entry.L object in the location.layers array. This looks like an internal OL property which should not be accessed directly.

I'd probably rethink the use of the location.Layers array. The idea behind this array is to provide a way to access the extent of multiple layer and to ensure that the layer is removed when the location is removed. This predates the removeCallbacks array which can be used to remove the Openlayers layer from the mapview.

It is probably easier to rethink the getExtent for geometry / pin entries than having a complex condition in the flyTo / getExtent methods for a location.

AlexanderGeere:
@dbauszus-glx I have added a  `getExtent` to the pin, geometry and layer entries. I have also added an Extents property to the location.

Extent functions are added to the Extents parameter when an entry is shown and removed when it is hidden. 

When clicking `flyTo`, `createExtent` is called with the location.Extents. Create extent returns an extent that includes all those from location.Extents. 

On the pin entry `getExtent` function, a buffer can be applied to the pin by either supplying `entry.buffer: true` or `entry.buffer: 500` for instance. supplying true will default the buffer to 200.

dbauszus-glx (changes requested):
The zoom to location breaks on a vector layer.

```
          {
            "type": "layer",
            "layer": "points"
          }
```

The points layer is a WKT vector layer.

image

layer.mjs:264 Uncaught (in promise) TypeError: e.getExtent is not a function
    at location.Extents.<computed> [as points-327] (layer.mjs:264:63)
    at Object._e [as createExtent] (createExtent.mjs:17:61)
    at async Object.un [as flyTo] (decorate.mjs:292:23)

AlexanderGeere:
@dbauszus-glx  the getExtent function is now always asssigned to the layer entry and teh check on featureSet or lookup is done in the function

dbauszus-glx:
The getExtent method should be assigned to the layer in the decorator. The layer will be this in the method.

The getFeatureExtent query should return values only.

I have also updated the conditions inside the getExtent method in regards to the check for the featureLookup, featureSet, or else.

https://github.com/GEOLYTIX/xyz/pull/2235/commits/e05e2c4186784a87f022b48e5ab1f6dc28d13cda

dbauszus-glx:
It should be possible to assign a table property in order to control the zoom visibility.

```json
{
  "type": "layer",
  "layer": "oa",
  "table": "geodata.uk_glx_geodata_oa_metrics_2021",
}
```

https://github.com/GEOLYTIX/xyz/pull/2235/commits/3c6961268a8cc16551a11c441a8689b0b99d1afd

dbauszus-glx:
The location.extents should not be processed in the layer.show method assigned in the layer decorator module. The layer entry module assigns it's own showLayer method. https://github.com/GEOLYTIX/xyz/pull/2235/commits/38c8c1f24ef29834072fe92b0b554d8469c2b113

dbauszus-glx:
There was quite a bit of duplicate code and some broken SQL.

The layer_extent SQL could not work due to a missing SELECT in the nested query and a surplus semi-colon.
image

Having a feature_extent and a layer_extent query which are identical apart from the body condition is too much duplicate code.

This can be solved by creating the condition string inside the template in regards to the body with an id array being available.

The query method still needs some documentation.

```js
export default (_) => {

  let condition = 'true'

  if (Array.isArray(_.body?.ids)) {
    const ids = _.body.ids.filter((id) => /^[A-Za-z0-9,"'._-\s]*$/.test(id));

    condition = `\${qID} IN ('${ids.join("','")}')`
  }

  return `
    SELECT
      ARRAY[
        st_xmin(box.box2d),
        st_ymin(box.box2d),
        st_xmax(box.box2d),
        st_ymax(box.box2d)]
    FROM(
      SELECT Box2D(
        ST_Transform(
          ST_SetSRID(
            ST_Extent(\${geom}),
            \${proj}),
          \${srid}))
      FROM \${table}
      WHERE ${condition} \${filter}) box`
};
```

The query has been taken out from the zoomToExtent method. The getExtent method should now return the Extent for any type of layer with one query. The body is provided as undefined if there are no ids.

```js
async function getExtent() {
  const layer = this;

  let ids = [];
  if (Array.isArray(layer.featureLookup)) {
    // Get array of ids from lookup features.
    ids = layer.featureLookup.map(
      (feature) => feature[layer.featureLookupId || 'id'],
    );
  } else if (Array.isArray(layer.featureSet)) {
    ids = [...layer.featureSet];
  } else if (layer.L && layer.format !== 'mvt') {
    // return extent from Openlayers source.
    return layer.L.getSource().getExtent();
  }

  const body = ids.length? JSON.stringify({
      ids: ids,
    }) : undefined;

  const response = await mapp.utils.xhr({
    url:
      `${layer.mapview.host}/api/query/layer_extent?` +
      mapp.utils.paramString({
        filter: layer.filter?.current,
        geom: layer.geomCurrent(),
        layer: layer.source_key || layer.key,
        proj: layer.srid,
        srid: layer.mapview.srid,
        table:
          layer.table ||
          Object.values(layer.tables)[0] ||
          Object.values(layer.tables)[1],
      }),
    body,
  });

  return response;
}
```

The feature_extent test is now breaking and needs to be updated. @AlexanderGeere

dbauszus-glx:
@AlexanderGeere Please check my last two commits.

https://github.com/GEOLYTIX/xyz/pull/2235/commits/05e42ffb81475efa2017938c22a68a6fe56f922a

Instead of an utility method to calculate the extents from an object of extents it is easier to have a location.getExtent method which calculates it's extent.

The getExtent method of pin, geometry, and layer entry should be assigned as a method where `this` is the entry. The entry does not have to be provided to this methods.

It is much easier to assign the getExtent method once rather than to assign and reassign in the show hide methods.

https://github.com/GEOLYTIX/xyz/pull/2235/commits/858456a66f69b7c203b085685888a076dd66307d

The location.Extents object is no longer required. The layer [entry] have each show, hide, getExtent methods and can be pushed into the Layers array of the location. You can now easily iterate over that array to call the remove method of each entry, or the hide method of each entry.

simon-leech (dismissed):
This is looking really good! 
A few little issues:
1. Zooming to the extent of just a `pin` entry, isn't working, it doesn't zoom in. 
2. Calling `flyTo()` in a custom view on a location is not doing anything. I think this may be to do with the flyTo requiring a `maxZoom` but we never previously had to define this, it had a default.

AlexanderGeere:
Added a entry.display ??= true on the pin, without this the flyTo does not fly to the pin if it is the only geometry on the layer.

AlexanderGeere:
@simon-leech I am not able to replicate the flyTo not working on custom views.

cityremade:
I noticed a behaviour on styled catchments when at the end of zoom tiles are not recreated. How does it work when map zoom is finished and mvt tiles need to be refreshed for the final zoom level?

cityremade:
feature lookup catchments for reference here bugs_testing/infoj/styled_catchments_with_feature_lookup.json

PR #2240, opened 2025-06-30
Graduated Tests

KieronFerrey:
This adds graduated theme tests to check the structure of the theme legend


KieronFerrey:
Added in additional categorized tests

PR #2241, opened 2025-07-01
feature(roles): Check for dot notation [nested] roles.

dbauszus-glx:
This PR introduces checks on dot notation [nested] roles.

A dot notation role is written out like so: `Europe.UK`.

The roles endpoint previously scrubbed the dot notation roles from the workspaces/roles endpoint in favour of a roles object tree returned with the `tree=true` params flag. This is not necessary since the tree can be constructed from parsing the dot notation roles like so:

```js
  const rolesTree = {};

  // Delete restricted Asterisk role.
  rolesSet.delete('*');

  rolesSet.forEach((role) => {
    const roles = role.split('.');

    if (roles.length > 1) {
      roles.reduce(
        (accumulator, currentValue) => (accumulator[currentValue] ??= {}),
        rolesTree,
      );

      // Pop last role from array into roleSet.
      rolesSet.add(roles.pop());
    } else {
      rolesTree[role] ??= {};
    }
  });
```

The first part of a dot notation role is shifted from role. eg. `Europe` from `Europe.UK`. Users with the `Europe.UK` role have therefore the `Europe` role for ressource/template role checks.

The last part of a dot notation role is popped out in the roles.check() method. Users with the UK role only have access to the `Europe.UK` template for legacy support.

This PR also adds the asteriks [*] role to the user object in auth module rather than in individual workspace module methods.

A user with the role `Europe` has access to `Europe` but not `Europe.UK` or `UK`.

A user with the role `UK` has access to `UK` or `Europe.UK` but not `Europe`.

A user with the role `Europe.UK` has access to `Europe` and `Europe.UK` but not `UK`.

### DIFF
--- a/lib/ui/layers/legends/categorized.mjs
+++ b/lib/ui/layers/legends/categorized.mjs
@@ -51,7 +51,11 @@ export default function categorizedTheme(layer) {
       layer.filter?.current[cat.field]?.ni?.indexOf(cat.value) >= 0;
 
     if (layer.featureFields && theme.distribution === 'count') {
-      cat.count = layer.featureFields[cat.field]?.[cat.value];
+      // Build a params object to pass to the numericFormatter.
+      const params = {
+        value: layer.featureFields[cat.field]?.[cat.value],
+      };
+      cat.count = mapp.utils.formatNumericValue(params);
       if (!cat.disabled && !cat.count) return;
     }
 


RobAndrewHurst (approved):

simon-leech:
> I have a built a test workspace whereby I have this set up
> 
> 1. Europe locale (restricted to Europe role)
> 2. UK locale (restricted to UK role)
> 3. Default locale (merged into locales)
> 
> image
> I shouldn't see UK in the dropdown and be able to access it if I have the Europe role - I can swap to the UK but then get a 'no locales accessible' pop up.
> 
> Review ongoing so will keep popping comments as I go

This is actually a bug off the PR too so unrelated. 

I am happy this is working as expected so approving!

PR #2255, opened 2025-07-07
layer hideCallback must conditionally check for display of the datavi…

simon-leech:
layer hideCallback must conditionally check for display of the dataview before calling hide

This prevents an error whereby an error is thrown that `dataview.remove() is undefined`.


AlexanderGeere (approved):

dbauszus-glx (approved):
I added a comment in this regard

PR #2256, opened 2025-07-07
feature(workspace): getCachedWorkspace Function Implementation

RobAndrewHurst:
This PR adds in the function`getCachedWorkspace`, which serves as a critical workspace pre-loading mechanism for testing and administrative operations.
This function performs comprehensive workspace caching/hydration.

## What `getCachedWorkspace` Does

The `getCachedWorkspace` function performs a **complete workspace cache(hydration)** by:

1. **Base Workspace Loading**: Gets the workspace from cache via `workspaceCache(options.force)`
2. **Locale Pre-loading**: Iterates through all workspace locales and loads each one with user context
3. **Layer Pre-loading**: For each locale, loads every layer with proper user permission checking
4. **Template Pre-loading**: Ensures all workspace templates are loaded and cached via `getTemplate()`

## what's it doing?

- **Eager Loading**: Unlike standard workspace operations that load resources on-demand, this function pre-loads everything
- **Permission Context**: Applies user permissions during the loading process, filtering out inaccessible resources
- **Template Hydration**: Ensures all templates are resolved and cached in memory
- **Error Resilience**: Continues processing even if individual locales/layers fail to load

## how is it being used?

**1. Testing** (`test` function):

```javascript
workspace = await getCachedWorkspace({ force: true, user: req.params.user });
```

- used to ensure complete workspace is available for comprehensive testing
- forces cache refresh to get latest workspace state
- loads everything with admin user context

**2. Roles** (`roles` function):

```javascript
await getCachedWorkspace({
  force: req.params.force,
  user: req.params.user,
});
```

- Used when forcing workspace refresh for role extraction
- Ensures all workspace components are loaded before role analysis

### DIFF
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -561,8 +561,6 @@ async function getCachedWorkspace(options) {
         user: req.params.user,
       });
 
-      console.log('test');
-
       locale.layers[layerKey] = layer;
     }
   }
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -86,11 +86,15 @@ export default async function getKeyMethod(req, res) {
 @description
 The method requests a JSON layer from the getLayer module.
 
+The layer is checked for user role access and will return an error if access is denied.
+
+All role information is removed from the layer before being returned to the client.
+
 @param {req} req HTTP request.
 @param {res} res HTTP response.
 @property {Object} req.params HTTP request params.
 @property {string} [params.locale] Locale key.
-@property {boolean} [params.layer] Layer key.
+@property {string} params.layer Layer key.
 @property {Object} [params.user] User requesting the layer.
 
 @returns {res} The HTTP response with either an error.message or the JSON layer.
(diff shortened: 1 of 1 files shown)


dbauszus-glx (dismissed):
The getCachedWorkspace does not need to return the workspace which is assigned to the module worksapace variable as returned from the imported workspaceCache method.

The getCachedWorkspace will always set the force param in the call to the workspaceCache method.

https://github.com/GEOLYTIX/xyz/pull/2256/commits/c81f82bf431c152120d4be78ccc47d32d663228a

The getCachedWorkspace should be renamed to better indicate the use.

The user parameter should be optional for the getCachedWorkspace method. This method will only ever be called from methods which are only accessible to admins.

The getLayer and getLocale method need to be updated. The method needs to cache all layer and locales. Not just locales or layer accessible to the admin user calling the method.

RobAndrewHurst:
> The getCachedWorkspace does not need to return the workspace which is assigned to the module worksapace variable as returned from the imported workspaceCache method.
> 
> The getCachedWorkspace will always set the force param in the call to the workspaceCache method.
> 
> [c81f82b](https://github.com/GEOLYTIX/xyz/commit/c81f82bf431c152120d4be78ccc47d32d663228a)
> 
> The getCachedWorkspace should be renamed to better indicate the use.
> 
> The user parameter should be optional for the getCachedWorkspace method. This method will only ever be called from methods which are only accessible to admins.
> 
> The getLayer and getLocale method need to be updated. The method needs to cache all layer and locales. Not just locales or layer accessible to the admin user calling the method.

I need to return the workspace conditionally for testing purposes

dbauszus-glx:
@RobAndrewHurst You can run conditional checks in the test method. The cache method should cache the full workspace.

Lets say you have a workspace [a,b,c] and one to test a and b, but the admin user has only access to b and c, then the a ressource is not cached. roles nested in the a ressource will not be returned from the roles method. The roles and test methods should both cache the complete workspace in the 'getCachedWorkspace' method. 

image

Whether the workspace variable is returned at the of the method and assigned to itself in the test method shouldn't matter.

image

dbauszus-glx (changes requested):
These issues will have to be resolved first before I get to review this. https://github.com/GEOLYTIX/xyz/issues/2266

dbauszus-glx:
@RobAndrewHurst The details parameter request can not be removed from the _workspace/roles method as this may be required by custom admin views which query the workspace.roles object.

This should happen before the cacheTemplates request since the roles object is set on the workspace directly.

https://github.com/GEOLYTIX/xyz/pull/2256/commits/95e4d0d60a1451fc4f33a6e675b9759aaf4c5aa1

dbauszus-glx:
@RobAndrewHurst 

Is there ever a need to not force cache the workspace in the cacheTemplates method?

https://github.com/GEOLYTIX/xyz/pull/2256/commits/a13c5ab3313f9dd758ca6c9d3155150225cc8b6c

dbauszus-glx:
Having had more time to test this the roles request probably shouldn't force clearing the cache to prevent potentially hundreds of templates being cached over and over again in repeated calls to the roles method from the admin panel.

AlexanderGeere (approved):

simon-leech (changes requested):
Perfectly happy to leave you guys to review this one! 
But just noticed that there's a sonarcloud warning about complexity, can someone have a look at that?

dbauszus-glx:
I am not too worried about the complexity it's 16/15 and working fine. Not sure I want to faff around with the method script making it more complex in order to pass a complexity check.

dbauszus-glx:
I have resolved the conflicts which were documentation related only.

dbauszus-glx:
@RobAndrewHurst 

The initializeTestConfig method doesn't have any conditional complexity. This method can stay inside the test method to not have to move through methods which require additional documentation.

testWorkspaceLocales does not have to be an async method since the method only has a for loop which will be processed before the function completes and returns to the calling closure.

testLocaleLayers is tiny can be processed within testWorkspaceLocales.  The method doesn't have to be async since there are no request and the method didn't use the localeKey param.

PR #2262, opened 2025-07-08
security(encryption): Support for RS256 SECRET_KEY encryption.

dbauszus-glx:
This PR adds support for RS256 token encryption.

Any calls to the jsonwebtoken library to sign or validate token will now make use of the SECRET_ALGORITHM environment variable which defaults to 'HS256'.

HS256 is less secure than RS256 encryption which requires a key.

For testing purposes I created a key with https://cryptotools.net/rsagen 

The minimum length for RS256 encryption must be 2048.

image

It is highly recommended to generate keys locally / private for use in production environments.

The keyvalue can be stored in a *.pem file in the root directory with the name provided as SECRET_KEY envrionment variable.

eg. `"SECRET_KEY": "secret.pem",`

This will set the algorithm automatically to RS256 though a different algo can be specified in the process environment.

The public key can be uses to test the validity of the private key when attempting to validate a token [cookie] signature.

image

### DIFF
--- a/mod/user/cookie.js
+++ b/mod/user/cookie.js
@@ -71,7 +71,12 @@ export default async function cookie(req, res) {
     xyzEnv.SECRET,
     { algorithm: xyzEnv.SECRET_ALGORITHM },
     async (err, payload) => {
-      if (err) return err;
+      if (err) {
+        return res
+          .status(401)
+          .setHeader('Content-Type', 'plain/text')
+          .send('Invalid token');
+      }
       // Get updated user credentials from ACL
       const rows = await acl(
         `


RobAndrewHurst:
@dbauszus-glx 

I have added in a response return where when verifying the token and we error we return 'Invalid Token'.

[`08b0e3c` (#2262)](https://github.com/GEOLYTIX/xyz/pull/2262/commits/08b0e3ccb8d0040179ec2e37122207b2ac955561)

RobAndrewHurst (approved):
Happy for this to go in!

PR #2268, opened 2025-07-10
fix(register): Changed Content-Type to return html

RobAndrewHurst:
### Summary	

I have changed the Content-Type set in the register modules to `html` as this will better cater for both text or html.


dbauszus-glx (approved):

RobAndrewHurst:
> ### Summary
> I have changed the Content-Type set in the register modules to `html` as this will better cater for both text or html.

I have now removed the content-type restrictions on this module

PR #2272, opened 2025-07-11
fix(dropdown): Multi select filter dropdown and selected flag.

dbauszus-glx:
The event.target in an options dialog is not the li element but the select input element itself.

The dropdown selectOnChange method should toggle a flag (selected) on the relevant entry and return the entry as third argument in the callback.

The filter dropdown should be multiselect showing the current filter and to allow for toggling filter on and off.

This requires the option element to be assigned as li property to the entry [filter].

https://github.com/user-attachments/assets/2285225f-f18b-4a44-97b0-0f9a1abb4e74


cityremade:
image

Issue #2274, opened 2025-07-11
Filter dialog and configuration


dbauszus-glx:
It should be possible to display the filter panel in a dialog.

It should be possible to create filter for any infoj entry where possible without specifically configuraing the infoj entry.

dbauszus-glx:
The draft PR creates a multi_filter plugin.

The configuration however should be integrated in the layer.filter configuration in line with other layer panel methods.

The use of the filters.reloadCallback should be reviewed. This doesn't feel very elegant.

```js
mapp.ui.layers.filters.reloadCallback = mapp.utils.compose(
  mapp.ui.layers.filters.reloadCallback.bind(),
  updateCount,
);
```

dbauszus-glx:
Resolved in linked PR.

PR #2275, opened 2025-07-11
feature(multi_filter): Multi filter plugin integration

dbauszus-glx:
### Description
The multi_filter plugin currently just turns the filter panel into a dialog. This PR moves that functionality into core
by adding an option to `layer.filter`, `layer.filter.dialog` which can hold the options previously held in the multi_filter config:

e.g.
```JSON
  "filter": {
    "dialog": {
      "showOnLayerDisplay": true
    },
    "drawer": false,
    "exclude": ["ctrycd", "city_search", "name", "subname"]
  },
```

alternatively: 
```JSON
  "filter": {
    "dialog": true
    "drawer": false,
    "exclude": ["ctrycd", "city_search", "name", "subname"]
  },
```

This will yield a button in the panel:
Screenshot from 2025-08-05 12-22-56

Which will open a dialog which functions the same as multifilter currently.

### Issue
[#2274](https://github.com/GEOLYTIX/xyz/issues/2274)

### Testing
Can be tested on any filter by simply adding `dialog: true`.
Currently set up on `/bugs_testing/filter/all_filter.json`

### DIFF
--- a/lib/ui/layers/panels/filter.mjs
+++ b/lib/ui/layers/panels/filter.mjs
@@ -305,12 +305,12 @@ Within the debounce the location count will be executed. The min max will be gen
 function updatePanel(layer) {
   if (!layer.display) return;
 
-  if (!Object.keys(layer.filter.current).length) {
-    layer.zoomToFilteredExtentBtn?.style?.setProperty?.('display', 'none');
-    return;
-  }
-
-  layer.zoomToFilteredExtentBtn.style.setProperty('display', 'inline-block');
+  if (Object.keys(layer.filter.current).length) {
+    layer.zoomToFilteredExtentBtn?.style?.setProperty?.(
+      'display',
+      'inline-block',
+    );
+  } else layer.zoomToFilteredExtentBtn.style.setProperty('display', 'none');
 
   clearTimeout(layer.filter.debounce);
 


AlexanderGeere:
@dbauszus-glx I deleted the plugin and the functionality now lives in `lib/ui/layers/panels/filter.mjs`. Theres some documentation and some testing to be added. Then i will make it ready for review

dbauszus-glx:
@AlexanderGeere We will need to review and merge this PR into patch, then patch into minor, and then the minor into this PR before we can take this out of draft.

dbauszus-glx:
The aim of this PR is to fully integrate multifilter with the filter panel. I have updated the dictionary entries to reflect this https://github.com/GEOLYTIX/xyz/pull/2275/commits/05d09c37ccfcf703af58169bf647d7de9a484d77

dbauszus-glx:
It should be possible to add allfilter with includeAll boolean flag, some with include, and any not in exclude array.

dbauszus-glx:
The popout does not currently work.

dbauszus-glx:
The filter reloadCallback should be revised. This a very complex logic which was introduced for the multi_filter plugin only. There is probably a smarter way to do this than passing a reloadCallback method to a layer.reload method.

dbauszus-glx:
@AlexanderGeere When I use the popout on a drawer in a layer view the caret of the layer view itself is getting display: none;

image

image

dbauszus-glx:
@AlexanderGeere The include, exclude, includeAll logic was very convoluted. There was so many duplicate filter conditions.

This is much cleaner with a for loop in a dedicated method to return the list. https://github.com/GEOLYTIX/xyz/pull/2275/commits/cf56b5c77362185c895f672ef44ed5350bf2a3d9

skipped, no field, or excluded continue straight away.

Include all or directly included or filter true will create a filter definition with type.

Thereafter only a check whether the type is valid is required otherwise the filter can be cloned into the list.

```js
function listFilter(layer) {
  const list = [];

  layer.filter.include ??= [];

  layer.filter.exclude ??= [];

  for (const entry of layer.infoj) {
    if (entry.skipEntry === true) continue;

    if (entry.field === undefined) continue;

    if (layer.filter.exclude.includes(entry.field)) continue;

    if (
      layer.filter.includeAll ||
      layer.filter.include.includes(entry.field) ||
      entry.filter === true
    ) {
      entry.type ??= 'text';

      if (!Object.hasOwn(filterByType, entry.type)) continue;

      entry.filter = filterByType[entry.type];
    }

    // The filter is defined as a string e.g. "like"
    if (typeof entry.filter === 'string') {
      entry.filter = {
        type: entry.filter,
      };
    }

    if (!entry.filter?.type) continue;

    if (!Object.hasOwn(mapp.ui.layers.filters, entry.filter.type)) continue;

    entry.filter.title ??= entry.title;

    entry.filter.field ??= entry.field;

    list.push(structuredClone(entry.filter));

    return list;
  }
}
```

dbauszus-glx:
The reloadCallback could just be removed as this was only used by the multi_filter plugin.

AlexanderGeere:
> @AlexanderGeere When I use the popout on a drawer in a layer view the caret of the layer view itself is getting display: none;
> image image

@dbauszus-glx 
This was a measure i added to hide the caret and the view if the item you popped out is the last one, this should be working as expected now. When you pop out the last element the view should hide, popping out an element before the last should not do anything to the view.

dbauszus-glx:
I added the slider condition to display a slider when you create a filter with no min/max and zoom to a viewport which has a range for the filter.

simon-leech (changes requested):
Noticed a few issues at the moment. 

1. There's no locationCount shown in the filter panel indicating how many features are now available. 
image

2. The filter icon that is applied to a layer when a filter is applied by a user is no longer being shown. 
image

3. I don't think the `listFilter` function is behaving like the multi filter did either. If I provide just `filter.exclude: ["a"]` - I would expect all fields except "a" to be included. Currently only filters that have an explict `infoj.filter` object are being included. 
4.  I updated the console warning to be more explicit on what needs to happen to resolve the warning.
image
5. I set layer.filter.drawer to false in the multi_filter legacy function as this is the current implementation.

AlexanderGeere:
@simon-leech the count and button should be back now. I added includeAll to the lagacy part so it works as before

RobAndrewHurst (approved):

simon-leech (changes requested):
Nearly there - still seeing one issue however. 

1. If I have a `filter.title` set - this is not being passed properly. The `entry.title` is always being used.

AlexanderGeere:
> Nearly there - still seeing one issue however.
> 
>     1. If I have a `filter.title` set - this is not being passed properly. The `entry.title` is always being used.

Should be fixed

RobAndrewHurst:
@simon-leech Are you still going to have a review of this?

simon-leech (dismissed):
One final thing - the filter pills method - when you remove the final one in the list does not update the locationCount correctly (it should be being removed). 
This results in a locationCount that is outdated.

simon-leech (approved):

PR #2286, opened 2025-07-17
fix(Utils/CSV Upload): Check Columns in CSV against params.fields; Missing params must return an error

simon-leech:
This PR introduces a check on the length of the columns in the CSV against the length of the `params.fields` provided for the csvUpload. 

This fixes a bug whereby an incorrect length would silently fail. 
We return an error if this happens which can be displayed on the UI or checked against `instanceOf Error`

### DIFF
--- a/lib/ui/utils/jsonDataview.mjs
+++ b/lib/ui/utils/jsonDataview.mjs
@@ -124,7 +124,11 @@ function csvupload(dataview) {
         // This is required as response can be an array of results or a single result.
         // A single result may be returned as an error if missing queryparams.
         // A response of an array of results is returned if the csv upload is chunked into multiple requests.
-        const errors = response?.filter((result) => result instanceof Error);
+        let errors;
+        if (Array.isArray(response))
+          errors = response.filter((result) => result instanceof Error);
+        else if (response instanceof Error) errors = [response];
+
         // If there are errors, show an alert with all the error messages.
         if (errors?.length) {
           mapp.ui.elements.alert({


RobAndrewHurst:
The column count that was being performed would break if a column value had a
comma in it eg: "Rob, Hurst". So its better to make use of the
`splitRowIntoFields` method to handle just this kind of potential issue.

```diff
-- const columnCount = rows[0].split(',').length;
++ const columnCount = splitRowIntoFields(rows[0]).length;
```

dbauszus-glx (changes requested):
Currently reviewing this.

dbauszus-glx (dismissed):
Should the method return an Error if the upload method short circuits.

image

Upload complete from this custom plugin does not provide information that the upload didn't execute at all.

image

dbauszus-glx:
This needs more work and needs to work with the csvupload toolbar for a jsonDataview.

I create a very simple test table.
```SQL
create table csvupload_test (
  name varchar,
  postcode text
);
```

I create a very simple query module to load data into this table.
```js
export default (_) => {
  const data = _.body.data;

  return `
  INSERT INTO csvupload_test (name, postcode) 
  VALUES ${data};`;
};
```

I configure my test dataview like so:
```json
"dataviews": {
  "json": {
    "target": "tabview",
    "dataview": "Json",
    "toolbar": {
      "csvupload": {
      "query": "csvupload_query"
      }
    }
  }
},
```

I would expect to load a file with this content.
```csv
name,postcode
John Doe,BN18 0BT
Jane Smith,BT17 0ZE
Emily Davis,BT38 8GF
Michael Brown,BT29 4BG
Jessica Wilson,KW15 1TU
David Lee,BT34 3YE
Sarah Moore,BT63 5ZU
William White,BT28 0AJ
Ashley Adams,BT94 4QZ
Robert Harris,SE1 2AX
Linda Clark,JK78 9LM
Daniel Lewis,NO90 1OP
Lisa Hall,QR12 3ST
Paul Robinson,UV34 5WX
Karen Allen,YZ56 7AB
James Wright,BC78 9DE
Mary Scott,EF01 2GH
Brian Young,HI23 4JK
Susan Green,LM45 6NO
Thomas Phillips,PQ67 8RS
```

Without the fields configuration the csvUpload method warns and returns nothing. The setData method then falls over with undefined data.

image

simon-leech:
@dbauszus-glx I have made some changes: 

1. The parameters now return an error instead of a console.warn if missing.
2. I have tidied up the docs as some of the parameters are only required if using the shared query, others may not be. 
3. I have updated the jsonDataview to check on the upload for `instanceof Error` and display the error message if so in an alert.

dbauszus-glx:
It should be possible to define a layer as a string in the queryparams.

The checks for the fields, query, queryparams should be outside the promise and return an error. The Error should be self explanatory.

The csvupload toolbar method should be documented and alert with the error.message returned from mapp.utils.csvUpload().

image

A varchar schemaMethods has been added to the csvUpload utils module.

dbauszus-glx (changes requested):
I provide a csv with mismatch fields.

image

This crashes out here.

image

The columnCount isn't reached.
image

simon-leech:
@dbauszus-glx Pushed two commits to hopefully resolve this: 

1. Now check the length of all rows and only take the maximum for the comparison against the fields length. This is required as the first column may not be the longest.
2. Updated the `jsonDataview` to check on all responses if any errored and documented why this is needed.

AlexanderGeere:
The reponse filtering was throwing an error as the response is not always an array. I added a check for this

 [eb800df](https://github.com/GEOLYTIX/xyz/pull/2286/commits/eb800df806663588440a0622214b4c87efbe7197)

@simon-leech / @dbauszus-glx this line `dataview.setData(response);` on line 141 in jsonDataview.mjs cant work as when the upload is sucessfull `mapp.utils.csvUpload` returns null.

dbauszus-glx (approved):

dbauszus-glx:
@AlexanderGeere This is fine. The main reason for the csvupload toolbar util is to test the application the csvupload util. The default query does not return anything but a custom query might.

image

AlexanderGeere (approved):

PR #2287, opened 2025-07-17
fix(templates): Get Template defined on Locale

### LINKED ISSUE #2283: Locale roles in template
The _workspace/locales method must load [and cache] locale templates in order to check for roles defined in templates.

The locale_template in this example has a role defined making this locale unavailable. However the template has not been loaded and merged with the locale hence there is no role to check.

Image

RobAndrewHurst:
There is an issue with **locales that are loaded from templates** that have
roles defined in the locale template.

This prevents the roles check being performed as the locale doesn't have a roles
object until we fetch it.

The fix for this is that when fetching the locales from
`/api/workspace/locales` we then get each locale instead of checking against
what's currently in the cached workspace. At this point a roles check is done
and if the user doesn't have access to the role it won't be returned.

Closes #2283

### DIFF
--- a/public/views/_default.html
+++ b/public/views/_default.html
@@ -356,7 +356,7 @@
       `${mapp.host}/api/workspace/locale?locale=${locale}&layers=true`,
     );
 
-    if (locale instanceof Error || locales.length === 0) {
+    if (locale instanceof Error) {
       // Create the Dialog.node
       mapp.ui.elements.dialog({
         css_style:


simon-leech (dismissed):
I'm not sure if I'm doing something wrong but this isn't working as I would expect. 

I have a config like so 
``` json
  "locale": {
    "layers": {
      "OSM": {
        "display": true,
        "format": "tiles",
        "attribution": {
        }
      }
    }
  },
  "locales": {
    "europe": {
      "template": "europe_locale",
      "locales": ["uk_locale", "roi_locale"]
    },
    "uk": {
     "template": "uk_locale"
    },
    "roi": {
      "template": "roi_locale"
    }
  }
```

Off the PR (on minor) - I see this when I have no roles 
image
This is wrong as you should not see a dropdown for locales you cannot access.

On this PR - I see this when I have no roles 
image

This is also not right - I think this is displaying the default locale (which is used to merge into the `locales`). 
Should you not see a message saying you don't have access to any locales?

RobAndrewHurst:
@simon-leech 

I have added in a check on the _default.html view that if we don't get a locale back that returns an error or we don't get any locales back from the workspace then we will show the modal.

We will however need a default locale ie the one we are seeing in your screen shot as we need to assign the base plugins (zoomBtn, admin, login) to the locale for the user to be able to interact with them.

simon-leech (approved):
Happy with this now - thanks @RobAndrewHurst ! 🥳

AlexanderGeere (approved):

dbauszus-glx (changes requested):
This is working but not as expected.

The default locale returns a layer but the warning is shown.

image

This is because the locales length is 0

image

RobAndrewHurst:
@dbauszus-glx How should this work with a default locale defined in a workspace? Because in order to show the plugins (admin, zoom and login/logout) we need a locale to assign them. 

This layer is present because in the workspace it's defined on the "locale"

dbauszus-glx:
@RobAndrewHurst  If the default locale is not role restricted then it should show just like any other locale.

RobAndrewHurst:
@dbauszus-glx Isn't that the case in this example?

```json
{
  "templates": {
    "OSM": {
      "src": "${RESOURCES}/bugs_testing/templates/osm_layer.json"
    }
  },
  "locale": {
    "layers": {
      "OSM": {
        "template": "OSM"
      }
    }
  },
  "locales": {
    "us": {
      "roles": {
        "you-cant-sit-with-us": null
      }
    },
    "uk": {
      "roles": {
        "stop-trying-to-make-fetch-happen": null
      }
    }
  }
}
```

if I don't have the roles : "you-cant-sit-with-us" or "stop-trying-to-make-fetch-happen" then I should only see a locale with the OSM layer?

dbauszus-glx:
@RobAndrewHurst correct.

This warning should not be triggered because locale is not instance of error.
image

RobAndrewHurst:
Ah ok. Then @simon-leech's example threw me off, because he was expecting the message even if we got that default locale.

I think we just need to clarify what should happen

PR #2290, opened 2025-07-17
fix(mvt): WKT Properties not allowing callback

simon-leech:
The reload method on an MVT layer for layers with `wkt_properties` set must check for the presence of a callback function and run it if so. 

This is required for when filtering so that the filter icon correctly appears when a filter is applied to a layer.


AlexanderGeere (approved):
Works as expected

PR #2296, opened 2025-07-22
fix(location.updateCount): Move updateCount to filter panel

dbauszus-glx:
I moved the updateCount method to the filter panel module since the method is only called from there.

The method should short circuit if the layer is not displayed or the current filter length is naught.


cityremade (approved):
looks good to me.

RobAndrewHurst (approved):

PR #2297, opened 2025-07-28
CSS - common - outlined class.

cityremade:
`common` stylesheet now adds a reusable `.outlined` class which can be used on buttons and other elements like report links. It adds a border and spacing around the element in order to make it stand out within the content.

Report links in location view now have outlined class as this is a known case when link may not be readable enough.

image


cityremade:
Link now documented.

css class made optional:
```json
{ 
   "type": "report", 
   "data_id": "my_report",
   "css_class": "outlined",
   "icon_name": "stacked_line_chart", /* but for reports this will always set to description icon */
   "report": {
       "template": "my_report_template",
       "label": "My report"
     }
}
```

I've removed outer div so that the element is now a pure hyperlink tag with inline content. The div must have been a legacy from when icons were blocks.

Included data_id - for element identification in analytics.


If it's a report the icon is always set to description - should it be 'open in new' by default and defined when needed? this blocky icon may not always be relevant.

simon-leech:
I pushed a little tweak to nullish assign the report icon as it should be possible to overwrite if required. 

Can it be possible for the width to not be 100% but instead just the width of the icon and label?

dbauszus-glx:
@simon-leech In that case we need to have the container in a container. Because the val container must be the same width as the other val otherwise it will look comical.

dbauszus-glx:
css_class was a bad name which wasn't helped by the documentation.

_css_class is the css class string applied to the link element. icon_class is the css class applied to the icon element._

Much easier to use link_class and icon_class. One is the class applied to the link element the other is the class applied to the icon element, The naming is self explanatory.

AlexanderGeere (approved):
I added `align-items: center` to `link-with-img`. Workking as expected for me.

dbauszus-glx:
@AlexanderGeere very nice. the whole .link-with-img > * was actually unnecessary. all the text styling should be on the class and inherited by its node children.

simon-leech (approved):

PR #2301, opened 2025-07-29
Change locales loop and names retrieval

dbauszus-glx:
The return from the get workspace locales should return an array locales with their name, key, and nested locales.

This behaviour changed in this PR for the 4.16 minor release. https://github.com/GEOLYTIX/xyz/pull/2287

An array of the complete locales was now returned.

I change the Promise.all and return array.filter in the first commit for better readability.

The correct locales array is now rertuned. 

image

However; The assignment of the name is still bugged. The name for the uk locale should be 'United Kingdom'.

image

### DIFF
--- a/mod/workspace/_workspace.js
+++ b/mod/workspace/_workspace.js
@@ -137,17 +137,25 @@ async function locales(req, res) {
     return;
   }
 
-  const locales = await Promise.all(
-    Object.keys(workspace.locales).map((localeKey) =>
-      getLocale({
-        user: req.params.user,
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
The naming issue should be resolved by removing the conditional assignment of the name in the workspace cache module which happens prior to the template being loaded in the getLocale module. The mergeTemplate method will not assign the name from the template if the locale object already has a name property.

simon-leech (approved):
This is working as expected for me - names are correctly being passed!

PR #2307, opened 2025-08-01
filter.viewport for numeric filter with slider and histogram

cityremade:
Adds viewport parameter in min max query used for calculating range on numeric filters.
Mn max range calculated on map change end event for all layers with 

* `layer.filter.viewport = true`
* existing numeric and integer filters

generateMinMax function reworked using query params and string utilities. 

filter panel - on map change end min max updated along with location count.


cityremade:
Note: layer.filter.viewport should be inherited initially from layer.viewport.

dbauszus-glx:
The updatePanel method is now debounced. min max is only calculated for current filter.

dbauszus-glx:
The histogram method should short circuit if the data is not an array.

The slider now shows the min and max below and the current value above.

The current value defines the range for the histogram.

The filter.field must be taken out from the min max filter query param as otherwise the min max could never exceed the current filter.

simon-leech (changes requested):
image

Couple of comments: 

1. What is the second set of numbers below the slider? I guess this is the min/max outside of the viewport but I don't have the viewport flag included so this shouldn't appear.
2. The numbers below the slider should be formatted using the `numericFormatter`
3. I'm seeing an error in the timeout as the `layer.filter.list` does not exist
image

dbauszus-glx:
@simon-leech 

Numbers below are the min / max. A slider without viewport still has a min and max. I will make the display optional with a flag but not related to viewport.

The numbers can be formatted.

Can I see the configuration for the filter panel without any filters available?

dbauszus-glx:
@simon-leech The showMinMax property will control whether the minmax-row is displayed. https://github.com/GEOLYTIX/xyz/pull/2307/commits/52e091e59ac322816f2e1b63e125871a39cd44fe

dbauszus-glx:
@simon-leech Optional min max labels are now formatted. https://github.com/GEOLYTIX/xyz/pull/2307/commits/d6b877abc47c5af3535389ddc491294e30befbec

image

dbauszus-glx:
@simon-leech  I made the filter.list conditional but really would like to see this happening. https://github.com/GEOLYTIX/xyz/pull/2307/commits/1c7d2ead9080dbfdb0419beebd1625cb6ed9925e

AlexanderGeere (approved):
Working as expected for me.

simon-leech (approved):

PR #2309, opened 2025-08-05
hideDisabled layer.view

cityremade:
`viewConfig` now supports `hideDisabled` flag for layers with multiple zoom levels.

When set to true layer view container will be hidden in the layer list until the layer is within restricted zoom level again.

`viewConfig` now merges defaults with layer configuration props - previously the object would be entirely overwritten with the loss of most common defaults.

Please note that with property `display: none` in css we cannot access the element style, classlist, attributes etc as it cannot be found within the page. This may matter for existing custom show/hide layer callbacks which would need some checks whether the tags within the layer view are accessible.

I've gone for the use of the common class `display-none`. An alternative approach would be to remove the element from the list and later put it back yet this would require tracking the parent and the index where the element should sit - index may be changed with upcoming tools for importing layers.

```json
{
  "format": "mvt",
  "srid": "3857",
  "geom": "geom_3857",
  "qID": "id",
  "tables": {
    "9": null,
    "10": "dev.scratch",
    "11": "dev.scratch",
    "12": "dev.scratch",
    "13": "dev.scratch",
    "14": null
  },
  "viewConfig": {
    "hideDisabled": true
  }
}
```

### DIFF
--- a/lib/ui/layers/view.mjs
+++ b/lib/ui/layers/view.mjs
@@ -137,10 +137,13 @@ export default function layerView(layer) {
   mapp.utils.render(layer.view, layer.drawer);
 
   // The layer may be zoom level restricted.
-  layer.tables &&
+  if (layer.tables) {
     layer.mapview.Map.getTargetElement().addEventListener('changeEnd', () =>
       changeEnd(layer),
     );
+    // Call changeEnd to set the initial state of the layer.
+    changeEnd(layer);
+  }
 
   return layer;
 }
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
@simon-leech I will look into the issue in regards to initial load.

As for the groups. This is too much spaghetti logic. Don't set the flag in a group. We also don't prevent users to create groups with a single layer which is just poor configuration.

simon-leech (approved):
This is working for me now :)

PR #2311, opened 2025-08-06
Test Plugin (Conditional Library imports) 

### LINKED ISSUE #2206: test script load in default view
The tests script should not always be loaded in the default view, whether used or not.

Script required to run tests should be optionaly loaded only when tests are actually run.

```html
<script
  type="module"
  src="{{dir}}/public/js/tests/_mapp.test.js"
  defer
></script>
```

RobAndrewHurst:
I have made the import of Codi & the bundled test library conditional if we are
running tests.

This is determined by the hooks defined in the URL params.

When the test plugin on load finds the urlparam 'test' then we will load the
bundles.

```js
export async function test(plugin, mapview) {
  // Only import and assign functions if we are testing
  if (mapp.hooks.current.test) {
    try {
      if (!globalThis.codi) {
        await import("https://esm.sh/codi-test-framework@1.0.37");
      }

      if (!globalThis._mappTest) {
        await import(`${mapp.host}/public/js/tests/_mapp.test.js`);
      }
    } catch (error) {
      console.log(error);
    }

    ...
  }
}
```

There were also a bunch of tests that either needed to be removed or updated
from code that had changed in the previous minor release.

When merged this will close #2206


RobAndrewHurst:
> When I add `?test=integrity` to the URL, it now only runs the Template Test Path, all the layer tests are not being ran.

Oh right. I rememeber why I now executed in the _default view.

I will think of how I can make this work

dbauszus-glx:
@RobAndrewHurst URL parameter are acessible inside the execution of the plugin. Just shortcircuit without this or that param.

dbauszus-glx:
You can avoid nesting everything by using a short circuit condition. https://github.com/GEOLYTIX/xyz/pull/2311/commits/bffde446d0258aa3ef21c863172f45b38c3b7776

Does the plugin have other properties than options? Why not put the options properties directly in the plugin object? Saves an unnecessary nesting since no grouping of proprerties is needed.

Can you try to use the esm import utility you don't need to check for the import in this case since the utlity takes care of these checks.

RobAndrewHurst:
> You can avoid nesting everything by using a short circuit condition. [bffde44](https://github.com/GEOLYTIX/xyz/commit/bffde446d0258aa3ef21c863172f45b38c3b7776)
> 
> Does the plugin have other properties than options? Why not put the options properties directly in the plugin object? Saves an unnecessary nesting since no grouping of proprerties is needed.
> 
> Can you try to use the esm import utility you don't need to check for the import in this case since the utlity takes care of these checks.

The configuration is designed to be extensible, allowing specific test libraries to be included in the global test environment. The 'options' object is passed directly to the Codi test runner, enabling customization of the test runner's report output. An example configuration might look like this:

```json
{
  "test": {
    "suites": ["file:./assets/suite-1.js"],
    "options": {
      "quiet": true,
      "showSummary": true
    }
  }
}
```

RobAndrewHurst:
I do however have an issue when it comes to the integrity tests.

I need to run this test after the layers have been loaded and decorated into the mapview.

This was done with a manual call from the _defaul.html view's js at the end of the script. 

@dbauszus-glx Do you have a potentially better way that we can call the test?

RobAndrewHurst:
@simon-leech The integrity tests should now be working as expected.

@dbauszus-glx I have made the plugin object is the configuration. - I also added in some backwards compatibility with a warning if the workspace has test.options.

This is ready for another review

AlexanderGeere (approved):
Working as expected, Nice!

simon-leech (dismissed):
review in progress

dbauszus-glx (approved):

PR #2318, opened 2025-08-08
Loading Class

simon-leech:
Addition of a `loading` class that will add 7 dots that animate to suggest loading visual to the user. 
The `loading` class can be applied to any element. 

WIP but opening for discussion as we need to test the various elements and make sure the CSS positions correctly. 

Screenshot 2025-08-08 085619

### DIFF
--- a/public/css/_ui.css
+++ b/public/css/_ui.css
@@ -26,6 +26,8 @@
 
 @import url('./elements/_key.css');
 
+@import url('./elements/_loading.css');
+
 @import url('./elements/_pills.css');
 
 @import url('./elements/_searchbox.css');
--- a/public/css/elements/_loading.css
+++ b/public/css/elements/_loading.css
@@ -1,0 +1,111 @@
+.loading {
+  position: relative;
+  pointer-events: none;
+}
+
+.loading::after {
+  --dot-size: 24px;
+  --dot-color: var(--color-primary);
+  --dot-gap: 24px;
+  content: "";
+  position: fixed;
+  top: 50%;
+  left: 20%;
+  width: var(--dot-size);
+  height: var(--dot-size);
+  border-radius: 50%;
+  box-shadow:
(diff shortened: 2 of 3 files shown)


dbauszus-glx (changes requested):
I will need to see a test case for this. The class should be applied to a location view during a blocking query.

Then we can add a test query which just waits for 10 seconds or so and see the loading animation on the location view while the blocking query executes.

cityremade:
this might depend what the spinner means. Horizontal row of dots may mean something else than radial or circling elements, in a way like 3 dots mean typing, circular movement means waiting for the end and horizontal bar means progress 🤔

shall we make a list of cases when we need to display loader?
this might apply to location view only or the viewport.

simon-leech:
I added a test case with the `loading` class being added for blocking queries. 
So we can now easily test it for locations by using a blocking query.

simon-leech:
> this might depend what the spinner means. Horizontal row of dots may mean something else than radial or circling elements, in a way like 3 dots mean typing, circular movement means waiting for the end and horizontal bar means progress 🤔
> 
> shall we make a list of cases when we need to display loader? this might apply to location view only or the viewport.

I chose to apply a horizontal circles for dots as loading as like you say, we don't want to indicate progress, just a visual cue that something is happening - and 3 dots would be for typing which is not happening here.

dbauszus-glx:
@simon-leech What would a sample geometry entry look like to test this?

cityremade:
what if user clicks all over the map while this is spinning?

simon-leech:
@dbauszus-glx I have applied this as a border animation now instead and like how it looks / behaves. 

Please re-test, should be working in both light and dark mode!

dbauszus-glx (approved):
Looking good now.

RobAndrewHurst (approved):

PR #2320, opened 2025-08-08
enhance(location): Geometry entry update

dbauszus-glx:
The geometry entry now uses the location.update method.

The logic is as follows. Unsaved edits must be confirmed for updates in order to begin the modify or draw interaction.

Edits made during the modify or draw interaction will be saved when the location update method is called.

The location.confirmUpdate() method has been added in this PR. The method checks for unsaved changes awaits a confirm dialog. Confirm calls location update and returns true. Cancel returns null. The modify/draw interaction is not executed in this case.

The location.update method now receives a callback parameter. With this the default callbacks which update the location view are not updated. This allows to render the location view with the edit still toggled on. This is required to enable to draw/modify interaction after the rendering the location view with updated entry values.

The pin entry method was poor. The show and hide methods where not consistent. Multiple properties were assigned multiple times. The buffer method does not require to be async and can always return a buffer [0 if undefined].

A removeLayer method has been added to the location to remove a layer from either the pin or geometry entry methods.

### DIFF
--- a/lib/ui/locations/entries/geometry.mjs
+++ b/lib/ui/locations/entries/geometry.mjs
@@ -212,6 +212,9 @@ The edit method will push elements for geometry edits into the entry.elements ar
 
 @param {infoj-entry} entry type:geometry entry.
 @property {Object} [entry.edit] editing must be configured for the geometry entry.
+@property {string} [entry.edit.modify_label] label for the modify button. If not defined, defaults to dictionary term.
+@property {string} [entry.edit.delete_label] label for the delete button. If not defined, defaults to dictionary term.
+@property {boolean} [entry.edit.delete] flag to enable delete button. Without it, only the modify button will be shown.
 @property {Object} [entry.value] entry must have a geometry value to be editable.
 */
 function edit(entry) {
@@ -239,7 +242,10 @@ function edit(entry) {
   entry.elements.push(entry.edit.modify_btn);
 
   entry.edit.delete_label ??= mapp.dictionary.delete_geometry;
-  entry.elements.push(mapp.utils.html`<button class="action wide"
+
+  // Delete is only shown if edit.delete flag is provided.
+  entry.edit.delete &&
+    entry.elements.push(mapp.utils.html`<button class="action wide"
     onclick=${() => deleteGeometry(entry)}>
     <span class="material-symbols-outlined color-danger">ink_eraser</span>
     <span style="color: var(--color-danger)">${entry.edit.delete_label}</span>`);


cityremade:
on geometry/isoline entries we use edit: true across resources. 

this line is failing on selecting a location with an entry with the following HERE isoline setup: 

```json
{
    "type": "isoline_here",
    "title": "Isoline HERE",
    "field": "isoline",
    "fieldfx": "ST_AsGeoJSON(isoline)",
    "display": true,
    "edit": true,
    "params": {
         "label": "Isoline HERE",
         "apiKey": "<apiKey>",
         "panel": ["datetime", "minutes", "transportMode"]
       }
}
```
Screenshot 2025-08-18 at 13 39 02

Change needed: 
entry.edit=true should be further on handled as an object `{}`

cityremade:
Second click on modify geometry button: interaction layer is cleared but the original feature won't show anymore.

Screenshot 2025-08-18 at 15 06 25

geometry displayed

Screenshot 2025-08-18 at 15 06 32

geometry in editing mode

Screenshot 2025-08-18 at 15 06 39

geometry after editing mode is turned off with the modify geometry button - feature no longer on the map

dbauszus-glx:
@cityremade entry.edit will now be an empty object if defined as true. https://github.com/GEOLYTIX/xyz/pull/2320/commits/792c4086433ebf555060848a398da193ef7fa90c

Actually. This is better done in the infoj module iteration as the numeric filter has already duplicate code for this. https://github.com/GEOLYTIX/xyz/pull/2320/commits/4d7b08c218cb33771b92689c1bce0e7fdab500c0

dbauszus-glx:
@cityremade The display of the geometry should be restored by calling the show method from the updateCallback method if the feature param is falsy. https://github.com/GEOLYTIX/xyz/pull/2320/commits/aa6376a0b6833b240578d1986fe26f0819ccaf42

dbauszus-glx:
@simon-leech Both issues should now be resolved in the recent commits.

simon-leech:
There's something wrong with non-editable type geometry entries now. 

I am unable to select a location with this configuration attached - removing these entries fixes it. 

``` json
 {
      "label": "Catchment 1",
      "field": "field_3856",
      "fieldfx": "ST_asGeoJSON(field_3856)",
      "srid": "3857",
      "type": "geometry",
      "style": {
        "fillOpacity": 0.8
      }
    },
    {
      "label": "Catchment 2",
      "field": "field_3857",
      "fieldfx": "ST_asGeoJSON(field_3857)",
      "srid": "3857",
      "type": "geometry",
      "style": {
        "fillOpacity": 0.8
      }
    },
```

simon-leech:
Couple of additional issues I'm seeing: 

1. Begin Modify Geometry > Move node > press cancel > The button is correctly set to not be active. However the mapview highlight interactions are lost (I can no longer select from any layers on the map, unless I refresh the page). 
2. Pressing Delete Geometry throws an error 
image

dbauszus-glx:
@simon-leech Highlight interaction should now be restored on cancel. I am working on the delete geometry method.

Scrap that. Not yet resolved.

dbauszus-glx:
@simon-leech The deleteGeometry method should be working now.

@cityremade Do we know why the danger-color button element is not colored as expected?

image

cityremade:
it's because class `.color-danger` is only defined for material symbols as we didn't want to manipulate buttons with arbitrary choices. Buttons are primary.

cityremade:
maybe good opportunity to make this button an action button with a coloured icon?

RobAndrewHurst (approved):

PR #2321, opened 2025-08-11
feature(user): added optional roles param on add api

RobAndrewHurst:
Added Support for an optional roles array in `mod/user/add.js`.

If provided, it conditionally includes roles in the INSERT, using parameterized SQL.
  - It accepts either:
    - an array (e.g. roles=a&roles=b), or
    - a comma-separated string (e.g. roles=admin,editor).

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -29,14 +29,18 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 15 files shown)


dbauszus-glx (approved):
This works well.

I moved the roles into one conditional block rather than three checks for roles.

```js
  if (typeof req.params.roles === 'string') {

    // Create roles array from string param.
    const roles = req.params.roles
      .split(',')
      .map((role) => role.trim())
      .filter((role) => role.length > 0);

    columns.push('roles');
    values.push(roles);
  }
```

PR #2323, opened 2025-08-12
fix(objTemplate): Template Merging WIP

simon-leech:
WIP re-write

### DIFF
--- a/mod/workspace/getLayer.js
+++ b/mod/workspace/getLayer.js
@@ -96,7 +96,11 @@ export default async function getLayer(params, locale) {
     ? Roles.objMerge(layer, params.user?.roles)
     : layer;
 
-  layer = await mergeTemplates(layer, params.user?.roles);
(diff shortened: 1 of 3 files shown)


dbauszus-glx:
While testing this I noticed that requesting an undefined template will crash the process.
image

dbauszus-glx:
I added check for an undefined template request. https://github.com/GEOLYTIX/xyz/pull/2323/commits/f97d9c1c93c65b9dbaf66cebc3922ea2d8f24dff

dbauszus-glx:
@simon-leech  The try catch has been removed from the getLayer module as the issue could happen with a locale as well. Most importantly though, the mergeTemplates method must not fail.

dbauszus-glx:
This issue was already resolved by returning the role check failed error object. I'll close this PR and create a patch PR for the undefined key issue.

PR #2327, opened 2025-08-15
Ensure that template fetch errs are caught in workspace test

dbauszus-glx:
The getFrom module should return the logged error.

The getTemplate method should assign an error response as err property on the template.

The workspace.templates should not be iterated through inside the iteration of the workspace.locales in the cacheTemplates method but afterwards.

The iteration of the workspace.templates in the test method should check whether the template is an error or whether the template object has an err property which is an error and log these.

image

### DIFF
--- a/mod/workspace/getTemplate.js
+++ b/mod/workspace/getTemplate.js
@@ -58,15 +58,12 @@ export default async function getTemplate(key) {
   let template;
   if (typeof key === 'string') {
     template = await getTemplateObject(workspace, key);
-
-    if (template instanceof Error) {
-      return template;
-    }
-
-    // The template.key property value must be the same as the key reference in the workspace.templates{}
-    template.key = key;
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
Instead of crashing the process the getTemplate method will now return an error if the key is undefined.

image

dbauszus-glx:
I unnested the template.key check to reduce the complexity. https://github.com/GEOLYTIX/xyz/pull/2327/commits/ecd2918480b15d2954ad368d240a39b9dae800e4

However I am not sure how to test this as I don't know any example where a template would not have a key.

dbauszus-glx:
The complexity can be further reduced by unnesting the error check which is possible if the key is assigned in the getTemplateObject method.

https://github.com/GEOLYTIX/xyz/pull/2327/commits/2f892cd6dd21786e18b9b9c1e9a164d103c0ca05

PR #2340, opened 2025-08-27
fix(filterDialog): unnest filterDialog method

dbauszus-glx:
The `filter.dialog{}` configuration object and it's properties were not documented in the 4.17 inclusion.

The dialog button creation has has been moved to the filterDialog method which documents these properties.


AlexanderGeere:
@dbauszus-glx The dialog button is .btn now, i just fixed in the onClose function in the dialog.

AlexanderGeere (approved):

simon-leech (approved):
Working as before for me :)

Issue #2319, opened 2025-08-08
Cross to close a tab


jfitzpatrick17:
What

- Provide an option to allow a user to close a tab from the tab itself

Why

- Currently a user has to go to the relevant area in the infoJ to untick the option if they don't want to see the tab - this isn't immediately obvious and isn't a great user flow when there are lots of tabs, some relating to different locations
- The addition of a close cross on the tab would allow the user to close the tab more quickly, helping to improve the user experience

<img width="438" height="117" alt="Image" src="https://github.com/user-attachments/assets/41c2d2db-079f-4ac5-90e8-36de51d9f0ad" />

dbauszus-glx:
Resolved in linked PR.

PR #2342, opened 2025-08-27
feat(Tabview): Add close button to tabs

AlexanderGeere:
## Description
This PR adds closing functionality to the tabs of a tabview. 
image

Clicking on the close button calls the `removeTab()` function. The tab can also be closed by middle clicking on the tab, which also calls `removeTab()`. 

The close button can be hidden by specifying `hideTabClose: true` on the entry/dataview:
image
This will preserve the middle close functionality

The close button and middle click can also be disabled by specifying `disableTabClose: true`, which will also remove the close button.

## GitHub Issue
#2319 

## Type of Change
- ✅ Documentation

## How have you tested this?
tested locally on `bugs_testing/dataviews/dataviews.json`

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/public/css/layout/_tabview.css
+++ b/public/css/layout/_tabview.css
@@ -27,18 +27,14 @@
 
     & > .tab > .header {
       color: var(--color-primary);
-      min-width: 102px;
+      min-width: 100px;
       height: 100%;
       display: flex;
-      align-items: center;
       white-space: nowrap;
       border: solid 1px var(--color-border);
-      border-radius: 3px 17px 0 0;
-      padding-top: 10px;
-      padding-bottom: 10px;
-      justify-content: space-around;
-      padding-left: 10px;
-      padding-right: 5px;
+      border-radius: 3px 15px 0 0;
+      padding: 5px;
+      justify-content: space-between;
       -webkit-user-select: none;
       -moz-user-select: none;
       -ms-user-select: none;
--- a/public/css/ui.css
+++ b/public/css/ui.css
@@ -1413,18 +1413,14 @@ dialog {
     background-color: var(--color-base);
     & > .tab > .header {
       color: var(--color-primary);
-      min-width: 102px;
+      min-width: 100px;
       height: 100%;
       display: flex;
-      align-items: center;
       white-space: nowrap;
       border: solid 1px var(--color-border);
-      border-radius: 3px 17px 0 0;
-      padding-top: 10px;
-      padding-bottom: 10px;
-      justify-content: space-around;
-      padding-left: 10px;
-      padding-right: 5px;
+      border-radius: 3px 15px 0 0;
+      padding: 5px;
+      justify-content: space-between;
       -webkit-user-select: none;
       -moz-user-select: none;
       -ms-user-select: none;


dbauszus-glx (changes requested):
A close button should never be top left. Convention is for the button to be on the right.,

chrome
image

edge
image

notepad
image

AlexanderGeere:
@dbauszus-glx The button has been moved:
image

AlexanderGeere:
> Please can you add `notranslate` to the close button that's added

Done

simon-leech (changes requested):
The functionality is working well for me! 

Couple of comments:

1. With a really long name (silly I know but possible) the button ends up above the text not to the right 
image
2. In the HTML structure - the close button is before the current button for the title, yet its positioned after. Can we swap these so that its easier to follow? 
image

dbauszus-glx:
Personally I prefer a max width with ellipsis in case of very long title.

AlexanderGeere:
> Personally I prefer a max width with ellipsis in case of very long title.

I prefer this as well

dbauszus-glx:
Or fade 
image

cityremade:
cross can be also in line with text. It's not that great in the rounded corner.
Something like this:

Screenshot 2025-08-27 at 16 24 08

AlexanderGeere:
> cross can be also in line with text. It's not that great in the rounded corner. Something like this:
> Screenshot 2025-08-27 at 16 24 08

@cityremade 
Centered!
image

cityremade:
this is so much better. Could it also use some more spacing against the right border? At times it has imbalance between vertical space and the distance to the right border.

AlexanderGeere:
> this is so much better. Could it also use some more spacing against the right border? At times it has imbalance between vertical space and the distance to the right border.

@cityremade 
I added some padding on the right

dbauszus-glx:
@AlexanderGeere Can you address this issue in the tabview module.

image

AlexanderGeere:
@dbauszus-glx 
> I don't understand the "disableTabClose" property. How does this work?
> 

I thought it would be nice to be able to disable the closing functionality so if you specify disableTabClose, The new close button will not display and middle clicking will not close the tab.

There's a check on line 134:
```
  if ((e.which === 2 || e.button === 4) && !entry.disableTabClose)
```

This disables the middle click close

and on line 142:
```        
   ${showCloseBtn ? closeBtn : ''}
```

Which will hide the close button.

 The checkbox will still work regardless. 

> How do I test a disabled tab? eg. a dataview tab.
> 

You can add `disabled: true` to a dataview entry, but its intended for when a query does not return data, then the checkbox 
will be disabled.

> Why is the close button a span element, but the tab label a button element?
> 
> It would make more sense to have the close button be a button and the label a span.
> 

I made them both buttons as clicking on the `entry.tab_btn` will activate the tab so i think it should still be a button. 

> Can you comment on the e.preventDefault in the closeBtn span onclick event listener?
>
Done

simon-leech (approved):

dbauszus-glx (dismissed):
@AlexanderGeere Is it possible to have a tab which is disabled but can be closed?

This tab has the closed disabled but looks fully disabled while it is not.

image

Can you update the CSS to use just pixel or em but not mixed. It's hard for me to wrap my head around a relation of min-width: 8em, max-width: 250px;

image

AlexanderGeere:
@dbauszus-glx 
> @AlexanderGeere Is it possible to have a tab which is disabled but can be closed?
> 

The disabled tab has `inert` applied to, which no interaction can occur so you wouldnt be able to close it. Disabled tabs should not appear though as the main function of the disabled property is to disable the checkbox in the infoj, when theres no data, which will prevent the tab from showing up any way. 

I have switched the css to use `px`.

AlexanderGeere:
@dbauszus-glx 
> This tab has the closed disabled but looks fully disabled while it is not.
> image
> 
Looking at the config:
screenshot-2025-09-08_16-21-17
This is working correctly, as disabled is set to true. 

the `field_limit` tab is the one that has disableTabClose
screenshot-2025-09-08_16-23-18
And it also correctly has no close button.

PR #2352, opened 2025-08-28
Revise iteration and conditional check in distributed theme module me…

dbauszus-glx:
This PR addresses SonarCloud issues in regards to the index condition in the ditributed theme method.

The category index will be assigned first.

Iteration happens if the category index is used by an intersecting feature.

Iteration breaks if an unused category index is found in the iteration.

The documentation for all theme modules has been updated to correct the confusion between params and their properties.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -71,7 +71,6 @@ export default function MVT(layer) {
   function reload() {
     if (layer.wkt_properties) {
       wktPropertiesLoad(layer);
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
@AlexanderGeere That is expected in a pseudo random distribution. You can reload on minor and patch and will see that the distribution is different on each load.

https://github.com/user-attachments/assets/f5b7db3a-dea7-483b-99d3-558e427b193a

RobAndrewHurst (approved):

PR #2354, opened 2025-08-28
Replace deprecated execcommand copy with clipboard api.

dbauszus-glx:
This PR replaces the deprecated document.execCommand('copy'); method in the copyToClipboard utility module with a call to the clipboard API.

A test case for this has been added to the jsonDataview module as otherwise the method was never called from any other core method.

https://github.com/user-attachments/assets/7e5f7ee6-cb2e-4be0-a82a-7a87ab88bf61


AlexanderGeere (approved):
I just changed the text to copy to clipboard

RobAndrewHurst (approved):

PR #2368, opened 2025-09-01
getLayer async in cacheTemplates workspace module

dbauszus-glx:
Instead of awaiting each getLayer request in the locale/layer iteration of the cacheTemplates workspace method the getLayer methods can be awaited async with Promise.allSettled() array.


simon-leech (approved):
This is working fine for me - tested on a default MAPP view and 2 different custom views and working as previously. 

Core admin view seems to be working fine too 👍

RobAndrewHurst (approved):

PR #2375, opened 2025-09-02
fix(login/register): language templates and forms

dbauszus-glx:
The aim of this PR is to introduce login and registration forms which make use of the dictionary. There should not be duplicate views for different languages.

It must be possible to load the mapp library without the mapp css or the ol library loaded. This previously crashed out and has been addressed in the mapp.mjs module.

It should be possible to load the ui library before or without the mapp library without the the ui.mjs crashing.

tbd
The styling for the loginForm element shoulld be defined in the ui.css
The loginForm element should work without the need for element id


simon-leech (changes requested):
I'm getting an error 'Cannot POST /register' 
when attempting to sign up a new account

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2375/commits/e5524f915470c4ee5d69f15a4e6f60920e017b9e

the input element can be assigned to their respective const. an id lookup is no longer required in the check method.

disabledSpan element can be assigned to a const. the check method should set the textContent and not innerHtml which is a potential risk.

all dictionary entries should be assigned to the params. this allos the method to be called with custom params and not require dictionary updates.

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2375/commits/f14c4ea9c43ccc582a6ce8528c9d89ee3bc237c0

The register form should never use innerHtml but text only. Otherwise there is a series risk of code injection.

The agreement and next info has been shortened. This isn't read anyways by most users and should not contain GEOLYTIX specific information. At best this might be misleading because a different email then the dictionary hardcoded email is configured as transport. At worst someone forks the dictionary and uses GEOLYTIX agreement without representing the organisation.

Custom templates with organisation specific agreements should be used anyways.

image

The dictionary texts are assigned to the params.

The language input form needs to get the value from the mapp.language property otherwise a user can not change their language through registration.

agreement_checkbox.querySelector('input') should be used to control the checkbox rather than assigning an id first in order to look up an id.

registerBtn is a button element. no need for an id lookup.

@cityremade 

I only updated the German dictionary apart from the English.

image

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2375/commits/6ab3323d905dbfff357f51e1c8d2394c7516e17d

It is easier to assign inputs to const than assigning inputgroups and then using an id lookup to get the input element.

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2375/commits/36014844462873c0ea9c8cd6c48a3bcf2069e45f

I updated the spanish dictionary and removed the entries from languages which I can not possibly review or confirm since I do not speak those languages.

dbauszus-glx:
I found this article interesting in regards to machine translations for interfaces. Especialy in regards to privacy agreements.

https://uxdesign.cc/linguistic-dead-ends-the-new-deceptive-pattern-plaguing-japan-0e4cb9506c48

cityremade (approved):

AlexanderGeere (changes requested):
The error messages for things like blocked user, wrong password dont show up anymore:
On patch it looks like this: 
screenshot-2025-09-22_11-06-20

On the PR no message appears:
screenshot-2025-09-22_11-39-32

dbauszus-glx:
@AlexanderGeere This issue should now be resolved by setting the msg param on the body of the login view.

cityremade:
Restored password input pattern attribute.
Added general validation error message.
Minor tweaks like asterisks for required fields.
One more entry needed for dictionaries then, register_error.

AlexanderGeere (approved):
Working as expected!

dbauszus-glx:
I added some missing dictionary entries for Spanish registration.

image

RobAndrewHurst (approved):

PR #2381, opened 2025-09-03
bug(geometry entry): Add tooltip to Geometry Entry

simon-leech:
We mistakenly didn't include a tooltip option on `geometry` entries but we did for all entries that use a title. 
This PR adds a tooltip option to the geometry entry method. 

Screenshot 2025-09-03 085225


dbauszus-glx:
Not a patch; and shouldn't be resolved in the geometry entry. I'll update the issue.

PR #2382, opened 2025-09-08
Workspace checksum

RobAndrewHurst:
I have implemented a workspace check sum value that gets added to the workspace whenever we cache it.

### DIFF
--- a/mod/workspace/cache.js
+++ b/mod/workspace/cache.js
@@ -7,6 +7,7 @@ Default templates can be overwritten in the workspace or by providing a CUSTOM_T
 @requires /provider/getFrom
 @requires /utils/merge
 @requires module:/utils/processEnv
+@requires crypto
 
 @module /workspace/cache
 */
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
In order to test this in the client we need to add the checksum to the locale returned from the getLocale request.

This will allow us to check whether a locale has changed at rest if a user locale has been stored with a checksum.

RobAndrewHurst:
@dbauszus-glx 

I have added the check sum value of a locale for each locale in a workspace.

There is also added logic to recalculate the check sum when we merge or have a nested locale.

dbauszus-glx:
@RobAndrewHurst The tests are failing after updating the logic in regards to user roles.

RobAndrewHurst:
> @RobAndrewHurst The tests are failing after updating the logic in regards to user roles.

Yeah the tests need to be written for your new logic that you added in.

RobAndrewHurst:
@dbauszus-glx I have updated the tests and I am merging it in.

PR #2383, opened 2025-09-08
Enable caching of cloudfront templates

dbauszus-glx:
This PR introduces the caching of cloudfront resources in a cachedMap.

Cloudfront resources should only be loaded if needed since the templates may change at rest and it is not possible to uncache resources in cloud functions.

The workspace test and the roles request require all templates to be loaded in order to check the templates for roles.

The cacheTemplates method of the workspace api module can thefore send a cache flag parameter.

This parameter flag must be passed the getLocale and getLayer methods, via the mergeTemplates, to the getTemplates method and from there to the cloudfront provider.

With the cache param set to null I see that many templates and fetched and parsed multiple resulting in the workspace caching to take about 17 seconds which will fail deployed to vercel.

https://github.com/user-attachments/assets/b0a08877-3018-40ec-8160-3e7c2bdb6873

With caching enabled I can load the same workspace in about 5 seconds.

https://github.com/user-attachments/assets/230b845c-ecc1-46b9-b93a-09d82e4f86ed

### DIFF
--- a/mod/provider/_provider.js
+++ b/mod/provider/_provider.js
@@ -7,7 +7,6 @@ Functions for handling 3rd party service provider requests
 
 import cloudfront from './cloudfront.js';
 import file from './file.js';
-
 import s3 from './s3.js';
 
 /**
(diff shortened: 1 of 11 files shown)


dbauszus-glx:
Please note that there is still quite a bit of documentation required along with some preliminary testing.

dbauszus-glx:
The caching logic has been moved to the getFrom module. The signed URL and buffer options have been removed from the provider cloudfront logic since the sign endpoint is to be specifically used for this.

simon-leech (approved):
This seems to be working fine for me - tested a couple of applications using both local files and also using cloudfront env keys and working as expected. No errors or anything going on

PR #2385, opened 2025-09-09
add(elements): tooltip

cityremade:
Adds mapp.ui.elements.tooltip module.

```js
const tooltip = mapp.ui.elements.tooltip({
   content: "My tooltip text"
});
```

Tooltip is an inline span document fragment with "help" symbol which shows plain text info on mouseover.
Can be included within any other element, at the moment only included in standalone title entry for review.

In order to view tooltip use configuration like the following:

```json 
{
   "title": "Hello this is a title entry",
   "tooltip": "some tooltip on hover"
}
```

To clarify: where else this should be available/optional

checkboxes?
editable fields?

At the moment `entry.tooltip` is used as immediate text with no fallback.

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -536,6 +536,7 @@ Adds a "question mark" symbol to indicate available information.
 Symbol will be added to label tag or to a div.label container if either is found in the entry node.
 
 @param {infoj-entry} entry
+@property {string} entry.tooltip The content to be shown in a tooltip.
 */
 function entryTooltipIcon(entry) {
   if (!entry.tooltip) return;


cityremade:
Screenshot 2025-09-12 at 17 52 35

```json
{ 
    "title": "Hello this is a title entry",
    "tooltip": "some tooltip on hover",
    "type":"title"
}
```

Apparently skipping `field` is sufficient to create entry as title alone. When set both `title` and `type: title` we get it twice.

cityremade:
Tooltip symbol now added to label tag (checkboxes) or .label container if either is found.
Screenshot 2025-09-15 at 12 52 51

simon-leech (approved):
Fix a little issue where the check should be `label_element`. 

Happy with this now - tested on a range of entry types including geometries, dataviews, titles, booleans etc.

PR #2392, opened 2025-09-16
Unnest panelClasslist template

dbauszus-glx:
This draft PR adds console.logs for the entry.panel and entry.target properties in the tabview addTab(entry) method.

The aim is to unnest the classList template used in the construction of the panel property element.

I do not understand when this is required in order to test this and remove the console logs.


dbauszus-glx:
The panel class issue has already been resolved in this PR. https://github.com/GEOLYTIX/xyz/pull/2342

Issue #1854, opened 2025-01-29
Ability to Control Display of HelpDialog


simon-leech:
We need a way to store a flag e.g. `hideHelpDialog: true` that allows the user to control whether or not the helpDialogs are shown.

dbauszus-glx:
At first it must be possible to store user settings in an UserIndexedDB.

https://github.com/GEOLYTIX/xyz/issues/2113

simon-leech:
This needs picking up in the next release. 
Using a draw method in a custom view opens the helpDialog on top of the content, but unlike in the default view, the whole content is masked (disabled) until you close it. 

I think we need to make it conditional to turn off on a layer level (default on) but if a locale flag is provided - this is passed to every layer with a draw object assigned

simon-leech:
This should be a flag that can be set on a locale, layer or entry level from either configuration or via a userLocale stored flag. 
This is as a custom view using a draw method opens a helpDialog so we want the ability to turn this off via configuration when needed, or for the user to do it via a button for themselves.

PR #2402, opened 2025-09-18
feat(helpDialog): Control help dialog display. 

AlexanderGeere:
## Description
This PR adds a plugin and a utility for controlling whether helpDialogs are displayed. 

The plugin adds a button the button column:
screenshot-2025-09-18_15-26-50

When clicked adds `hideHelp` to all the layers in the mapview where applicable and reloads said layers. 

`hideHelp` will also be applied to the locale and written to the indexDB

The plugin is called with `disable_help: {}` in the locale. 
## GitHub Issue
#1854 

## Type of Change
- ✅ Documentation

## How have you tested this?
Tested Locally on `bugs_testing\plugins\process_pipe.js`. Can also be tested by simply adding
`disable_help: {}` to any locale.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/plugins/_plugins.mjs
+++ b/lib/plugins/_plugins.mjs
@@ -8,6 +8,7 @@ The module exports a collection of core mapp plugins.
 @requires module:/plugins/admin
 @requires module:/plugins/feature_info
 @requires module:/plugins/fullscreen
+@requires module:/plugins/disable_help
 @requires module:/plugins/link_button
 @requires module:/plugins/locator
 @requires module:/plugins/login
@@ -21,6 +22,7 @@ The module exports a collection of core mapp plugins.
 
 import { admin } from './admin.mjs';
 import { custom_theme, dark_mode } from './dark_mode.mjs';
+import { disable_help } from './disable_help.mjs';
 import { feature_info } from './feature_info.mjs';
 import { fullscreen } from './fullscreen.mjs';
 import { link_button } from './link_button.mjs';
@@ -51,6 +53,7 @@ import { zoomToArea } from './zoomToArea.mjs';
 const plugins = {
   admin,
   custom_theme,
+  disable_help,
   dark_mode,
   feature_info,
   fullscreen,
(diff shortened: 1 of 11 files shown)


dbauszus-glx (changes requested):
You can ensure that the helpDialog receives a params object like so:

```
helpDialog(params={})
```

You do not need to assign locale to layer, locale to entry, layer to entry etc.

The help dialog checks for the flag on the user if the user has the flag to hide the method shortcircuits.

An entry or draw method can have the flag which will be applied even if the user doesn't have the flag.

You don't need an extra plugin and plugin element to test user flags. There is a debug plugin which allows you to control the user properties stored in local storage.

We can discuss the need to assign default user properties through a user property on a locale but that will be a different PR.

AlexanderGeere:
> You can ensure that the helpDialog receives a params object like so:
> 
> ```
> helpDialog(params={})
> ```

I believe this is done as a way to close the dialog as the object only exists in the context of the module.

> 
> You do not need to assign locale to layer, locale to entry, layer to entry etc.
> 
> The help dialog checks for the flag on the user if the user has the flag to hide the method shortcircuits.
> 
> An entry or draw method can have the flag which will be applied even if the user doesn't have the flag.
> 
> You don't need an extra plugin and plugin element to test user flags. There is a debug plugin which allows you to control the user properties stored in local storage.
> 
> We can discuss the need to assign default user properties through a user property on a locale but that will be a different PR.

This is now done through mapp.user.

The draw and geometry module still assign a hideHelp property to the helpDialog as well.

cityremade:
I can see help control flag called only inside geometry entry. Is this correct?

dbauszus-glx (approved):

dbauszus-glx:
This is actually a fairly simple process.

mapp.user.hideDialog will never show any dialog. The helpDialog element method will short circuit.

interaction.hideDialog, eg. `polygon: {hideDialog: true}` will not show the hideDialog for the layer, location polygon draw interaction.

`draw.hideDialog` will not show the dialog for any drawing interaction.

RobAndrewHurst (approved):

PR #2405, opened 2025-09-19
Move dotenv config to processEnv

dbauszus-glx:
The dotenv import must be in the processEnv module for use in deployments which do not use the express app, eg. vercel.


dbauszus-glx:
Moved to dependencies and bumped to 17.2.2

cityremade (approved):

PR #2406, opened 2025-09-22
feat(radio): Radio element and radiogroup init

cityremade:
Creates:

- ui element radio
- location entry radiogroup
- radio element stylesheet

Possible configuration requires **at least 2 boolean fields** meant to be mutually exclusive.

Checking one unchecks all others.

Use case: select model type and use only relevant input - conditionals within on change callbacks at a further stage.

```json
{
  "field":"model_food",
  "title":"Model Type",
  "label":"Food",
  "type":"radiogroup",
  "edit":true,
  "radiogroup":{
    "name":"my_model_type",
    "caption":"Select model type to run for this site"
  }
},
{
  "field":"model_clothes",
  "label":"Clothes",
  "type":"radiogroup",
  "edit":true,
  "radiogroup":{
    "name":"my_model_type"
  }
},
{
  "field":"model_home",
  "label":"Home",
  "type":"radiogroup",
  "edit":true,
  "radiogroup":{
    "name":"my_model_type"
  }
}
```
Screenshot 2025-09-22 at 15 33 17

### DIFF
--- a/lib/plugins/admin.mjs
+++ b/lib/plugins/admin.mjs
@@ -26,7 +26,7 @@ export function admin(plugin, mapview) {
   if (!mapp.user?.admin) return;
 
   btnColumn.appendChild(mapp.utils.html.node`
-    <a
+    <a data-id="admin-view"
       title=${mapp.dictionary.toolbar_admin}
       href="${mapp.host + '/api/user/admin'}">
       <span class="notranslate material-symbols-outlined">supervisor_account`);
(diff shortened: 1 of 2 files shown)


cityremade:
Radiobutton is a boolean at its root. It only knows if it's on or off.
Think about it as a group of mutually exclusive checkboxes. 

Value updates and callbacks should be handled with onchange radiogroup event and will be added later outside this pr.

dbauszus-glx (changes requested):
This doesn't match the specification of the ticket. I'd like to see the radio element first as a susbtitute for a single select dropdown element before creating a complex multiple related boolean fields use case and then finding a problem for this solution.

PR #2408, opened 2025-09-24
Templates Properties

RobAndrewHurst:
This is a WIP PR for the Templates properties issue.

### DIFF
--- a/mod/workspace/mergeTemplates.js
+++ b/mod/workspace/mergeTemplates.js
@@ -33,7 +33,7 @@ The method will check for a template matching the obj.key string property if obj
 
 An array of templates can be defined as obj.templates[]. The templates will be merged into the obj in the order the template keys are in the templates[] array.
 
-@param {Object} obj 
+@param {Object} obj
 @param {array} [roles] An array of user roles from request params.
 @param {boolean} cache Templates should be cached and not requested multiple times.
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
The exclude check must happen before the include check as otherwise the exclude property itself would be removed from the template object.

dbauszus-glx (approved):

simon-leech (changes requested):
This is working well for single template usage! 

I have a example whereby the same template is loaded (using a different key) but the exclude / include is being applied to both incorrectly. It should only be applied to `layer_a` in this configuration example below.


``` json
"layer_a": {
  "template": {
    "key": "layer_a",
    "src": "${CLIENT}/layers/layer.json",
    "exclude_props": ["style"]
  }
},
"layer_b": {
   "roles": {
     "Edit": null
   },
  "template": {
    "key": "layer_b",
    "src": "${CLIENT}/layers/layer.json"
  }
}
```

RobAndrewHurst:
@simon

> This is working well for single template usage!
> 
> I have a example whereby the same template is loaded (using a different key) but the exclude / include is being applied to both incorrectly. It should only be applied to `layer_a` in this configuration example below.
> 
> ```json
> "layer_a": {
>   "template": {
>     "key": "layer_a",
>     "src": "${CLIENT}/layers/layer.json",
>     "exclude_props": ["style"]
>   }
> },
> "layer_b": {
>    "roles": {
>      "Edit": null
>    },
>   "template": {
>     "key": "layer_b",
>     "src": "${CLIENT}/layers/layer.json"
>   }
> }
> ```

I will look into this @simon-leech

simon-leech (approved):
Seems to be working as expected for me!

AlexanderGeere (approved):

PR #2409, opened 2025-09-24
Toolbar plugin - initial commit;

cityremade:
Hackathon September 2025!


RobAndrewHurst:
@cityremade Can we move this to the plugins repo first so we can work through all the complications and nuances before bringing it into core?

PR #2415, opened 2025-09-24
feat(distributed): `showLegend` for `type: distributed`; distributed theme on point features

simon-leech:
# Pull Request Template

## Description

Please include a summary of the changes, and the motivation for the change.

## GitHub Issue

## Type of Change

- ✅ Documentation
- ✅ Testing

## How have you tested this?

Please describe how to test this change and how you verified it worked.

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/layer/featureStyle.mjs
+++ b/lib/layer/featureStyle.mjs
@@ -45,7 +45,11 @@ export default function featureStyle(layer) {
 
     if (Object.hasOwn(mapp.layer.themes, layer.style.theme?.type)) {
       // Apply theme style to style object.
-      mapp.layer.themes[layer.style.theme?.type]?.(layer.style.theme, feature, layer);
+      mapp.layer.themes[layer.style.theme?.type]?.(
+        layer.style.theme,
+        feature,
+        layer,
+      );
     }
 
     // The feature must not be displayed.
--- a/lib/layer/featureStyle.mjs
+++ b/lib/layer/featureStyle.mjs
@@ -45,7 +45,7 @@ export default function featureStyle(layer) {
 
     if (Object.hasOwn(mapp.layer.themes, layer.style.theme?.type)) {
       // Apply theme style to style object.
-      mapp.layer.themes[layer.style.theme?.type]?.(layer.style.theme, feature);
+      mapp.layer.themes[layer.style.theme?.type]?.(layer.style.theme, feature, layer);
     }
 
     // The feature must not be displayed.
--- a/lib/layer/themes/distributed.mjs
+++ b/lib/layer/themes/distributed.mjs
@@ -1,33 +1,36 @@
 /**
-### mapp.layer.themes.distributed()
-This module exports a function that applies a unique value theme to a feature based on a specified field value.
+## /layer/themes/distributed
+
+The layer theme module exports a method for distributed themes.
+
 @module /layer/themes/distributed
- */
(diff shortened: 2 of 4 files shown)


simon-leech:
Need to merge #2422  before this can be continued as touching the same files.

AlexanderGeere (approved):

dbauszus-glx:
@simon-leech Please review the imports inside lib/ui/layers/_layers. It feels odd that we now import the legends module, but also the basic legend seperately which is also imported in the legends module and used in the style object. The styles object with the warning can probably be retired.

image

The featureStyle should not require to send the layer argument to the theme method.

image

I reverted the theme method and added the assignment of the values array as outlined in the ticket associated with this PR.

```js
    theme.lookup[val] = theme.categories[theme.index];
    theme.categories[theme.index].values ??= [];
    theme.categories[theme.index].values.push(val);
```

You can not call the legend method from a style method. Openlayers style methods will absolutely kill the performance of the application since the render should not wait. The bbox check is already putting the performance for the style render to the brink.

You need to call the distributed theme legend method from the postrender event of the layer. I added an example for you. The method will call itself once from the postrender event which will assign the event listener again.

```js
export default function distributedTheme(layer) {
    console.log(layer.style.theme.categories)
    layer.L.once('postrender', ()=>{
        distributedTheme(layer)
    })
}
```

The legend should show the array of associated values for each categorie. You can then add a whitelist to show ellipses if the array is too long.

image

The distributed theme legend should not be mushed into the categorical theme legend module. This is messy and hard to debug.

Please use themes/distributed module with the log and create shared utils method for items that are used in the categorical, distributed, and perhaps graduated theme as well.

The flag should be to hide legend, not showLegend is distributed. Otherwise this will become messy should we decide to hide other legends. The mapp logic is to show legends by default. Hence hideLegend would be the optional flag. Not some legends are shown by default others are hidden by default with flags to reverse the logic.

Here is a video showing the original logic creating an infinity legend and breaking the render.


https://github.com/user-attachments/assets/ace68665-3c99-48cf-a94c-56616403042c

dbauszus-glx:
Almost there. I'd use ellipsis to shorten the label.

image

Please have a look at my last commit. https://github.com/GEOLYTIX/xyz/pull/2415/commits/f5088ca3c2f8fd34e3d4dd2d088a2e4de526335d

I just join the values to a label.

```js
  for (const cat of theme.categories) {

    cat.label = cat.values?.join(', ')

    createLegend(
      cat,
      theme,
      layer,
    );
  }
```

I call the method in the postrender listener before returning. This reduces a lot of nesting and makes the method easier to debug. 
```js
  layer.L.once('postrender', () => {

    distributedTheme(layer)
  });

  return theme.legend.node;
```

I renamed the legendHelper module to utils since these are utils in the context of the legend folder. These should not be assigned as a type of legend. I also changed the export rather than exporting a default object with all the method I export the individual methods and import them where needed like so:

```
import { clusterLegend, createLegend, distributedLegend, themeParser } from './utils.mjs'
```

RobAndrewHurst (approved):

simon-leech:
I'm happy with the way this is working now too! Tried to test everything and all looks good!

PR #2431, opened 2025-10-02
feat(radio): ui element

cityremade:
Adds a radio element and its own css stylesheet.

This is an extract from previous pr narrowed down to a new ui element only.

Structure is in line with checkbox element.

In order to create a radio

`const my_radio = mapp.ui.elements.radio({ label: "My only option" });`

In order to create radio buttons as a group:

```js
const my_radio1 = mapp.ui.elements.radio({ label: "My option 1", name: "my_group" });
const my_radio2 = mapp.ui.elements.radio({ label: "My option 2", name: "my_group" });
const my_radio3 = mapp.ui.elements.radio({ label: "My option 3", name: "my_group" });
```

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -178,7 +178,7 @@ A callback function can be provided to override the execution of methods in the 
 */
 async function update(callback) {
   if (this.infoj.some((entry) => entry.invalid)) {
-    alert(`Unable to update location with invalid entry value.`);
+    mapp.ui.elements.alert({ text: mapp.dictionary.invalid_entry_value });
     return new Error('Unable to update.');
   }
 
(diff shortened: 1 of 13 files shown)


cityremade:
Now supporting text options as radio group
configuration just like options + radio: true flag.
In active colour when value has been changed.

**On load**
Screenshot 2025-10-06 at 11 59 59

**On change**
Screenshot 2025-10-06 at 12 00 05

simon-leech:
> Now supporting text options as radio group configuration just like options + radio: true flag. In active colour when value has been changed.
> 
> **On load** Screenshot 2025-10-06 at 11 59 59
> 
> **On change** Screenshot 2025-10-06 at 12 00 05


@cityremade Is it possible to just have the selected one that's changed be orange, not the whole options? So its really clear that its been changed to that value?

cityremade:
@simon-leech in reality they've all been changed as they are a group.

AlexanderGeere (changes requested):
I think the options should work like dropdown, If you provide an empty options array the distinct query is used to populate the options.

cityremade:
The foo bar test failed as existing value didn't match any of the values in the options list.
In this case current value will not be set in the element.

cityremade:
@AlexanderGeere empty array gives distinct values 
Screenshot 2025-10-10 at 13 43 29

PR #2435, opened 2025-10-03
Patch for distributed themes

dbauszus-glx:
The value must be parsed as a string. Otherwise the `!val` condition will shortcircuit on an integer or numeric properties.

Object keys are always type string and the lookup cannot work if the val is not type string.

### DIFF
--- a/lib/layer/styleParser.mjs
+++ b/lib/layer/styleParser.mjs
@@ -342,6 +342,12 @@ The parseTheme method checks whether a theme and it's categories have consistent
 function parseTheme(theme, layer) {
   if (typeof theme !== 'object') return;
 
+  // Check on if the theme.type exists.
+  if (!Object.hasOwn(mapp.layer.themes, theme.type)) {
+    console.warn(`Layer: ${layer.key}; Theme Type: ${theme.type} unavailable.`);
+    return;
+  }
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
The theme.index should be assign or increased before needed. A point has no bbox extent all the intersect checks and filter uinnecessary. Simply assign the next theme.index from categories as style and return.

```
  // Define as 0 or increase theme.index
  if (theme.index === undefined) {
    theme.index = 0;
  } else {
    theme.index++
  }

  // Reset cat index to 0 if the index reaches the length of the cat array.
  if (theme.index === theme.categories.length) {
    theme.index = 0;
  }

  // Get feature bounding box from geometry extent.
  const bbox = {
    extent: feature.getGeometry().getExtent(),
  };

  // A point feature will have matching bbox corner coordinates.
  if (bbox[0] === bbox[2] || bbox[1] === bbox[3]) {

    // Assign the style to the lookup object for the feature field property value.
    theme.lookup[val] = theme.categories[theme.index];

    feature.style = theme.lookup[val].style;
    return;
  }
```

AlexanderGeere (approved):

cityremade (approved):

PR #2442, opened 2025-10-07
fix(entry - tooltip): Check if tooltip exists and remove if so

simon-leech:
## Description
When a `type:dataview`, that is `display:true` is reloaded after a different `infoj` field is updated, the tooltip is being duplicated.

## Type of Change

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/locations/infoj.mjs
+++ b/lib/ui/locations/infoj.mjs
@@ -547,6 +547,14 @@ function entryTooltipIcon(entry) {
 
   if (!label_element) return;
 
+  // Check if tooltip already exists, if so remove as its content may have changed.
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
I updated the method to shortcircuit if the tooltipElement exists.

No need to lookup, remove, recreate and add the element.

```js
  // The tooltipElement already exists.
  if (entry.tooltipElement) return;

  entry.tooltipElement = mapp.ui.elements.tooltip({
    content: entry.tooltip,
  });

  label_element.append(entry.tooltipElement);
```

PR #2452, opened 2025-10-09
Update documentation for logger module

dbauszus-glx:
This PR updates the documentation for the logger module.

A list of the available LOG keys has been added to the module documentation.

image


simon-leech:
@dbauszus-glx  Would it be possible to add a short description of what each means? 
For example what's the difference between req and req-url?

dbauszus-glx:
@simon-leech Yes, please do.

AlexanderGeere:
@simon-leech 
I added descriptions for what gets logged.

@dbauszus-glx 
I believe the fetch option is no longer applicable as we use node:fetch now.

AlexanderGeere:
The logger call in the mailer will now log the messageId and the from and to, or if mailer_body is specified you get the body of the email as well.

Issue #2446, opened 2025-10-07
key authentication


dbauszus-glx:
Currently it is possible to use an api key as token property to authenticate a request eg. for the provider/file method. The key is tied to a user in the ACL.

We would like a key authentication which is not tied to a user but to a pem key file in the root, like the KEY_CLOUDFRONT authentication in the cloudfront signer module.

This should allow one instance with the right key file to request private files eg in the resources directory from another instance by providing the key in the request which will be checked against the PEM file in the root of the instance which has the resource file.

PR #2453, opened 2025-10-09
feat(sign): Add a file signing endpoint

AlexanderGeere:
## Description
Provides a file signing endpoint which can be called from another instance using a key file as authentication. The key will not be tied to a user.

The key should be put in the root directory of the app and referenced in the environment settings as `KEY_LOCAL_FILE: XXXXX`. You should also supply a `SIGN_LOCAL_FILE`, which should be the instances url.

In the case of setting up repo to be used by other instance this should be `KEY_<repo>_<name>` accompanied by a `SIGN_<repo>_<name>`

You will need a FILE_RESOURCES key in your environment e.g `FILE_RESOURCES: resources`, as this is the folder the endpoint will be allowed to access.

example usage of the endpoint:
http://localhost:3000/latest/api/sign/file?key=./resources/example/file/layer.json 

Should yield:
https://localhost:3000/api/provider/file?expires=xxxxxx&key_id=xxxx&signature=xxxxx&url=.%2Fresources%2Fexample%2Ffile%2Flayer.json

## GitHub Issue
#2446

## Type of Change
- ✅ Documentation
- ✅ Testing

## How have you tested this?
To test generate a key: 
`openssl genrsa -out private-key.pem 3072`
or
https://emn178.github.io/online-tools/rsa/key-generator/
and paste the private key into a file

Add a KEY_LOCAL_FILE variable to your enironment as such: `KEY_LOCAL_FILE: private-key`
Add a SIGN_LOCAL_FILE variable to your environment e.g. : `SIGN_LOCAL_FILE: localhost:3000/latest`
 
## Testing Checklist
- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/mod/user/register.js
+++ b/mod/user/register.js
@@ -180,10 +180,12 @@ function checkUserBody(req, res) {
   // Test whether email domain is allowed to register
   if (xyzEnv.USER_DOMAINS) {
     // Get array of allowed user email domains from split xyzEnvironment variable.
-    const allowed_domains = xyzEnv.USER_DOMAINS.split(',');
+    // Note this should be converted to lower case for comparison, as email domains are case insensitive.
+    const allowed_domains = xyzEnv.USER_DOMAINS.toLowerCase().split(',');
 
     //  Get the user_domain from user email.
-    const user_domain = req.body.email.split('@')[1];
+    // Convert to lower case for comparison, as email domains are case insensitive.
+    const user_domain = req.body.email.split('@')[1].toLowerCase();
 
     // Check whether not some of the allowed_domain includes the user_domain.
     if (!allowed_domains.some((domain) => user_domain.includes(domain))) {
--- a/tests/mod/user/register.test.mjs
+++ b/tests/mod/user/register.test.mjs
@@ -115,6 +115,30 @@ await codi.describe(
         codi.assertTrue(res.statusCode === 200);
       },
     );
+
+    await codi.it(
+      {
+        name: 'USER_DOMAINS - ensure case insensitivity',
+        parentId: 'user_register',
+      },
+      async () => {
+        globalThis.xyzEnv.USER_DOMAINS = 'geolytix';
+
+        const { req, res } = codi.mockHttp.createMocks({
+          headers: {
+            host: 'localhost',
+          },
+          body: {
+            email: 'dev_1@GEOLYTIX.COM',
+            password: 'ValidPass123!',
+            language: 'en',
+          },
+        });
+
+        await register(req, res);
+        codi.assertTrue(res.statusCode === 200);
+      },
+    );
   },
 );
 


dbauszus-glx:
I am unsure how to call the file in a template src string.

I create a private test deployment with a private key file which has a private json resource.

https://signed-file-provider.vercel.app/api/provider/file?url=./resources/scratch.json

This is a layer template which I want to call securely in a local [or deployed] instance.

```
{
  "locale": {
    "layers": {
      "OSM": {
        "display": true,
        "format": "tiles",
        "attribution": {
        }
      },
      "scratch": {
        "display": true,
        "dbs": "NEON",
        "template": {
            "src": "https://signed-file-provider.vercel.app/api/provider/file?url=./resources/scratch.json"
        }
      }
    }
  }
}
```

How will I need to set the `src` string to get the template into this workspace?

AlexanderGeere:
@dbauszus-glx 
 > ```
> {
>   "locale": {
>     "layers": {
>       "OSM": {
>         "display": true,
>         "format": "tiles",
>         "URI": "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png",
>         "attribution": {
>           "© OpenStreetMap": "http://www.openstreetmap.org/copyright"
>         }
>       },
>       "scratch": {
>         "display": true,
>         "dbs": "NEON",
>         "template": {
>             "src": "https://signed-file-provider.vercel.app/api/provider/file?url=./resources/scratch.json"
>         }
>       }
>     }
>   }
> }
> ```
> 
> How will I need to set the `src` string to get the template into this workspace?

I have modified how the endpoint works slightly. I have added a redirect param and a host param. 
So in the above case `src` would be:

`https://<current_instance>/api/sign/file?key=./resources/scratch.json&host=signed-file-provider.vercel.app&redirect=true`

dbauszus-glx:
`https://<current_instance>/api/sign/file?key=./resources/scratch.json&host=signed-file-provider.vercel.app&redirect=true`

This would work but will require a lot variable replacement in the configuration.

The solution should be more in line with the cloudfront provider/signer.

We can get a cloudfront resource from a src string like this:

`"src": "cloudfront:templates.geolytix.io/v4.13.0/express/help_centre/help_centre.html"`

However with cloudfront there is only one key. It is potentially required to get resources from different XYZ instances hence we need to associate the key with a signer endpoint.

A potential solution could work with following variables.

```
"KEY_MAPP_TEMPLATES": "2453",
"SIGN_MAPP_TEMPLATES": "signed-file-provider.vercel.app"
```

You can iterate through the variables in the _sign module and add signer where a `KEY_` matches a `SIGN_` variable.

image

You then add a signer endpoint to be exported to the provider.

In the workspace the src string could be defined like this.

`"src": "MAPP_TEMPLATES:resources/scratch.json"`

AlexanderGeere:
@dbauszus-glx 
> I was unable to get this working with codespaces. I create a codespace and pull the PR.
> 
> I create a workspace.json and a scratch.json in the resources directory.
> 
> I know launch localhost with the same key and the codespaces address.
> 
> ```
>         "KEY_MAPP_TEMPLATES": "2453",
>         "SIGN_MAPP_TEMPLATES": "bookish-xylophone-xr5r7qpv4vxcvpx7-3000.app.github.dev"
> ```
> 

The codespaces env will also need these keys. The key_id parameter that gets sent is the name of the key file, the sign endpoint will set key_id to KEY_MAPP_TEMPLATES.

dbauszus-glx:
Please note that codespaces will make a private if an instance is restarted. The port must be set to public in order to access the instance from another instance.

image

Please not the file ending in a key file. 

I copied the key from https://emn178.github.io/online-tools/rsa/key-generator/  into a new file.

image

The default End of Line Sequence on my Windows localhost was CRLF, but LF in the codespaces test instance. That was the reason why the signature was different.

dbauszus-glx:
@AlexanderGeere I shaved some of the complexity off by using the xyzEnv key as key name.

https://github.com/GEOLYTIX/xyz/pull/2453/commits/394fec83ba52b724952d93cefb5325df75ce6b38

```
"src": "MAPP_TEMPLATES:resources/scratch.json"
```

requires

```
"SIGN_MAPP_TEMPLATES": "bookish-xylophone-xr5r7qpv4vxcvpx7-3000.app.github.dev"
```

with the key named `MAPP_TEMPLATES.pem`

I event consider to just use SIGN_MAPP_TEMPLATES as keyname and directive in the src string. Makes much easier to associate key and value with directive.

dbauszus-glx:
@AlexanderGeere the readdirSync must be tried to return the 405.

The fetch response need to be checked to be OK or return an error otherwise the parsing may attempt json from error message which crashes the process.

```js
      const response = await fetch(signedUrl, {
        headers: {
          'Content-Type': contentType,
        },
      });

      if (!response.ok) {
        return new Error(`Failed to fetch`);
      }
```

https://github.com/GEOLYTIX/xyz/pull/2453/commits/3de9939359b43e678be27d39cdbfd72f0b389b21

dbauszus-glx:
@RobAndrewHurst let's ignore the file test until the process is optimised.

dbauszus-glx:
@AlexanderGeere 

The file provider method should only return files from the filesystem.

I created an async XYZ method in the getFrom module which returns a file from XYZ instance by signing the request.

This makes the assignment easier to read and the methods easier to debug as they do one thing only.

```js
const getFromModules = {
  cloudfront: Cloudfront,
  file: File,
  https: Https,
};

// Assign XYZ method for each SIGN_* key in xyzEnv
for (const key in xyzEnv) {
  const PROVIDER = new RegExp(/^SIGN_(.*)/).exec(key)?.[1];
  if (PROVIDER === undefined) continue;
  getFromModules[PROVIDER] = XYZ;
}

export default getFromModules;
```

https://github.com/GEOLYTIX/xyz/pull/2453/commits/aa9e14eb7cdcf42cbe8074232b981642164c779b

dbauszus-glx:
The key files can be read in the processEnv module and be added to a wallet there.

https://github.com/GEOLYTIX/xyz/pull/2453/commits/0cec1bd3482b280034c8c9bf88ec0a227ef612f5
https://github.com/GEOLYTIX/xyz/pull/2453/commits/26644ec9ff50da94af23d9361b35b31ca8929577

dbauszus-glx (approved):

dbauszus-glx:
FILE_RESOURCES: 'resources' should be assigned as default

The provider file must try for this otherwise the error thrown will chuck the process.

dbauszus-glx (approved):

dbauszus-glx:
The templates can by sync by awaiting for all template promises to be settled.

```js
  logger(`cachelocales: ${Date.now() - timestamp}`, 'cachelocales');

  const templatePromises = Object.keys(workspace.templates).map(async (key) => {
    await getTemplate(key, true);
  });

  await Promise.allSettled(templatePromises);

  logger(`cachetemplates: ${Date.now() - timestamp}`, 'cachetemplates');
```

I updated and added to logger.

cachelocales - logs the time for all locales to be cached.
cachetemplates - logs the time for all locales and templates to be cached.
xyzfetch - logs the time for fetch from an XYZ signed provider
cloudfront - logs the time it takes to fetch a resource from cloudfront

I can see in my test that the first time it takes longer to fetch templates from cloudfront, but sucessive calls are faster.

image

The performance against the XYZ-TEMPLATES instance is not much difference looking at the calls now.

image

PR #2468, opened 2025-10-17
Set locale hook in default script

dbauszus-glx:
The default script sets the locale hook.


simon-leech (approved):

cityremade:
does it matter if there's just one locale?

dbauszus-glx:
@cityremade  as in "locale" or one locale in "locales"?

PR #2470, opened 2025-10-17
Fix spacing in {{title}} variable

dbauszus-glx:
A previous minor commit introduced false spacing for the title variable in the _default view markup.


RobAndrewHurst:
image

This PR will be resolved after we upgrade and add in a setting for the html formatter.

PR #2471, opened 2025-10-17
chore(biome): v2.2.6 bump & enable html parsing interpolation

RobAndrewHurst:
# Biome v2.2.6

I have bumped the biome package as it helps address parsing double text expressions that we have in our html views.

eg: `{{title}}`

Enabling this to be parsed in the formatter will prevent it from being formatted in biome (in html files).


dbauszus-glx (approved):

dbauszus-glx:
I closed https://github.com/GEOLYTIX/xyz/pull/2470 in favour of this PR.

Issue #2450, opened 2025-10-08
Control tab extension


jfitzpatrick17:
It should be possible to extend the control tab so that we can have more tabs available - this could be for plugins or something else.

As an example, we might want to split search and locations:

https://github.com/user-attachments/assets/2a1418cd-6568-454a-a496-3291a509bd21

In the future it could also house a 'data catalogue'

dbauszus-glx:
This is the perfect chance to make the #ctrl tabs element an actual ui element. Currently the behaviour is mostly controlled directly in the _default view script.

```js
    const locationsTab = document.getElementById('locations');
    const layersTab = document.getElementById('layers');

    const tabs = document.querySelectorAll('#ctrl-tabs > div');
    const tabPanels = document.querySelectorAll('#ctrl-panel > div');

    tabs.forEach((tab) => {
      // Set help text from dictionary.
      tab.title = mapp.dictionary[tab.dataset.id];

      tab.onclick = (e) => {
        // Change active class for the tab.
        tabs.forEach((el) => el.classList.remove('active'));
        e.target.classList.add('active');

        // Change active class for the panel.
        tabPanels.forEach((el) => el.classList.remove('active'));
        document.getElementById(e.target.dataset.id).classList.add('active');

        // Put focus on the gazetteer if the locations tab is activated.
        if (e.target.dataset.id === 'locations') {
          let gazetteerInput = document.getElementById('gazetteerInput');
          gazetteerInput && window.innerWidth > 768 && gazetteerInput.focus();
        }
      };
    });
```

simon-leech:
+1 on this as then we could use the control tab in custom views much more easily and flexibly!

dbauszus-glx:
This has now been implemented in the linked PR.

PR #2486, opened 2025-11-05
feat(Controls): Adds a way to add more options to the `ctrls` element. 

AlexanderGeere:
## Description
This PR introduces a `ui.elements.controls` element as well as  a new utility called `controls`. 

The default view now calls `mapp.ui.elements.controls` to create the control tabs and the panel associated to the tab. The default controls are `layer` and `locations`, If no controls are specified in the configuration.   

custom controls can also be created by specifying a plugin and a control as follows:

```json
"locale": {
   "plugins": ["${LOCAL}/bugs_testing/test/my_control.js"]
   "controls": ["layers",
                  {"key": "my_control", "icon": "history"}, 
                  {"key": "locations", "icon": "location_on", "style": {"display": "block"}}]
   "my_control": {}
}
```

and then in the plugin.

```javascript
mapp.controls.my_control = () => {console.log('This is a custom control')}
```

This would yield a `ctrls` element that looks as follows:
screenshot-2025-11-05_15-41-46

## GitHub Issue
#2450 

## Type of Change
- ✅ Documentation

## How have you tested this?
This can be tested on `bugs_testing/dataviews/dataviews.json`

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ docs/*
 *.pem
 *.crt
 /*.json
+/*.jsonc
 !tests/*.json
 !tests/**
 !package.json
--- a/mod/provider/getFrom.js
+++ b/mod/provider/getFrom.js
@@ -3,6 +3,8 @@
 
 The getFrom provider module allows XYZ modules to get resources either from configured sources.
 
+Cloudfront resources get cached for 60seconds to prevent excessive requests for the same resource.
+
 @requires ../sign/file
 @requires ../utils/logger
 @requires ./cloudfront
@@ -32,6 +34,7 @@ for (const key in xyzEnv) {
 export default getFromModules;
 
 const cacheMap = new Map();
+let cacheTime = Date.now();
 
 /**
 @function Cloudfront
@@ -45,11 +48,10 @@ The fetch request will be created from the cloudfront provider module with the c
 The fetch request will be stored in a cache Map object for requests from the [cacheTemplates workspace module]{@link module:/workspace~cacheTemplates}. 
 
 @param {string} ref Cloudfront resource reference.
-@param {boolean} cache The resource fetch request should be cached.
 
 @returns {Promise<String|JSON|Error>} The fetch is resolved into either a string or JSON depending on the url ending.
 */
-async function Cloudfront(ref, cache) {
+async function Cloudfront(ref) {
   if (!xyzEnv.KEY_CLOUDFRONT) {
     return console.error('Cloudfront key is missing');
   }
@@ -58,7 +60,7 @@ async function Cloudfront(ref, cache) {
 
   let response;
 
-  if (cache) {
+  if (Date.now() - cacheTime < 60000) {
     let cachedURL = cacheMap.get(url);
 
     if (!cachedURL) {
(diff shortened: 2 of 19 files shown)


dbauszus-glx (changes requested):
There shouldn't be any breadcrumbing by creating a new control.gazetteer module/method in the library and then changing the ui gazetteer to lead to the new module.

Any ui stuff should be in the ui module. The mapp.ui module does depend on the mapp library. But the mapp library must not have any dependencies on the ui module.

The idea is that the mapp library can be used to build an interface with your own custom elements without loading the ui extension of the mapp library.

We should aim to reduce inline styling not add more as in the listview module.

The css for the control tabs element should be defined in the ui.css sheets as much as possible.

dbauszus-glx:
This is way too complicated at the moment.

The control panel element should not require nested tabs. The tabs should be generated when required.

```
    <div id="ctrls">
      <div id="ctrl-tabs" class="hover"></div>
      <div id="ctrl-panel"></div>
    </div>
```

Decorating the mapview with controlTabs and controlPanels elements, then assigning mapview.locale.controls, then running mapp.ui.elements.controls which doesn't return anything, then defining the layersTab, then do other things before at the end loading controls from the mapp library into the ui.controls element feels too many steps.

dbauszus-glx:
I put this into draft for now and removed the mapp.control logic. The control element should be handled by ui.elements.control.

This is a decorator method which returns a decorated object.

The decorator adds an add method to add controls to the object.

```js
    //Create tabs and panels for controls.
    const control = mapp.ui.elements.control({
      target: document.getElementById('ctrls')
    })

    const layersPanel = mapp.utils.html.node`<div id="layers" class="active">`

    control.add({
      key: 'layers',
      panel: layersPanel
    })
    
    mapp.ui.layers.listview({
      target: layersPanel,
      layers: mapview.layers,
    });

    const locationsPanel = mapp.utils.html.node`<div id="locations">`

    control.add({
      key: 'locations',
      icon: 'location_on',
      panel: locationsPanel
    })
    
    mapp.ui.locations.listview({
      target: locationsPanel,
      mapview,
    });   
```

The ui.gazetteer method needs to be reviewed proper, the params/properties were all over the shop.

There is no need for a mapp.controls.listview method which is dependent on the ui.listview methods.

I added an elements/css module and removed the relevant css for this element from the view document head.

AlexanderGeere:
@dbauszus-glx 

I added the gazetteer by using the `control.add` and made it so the location tab is always there. 
The gazetteer doesnt replace the lcoation tab anymore.

I made a class for the no location selection text i added. 

I added the control element to the mapview in the, I thought we could just call add in the plugins or elsewhere if we wanted to add more controls down the line. 

I will do the documentation if you are happy with the structure.

simon-leech (dismissed):
image

The hover text should be more clear than 'gazetteer' . Maybe something like 'Search for a location' ? 

The rest is working fine for me though! 
Added to a custom view a `controlPanel` and was nice and easy, added a custom control 👍 
Gazetteer still working fine in a custom view too.

RobAndrewHurst:
@simon-leech I added in a gazetteer language entry to say 'Search for a location'

image

PR #2498, opened 2025-11-14
fix(saml): Logout Logic

RobAndrewHurst:
The logout logic in the api.js module was difficult to deal with multiple different types of users.

You wouldn't be able to tell which user is signed in via saml or vanilla until we actually validate the requests auth. 

I then decided to move the logic to the login plugin, where we conditionally set the href of the login/logout button based on if the user is a saml user ie has a `sessionIndex` or if we have a regular mapp user.

There was also an issue with terminating a users SAML session without providing a nameID from the user. I have not ensured that the nameID is passed from the users cookie.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -118,10 +118,6 @@ export default function api(req, res) {
   }
 
   if (req.params.logout) {
-    if (xyzEnv.SAML_SLO) {
-      res.setHeader('location', `${xyzEnv.DIR}/saml/logout`);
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
I am ok with this. I can logout with SAML configured but not used.

There is another security concern which should be addressed in a different ticket/PR.
https://github.com/GEOLYTIX/xyz/issues/2499

PR #2502, opened 2025-11-17
feature(mails): Replace mailer module with Resend 📩

RobAndrewHurst:
# Resend 📩

I have retired the mailer module in favor of the resend module which uses the
resend sdk to send emails.

The logic between the old mailer module and the resend module are very similar
except for the way we handle logging. But it's essentially the same.

## 5 Keys have also been retired

- `TRANSPORT_HOST`
- `TRANSPORT_NAME`
- `TRANSPORT_USERNAME`
- `TRANSPORT_PORT`
- `TRANSPORT_TLS`

None of these keys are needed when authenticating with resend

### DIFF
--- a/mod/utils/resend.js
+++ b/mod/utils/resend.js
@@ -134,6 +134,14 @@ async function batch(emails) {
   if (error) {
     console.error(error);
   }
+
+  let result = `From: ${xyzEnv.TRANSPORT_EMAIL}\nTo: ${emailTemplates.map((template) => template.to)?.join?.(',')}`;
+
+  logger(result, 'mailer');
+
+  result += `\nBody:\n ${emailTemplates?.[0].text?.replace?.('    ', '')}`;
+
+  logger(result, 'mailer_body');
 }
 
 /**


dbauszus-glx (changes requested):
I have updated the resend dependency and the readme.

I tested this and got this working fairly quick though we need to update the wiki in this regard.

Do we need to introduce a new env variable eg. RESEND_KEY? This works fine with the resend API key defined as TRANSPORT_PASSWORD.

The module docs state that:
```
The resend dependency will be imported dynamically on the condition that a TRANSPORT_EMAIL [sender] Or a RESEND_KEY is defined the process environment.
```

Should the statement and related condition check be for both TRANSPORT_EMAIL && TRANSPORT_PASSWORD?

RobAndrewHurst:
> I have updated the resend dependency and the readme.
> 
> I tested this and got this working fairly quick though we need to update the wiki in this regard.
> 
> Do we need to introduce a new env variable eg. RESEND_KEY? This works fine with the resend API key defined as TRANSPORT_PASSWORD.
> 
> The module docs state that:
> 
> ```
> The resend dependency will be imported dynamically on the condition that a TRANSPORT_EMAIL [sender] Or a RESEND_KEY is defined the process environment.
> ```
> 
> Should the statement and related condition check be for both TRANSPORT_EMAIL && TRANSPORT_PASSWORD?

I put it as an optional legacy support for the time being. So new instances don't need to change the name of the key until we phase it out entirely. We should probably deprecate the key then

RobAndrewHurst:
@dbauszus-glx I have removed the RESEND_KEY env variable as we don't need this, and I have added in the condition that we need a TRANSPORT_EMAIL && TRANSPORT_PASSWORD in order to have a successful instance of resend

dbauszus-glx (changes requested):
Very odd. But I get a role access error if there are no roles on a simple OSM layer workspace with public access.

PR #2506, opened 2025-11-18
fix(msg): param injection

dbauszus-glx:
This PR removes the msg param from the logout condition in the API script.


cityremade (approved):
Parameter `msg` still sits as optional in login and register forms where it defaults to ''.
Useless.

PR #2516, opened 2025-11-20
chore(refactor): Layer view method and module documentation

dbauszus-glx:
The method does not need to return the layer object. It's not decorator method but assigns the layer.view property.

Rather than coditionally assigning an empty object, defining the defaults and then spreading the config into the defaults this can be done in one line like so:
```js
  layer.viewConfig = {
    displayToggle: true,
    zoomBtn: true,
    zoomToFilteredExtentBtn: true,
    hideDisabled: false,
    ...layer.viewConfig,
  };
```

todo: the layer display toggle is broken when the layer view is popped out in a dialog.
image

todo: the drawer element should not be expandable if changeend event methods remove all content.
image

The layer view defines the layer drawer as raised but then nests the drawer in a layer-view div of the same name with the same data-id and override the style if the nested container is the direct descendent of the `#layers` specificity. 🤯
image

By default the raised class should not be assigned. Though it is possible to assign this with the classList property.

image

I cleaned up the nested assignment, css nonesense, classlists in this commit. https://github.com/GEOLYTIX/xyz/pull/2516/commits/cdc4c682dda2c7dd1ddac135db8b53a6740fc6e4

layer.drawer property will not over assigned. The layer.view becomes the drawer.

@simon-leech I don't understand the layer.drawer property. If layer.drawer is null content is assigned and the method shortcircuits. If layer.drawer is false the method does not short circuit but the content is also assigned to the layer view at the end of the method.

This is nonesense:
```js
layer.viewConfig.content.forEach((el) => layer.view.append(el));
```

The whole point of the append method is that multiple elements can be appended:
```js
layer.view.append(layer.viewConfig.content);
```
image

Better even. Assign the layer.view and return. The layer.view isn't referenced at all in the viewConfig or the rest of the layerView method until re-assigned with the drawer.

```js
  if (layer.drawer === null) {
    layer.view = mapp.utils.html.node`<div
      data-id=${layer.key}
      class="layer-view">
      ${layer.viewConfig.content}`;
    return;
  }
```

Rather than a condition for layer.viewConfig.zoomToFilteredExtentBtn to assign the layer..zoomToFilteredExtentBtn element property and then more conditions with more conditions to check the style of a conditional element in the filter update panel method.

This can be solved by one method. Shortcircuits without layer.viewConfig.zoomToFilteredExtentBtn adds a changeEnd method to check the display for the btn style.

```js
function zoomToFilteredExtentBtn(layer) {
  if (!layer.viewConfig.zoomToFilteredExtentBtn) return;

  const btnStyle = Object.keys(layer.filter.current).length
    ? ''
    : 'display: none;';

  const zoomToFilteredExtentBtn = mapp.utils.html.node`<button
    data-id=zoomToFilteredExtentBtn
    title=${mapp.dictionary.layer_zoom_to_extent}
    style=${btnStyle}
    class="notranslate material-symbols-outlined"
    onclick=${async (e) => {
      // disable button if no locations were found.
      e.target.disabled = !(await layer.zoomToExtent());
      layer.show();
    }}>filter_alt`;

  function zoomToFilteredExtentBtn_changeEnd(layer) {
    if (Object.keys(layer.filter.current).length) {
      zoomToFilteredExtentBtn.style.display = 'inline-block';
    } else {
      zoomToFilteredExtentBtn.style.display = 'none';
    }
  }
  layer.changeEndCallbacks.push(zoomToFilteredExtentBtn_changeEnd);

  layer.viewConfig.headerBtn.push(zoomToFilteredExtentBtn);
}
```

With the individual method cleaned we can move these to their own module and assign the methods as mapp.ui.layers.viewHeader{}.

We are now able to order the methods and extend this object with custom methods.

```js
  if (!Array.isArray(layer.viewConfig.headerOrder)) {
    // Assign default panelOrder if not a config array.
    layer.viewConfig.headerOrder = [
      'zoomToFilteredExtentBtn',
      'zoomBtn',
      'popoutBtn',
      'displayToggle',
    ].filter((key) => layer.viewConfig[key]);
  }

  for (const method of layer.viewConfig.headerOrder) {
    mapp.ui.layers.viewHeader[method]?.(layer);
  }
```

### DIFF
--- a/lib/ui/elements/drawer.mjs
+++ b/lib/ui/elements/drawer.mjs
@@ -56,7 +56,7 @@ export default function drawer(params) {
       class="notranslate material-symbols-outlined">
       open_in_new`;
 
-      //Identeify the caret element.
+      //Identify the caret element.
       const beforeElement = params.drawer
         .querySelector('.header')
         .querySelector('.caret');
@@ -124,7 +124,6 @@ the popoutDialog creates the popout element for a drawer and appends the popout 
 @property {HTML} params.originalTarget The element that holds the content within the drawer being popped out.
 */
 function popoutDialog(params) {
-  console.log(params);
   if (params.popout?.dialog) {
     //Reappend the content and header
     params.popout.node


dbauszus-glx:
@AlexanderGeere Can you pick up this PR? The drawer popout logic should not have any references to the display-toggle. The logic should work the same on any drawer no matter the type of drawer.

I am happy for the popout button itself to be positioned either first [before title], after title, or last, as long as the logic is not dependent on specific header elements. The only consideration should be for the `.caret` class item.

AlexanderGeere:
@dbauszus-glx I removed any references to the display-toggle. The toggle is also in the proper spot now in the dialog. 
The popoutBtn is added before the caret if it is not already defined in the header of the the drawer.

PR #2518, opened 2025-11-24
feature(interaction): Modify cancel with Escape

RobAndrewHurst:
## Escape key to cancel Modify interaction

I have added in an eventListener to finish the mapview interaction


dbauszus-glx:
Lovely implementation.

PR #2521, opened 2025-11-26
correct fonts

RobAndrewHurst:
# Fonts loading

The local fonts used in the public/css/ directory were not actually 'Titillium Web' and seemed to be sans-serif. After downloading a new set and replacing them it fixes the font issue we are seeing.


cityremade (approved):
looks good to me.

simon-leech (approved):

PR #2522, opened 2025-11-26
feature(button CSS) : Active Button and Active Tab

simon-leech:
# Active Colour CSS

## Description

Clients have flagged that its hard to tell when a button is on or off. 
This PR makes two changes
1. Buttons are now active colour with contrasting text
BEFORE 
image

AFTER
image
2. Tabs now have an active colour underline to indicate which is selected
BEFORE
image

AFTER
image

## Type of Change

- ✅ CSS

## How have you tested this?
Tested locally on a few applications.

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -65,6 +65,8 @@ export default async function decorate(layer) {
     zoomToExtent,
   });
 
+  layer.name ??= layer.key;
+
   // Format methods, eg mvt may already assign a changeEnd callback method.
   layer.changeEndCallbacks ??= [];
 
--- a/lib/layer/featureHover.mjs
+++ b/lib/layer/featureHover.mjs
@@ -10,7 +10,7 @@
 @description
 The featureHover method will send an xhr request for the provided feature argument. The returned string value will be displayed in an infotip element in the feature layer mapview.
 
-@param {object} feature 
+@param {object} feature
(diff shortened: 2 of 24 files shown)


RobAndrewHurst (approved):
LGTM

cityremade:
what about full width buttons with orange background? atrocity.

cityremade:
@simon-leech can you experiment a bit with slightly less aggressive hue?
Here's how to apply opacity.
rgb(from var(--color-active) r g b / 0.5);

cityremade:
I've updated this pr but it's still awkward here with fonts, no differentiation in weight, 
has adding local fonts affected how css is rebuilt ?

Screenshot 2025-12-02 at 14 33 23

cityremade:
so I've added some transparency on the background and made the font dark. 
Changes only in `.flat.active` rule.

simon-leech:
@cityremade Can I check two things? 

1. The addition of the border-radius on the active class means everything shifts slightly down when you toggle the state as the button is larger in active than not. Is this expected / desired? 
2. Are we happy that the button text is blue with orange background when active? It was white before but I'm easy just checking! 
image

cityremade:
ok, I think I should remove border. 
I've set the font to our default font colour - white wasn't readable and the primary colour should not sit on our active colour 🤷‍♀️

simon-leech:
Thanks @cityremade  - all good with me now! 👍

PR #2526, opened 2025-11-27
feat(utils/xhr): flexible throttling

RobAndrewHurst:
This PR enhances the `/utils/xhr` utility by adding **optional request throttling** that automatically cancels any previous in-flight request when a new one is made — ensuring only the **latest** request completes.

This is especially valuable for rapid user interactions like:

- Search-as-you-type
- Live preview
- Form autosave
- Debounced API calls

### Key Features

- `throttle: true` → Global "last-wins" behavior (cancels any previous throttled request)
- `throttle: 'group-name'` → Per-group throttling (multiple independent streams)
- Backward compatible — existing `xhr()` calls behave exactly as before
- Clean abort handling (aborted requests are silently ignored)

### Usage Examples

```js
// Global throttle — only the last request runs
mapp.utils.xhr({ url: "/api/query?", throttle: true });

mapp.utils.xhr({ url: "/api/query?", throttle: true }); // cancels previous

mapp.utils.xhr({ url: "/api/query?", throttle: true }); // cancels all above

// Named groups — independent throttling
mapp.utils.xhr({ url: "/api/query?", throttle: "hover" });
mapp.utils.xhr({ url: "/api/query?", throttle: "hover" }); // only last completes

mapp.utils.xhr({ url: "/api/query?", throttle: "gazeteer" }); // runs separately

// Normal requests — unaffected
mapp.utils.xhr("/api/user");
mapp.utils.xhr("/api/query?");

```

### Changes

- Added smart `throttleKey` resolution:
  - `true` → `'global'`
  - string → custom group
  - omitted/false → no throttling
- Added `cleanup()` helper to avoid duplication

To enable throttling:

```js
// Before
mapp.utils.xhr("/api/query?");

// After (recommended for rapid inputs)
mapp.utils.xhr({ url: "/api/query?", throttle: true });
// or with custom group
mapp.utils.xhr({ url: "/api/query?", throttle: "hover" });
```

### DIFF
--- a/lib/utils/xhr.mjs
+++ b/lib/utils/xhr.mjs
@@ -140,7 +140,7 @@ function setThrottle(throttleKey, xhr) {
     const controller = {
       abort: () => {
         xhr.abort();
-        throttleMap.delete(throttleKey); // Auto-cleanup
+        throttleMap.delete(throttleKey);
       },
     };
     throttleMap.set(throttleKey, controller);


RobAndrewHurst:
https://github.com/user-attachments/assets/090e16de-dc93-412a-921e-600208e04797

Here is an example of what you would see in the network tab

PR #2537, opened 2025-12-08
Filter dialog button text and icon

cityremade:
Minor change for multi filter button.
Changed wording to get rid of the word `panel`
Added filter_alt icon.
Button css now has a `color:inherit` rule on button inner elements to align symbol icon with font if a symbol icon is used.


RobAndrewHurst:
image

I don't think "Apply Filters" is the correct wording for this because it feels like an instant action. Not to bring up a dialog to then apply the filters

jfitzpatrick17:
Is there anything you would suggest instead @RobAndrewHurst?

RobAndrewHurst:
@jfitzpatrick17 "Choose Filters" Or "Filters" or "Filter" might work better?

jfitzpatrick17:
Thanks @RobAndrewHurst. I think "Filters" is my favourite of the 3.

PR #2540, opened 2025-12-09
feature(Mapview): Concat layer.plugins into locale.plugins

dbauszus-glx:
The mapviewPromise method which loads locale.plugins will now iterate over the locale.layers and concat any layer.plugins[] array into the locale.plugins[] array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.20.0**
+**v4.21.0**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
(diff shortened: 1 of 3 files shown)


AlexanderGeere (dismissed):
I made a test worskpace, bugs_testing/plugins/isurf.json, where i put a couple plugins in the layer, measure distance works fine but csv upload throws an error:
screenshot-2025-12-10_11-52-40

Configuration sample:
screenshot-2025-12-10_11-54-34

Have I misunderstood the use case?

AlexanderGeere:
> I made a test worskpace, bugs_testing/plugins/isurf.json, where i put a couple plugins in the layer, measure distance works fine but csv upload throws an error: screenshot-2025-12-10_11-52-40
> 
> Configuration sample: screenshot-2025-12-10_11-54-34
> 
> Have I misunderstood the use case?

This was a bug in the plugin.

PR #2541, opened 2025-12-10
fix(xhr): Replace Throttle logic with Debounce

RobAndrewHurst:
# XHR Debounce Logic

The XHR throttle logic has been replaced with a request debounce.

The way the debounce works is that we create a request that will have a delay
given from the xhr function call. The request will have a timeout of 300ms by
default.

If a request is made before the **timeout has elapsed**, then we clear the
timeout, by the timeout id and create a new one.

Here are some examples of how to use the debounce.

```js
// Default 300ms debounce
xhr({ url: "/query", debounce: true });

// Custom delay
xhr({ url: "/query", debounce: 500 });

// Named group with default delay
xhr({ url: "/query", debounce: "hover" });

// Named group with custom delay
xhr({ url: "query", debounce: "hover:1000" });
```

### DIFF
--- a/lib/utils/xhr.mjs
+++ b/lib/utils/xhr.mjs
@@ -122,7 +122,7 @@ function getDebounceConfig(params) {
   }
 
   if (params.debounce === true) {
-    return { key: 'global', delay: DEFAULT_DELAY };
+    return { key: 'global', delay: params.delay };
   }
 
   return params.debounce;


dbauszus-glx (changes requested):
I'd provide the delay as a seperate param.
```
{
  debounce: 'hover',
  delay: 100
}
```

And then assign the default as `params.delay ??= 300`

dbauszus-glx (approved):

PR #2548, opened 2025-12-16
fix(mvt): Revise wktPropertiesLoad for mvt format

dbauszus-glx:
The wktPropertiesLoad method should be async and not require a nested onLoad method on the layer object.

The xhr request can be debounced for the layer to prevent executing the same query twice on initial loads.

The xhr request can be awaited to process the features by sending an array response to mapp.layer.featureFormats.wkt_properties()

@RobAndrewHurst I don't think the onLoad function param is required for the xhr utility method.

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -336,6 +336,12 @@ async function filter_numeric(layer, filter) {
     await generateMinMax(layer, filter);
   }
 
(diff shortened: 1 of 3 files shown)


simon-leech:
Working all good from me - tested in default view and custom view, just a question on what one line of code is doing

dbauszus-glx:
@simon-leech I added a comment in regards to the featuresObject property

Issue #2550, opened 2025-12-17
Filtering with no min/max


simon-leech:
When no data is present for the `filter.min` and `filter.max` - in scenarios where there is no data, applying this filter would throw an error as no values are available to apply as the `gte` or `lte`.

PR #2551, opened 2025-12-17
fix(Filters): Return before applying Filters when no min or max value

simon-leech:
## Description

This PR addresses an issue whereby when the `min`/ `max` of the filter were both null (for example when no data), the filter was still being applied. 
This would result in a error as there was no values to apply to the filter. 

This PR moves the return to before the `applyFilter` method - ensuring that its not hit when no data.

## GitHub Issue

[Provide the link to the relevant GitHub issue it addresses.](https://github.com/GEOLYTIX/xyz/issues/2550)

## Type of Change

## How have you tested this?

Tested locally when the min/max query returns nothing. 

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/mod/view.js
+++ b/mod/view.js
@@ -22,6 +22,8 @@ import logger from './utils/logger.js';
 @description
 The View API method will request a view [template] from the languageTemplates module method.
 
+Only string values with whitelisted characters can be substituted in view templates.
+
 The optional params.msg string property may have a languageTemplate which should be assigned to the params string before the substitution of template variables.
 
 The view [template] is a HTML string. Template variables defined within a set of brackets `{{var}}` will be substituted with params property values before the view string is sent from the HTTP Response object.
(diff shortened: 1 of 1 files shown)


simon-leech:
> Working well.
> 
> I set the integer_field to null in the table.
> 
> image
> But to confirm; Every item will be filtered out. Is this the intended behaviour?
> 
> I now update one location.
> 
> I still won't see the slider but one location passes the filter.
> 
> image
> Only once there is a second location with a different value, eg 100 and 200 do I get the slider element.
> 
> image

I'm happy with this logic yeah! 

No data = no matching locations which is highlighted to the user. 
1 matching location and range true = no slider as only one value but 1 matching location which is highlighted to the user. 
More than 1 matching location = slider and filter working as expected.

PR #2566, opened 2025-12-23
feature(Jenks/Legend): Format cat.label as localeString

dbauszus-glx:
The cat.label should be formatted as localeString in the jenks featureFields method.

image

### DIFF
--- a/lib/utils/numericFormatter.mjs
+++ b/lib/utils/numericFormatter.mjs
@@ -60,7 +60,7 @@ export function formatNumericValue(params) {
 
     // "integer" type values must not have fraction digits.
     params.formatterParams.options.maximumFractionDigits ??=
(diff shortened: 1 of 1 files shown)


cityremade:
how would the decimals be specified?

dbauszus-glx:
@simon-leech I put this back into draft and will update the related issue.

dbauszus-glx:
This supports now the formatNumericValue with theme.legend properties provided.

image

simon-leech:
I have fixed the issue with `params.round` 

This statement was missing brackets around the OR part of the ternary. 

``` js
// "integer" type values must not have fraction digits.
    params.formatterParams.options.maximumFractionDigits ??=
      params.round || (params.type === 'integer' ? 0 : 2);

PR #2568, opened 2025-12-24
svgTemplates plugin from utils

### LINKED ISSUE #2567: icons interface / svgTemplates plugin
The svg_templates plugin which which maps utils.svgTemplates should be renamed to svgTemplates and doesn't have to be flagged as deprecated.

The shouldn't be any reason to run this method seperate from other plugins from the mapview module.

An icons interface should allow to add or overwrite existing svgTemplates which must be stored as string.

It must be possible to store a user locale with custom svgTemplates.

dbauszus-glx:
Instead of having deprecated plugin hanging around which calls the utility method it will be easier to remove the call for the util from the mapview decorator and map the svgTemplates util method to plugins.svg_templates.

This way the async plugin method is called along with the other plugins.

The utility method can be called to load additional templates stored in mapp.utils.svgSymbols.templates{}.


AlexanderGeere (approved):
Templates load as expected.

RobAndrewHurst (approved):

PR #2576, opened 2026-01-06
feature(featureFormat): Add wkth3 featureFormat

dbauszus-glx:
This PR orders the methods in the featureFormats module and adds the async wkth3 featureFormats method which requires the h3-js@4.4.0 library through esmImport.

### DIFF
--- a/lib/layer/featureFormats.mjs
+++ b/lib/layer/featureFormats.mjs
@@ -205,7 +205,15 @@ Otherwise the coordinates will be parsed from the cell id with h3js.cellToBounda
 
 An Openlayers polygon geometry will be created from the parsed coordinates. The geometry object will be stored in the cellGeomMap.
 
-The wkt layer format is recommended for the "wkth3" featureFormat. The geometry field is required for the request viewport parameter. The "no_geom" flag should be set to not return the geometry with the wkt features.
+The wkt layer format is recommended for the "wkth3" featureFormat. 
+
+The geometry field is required for the request viewport parameter. This is used to limit the data displayed to the current mapview extent.
+
+The "no_geom" flag should be set to not return the geometry with the wkt features.This is as it is not required for rendering the h3 cells.
+
+The "qID" flag should return the h3 cell id from the database table for lookup to find the h3 geometry. 
+
+The assignment of properties to the cellGeomMap requires each h3 cell to be unique, thus the qID should return a unique h3 cell id for each feature.
 
 ```js
 "format": "wkt",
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
I'm a little confused on how this works - can we update the documentation? 

The issue I'm facing, I have a table with duplicated `gridcd` (that is the hex code for the geometry). I have a unique id `qID: id` but I'm unable to use that currently as it has to use `gridcd` for the lookup? 

My suggestion would be that we have a h3_id field instead that can be set to the `gridcd` field while retaining the `qID` as a unique `id`? 

Second question - what does the `geom` do in this scenario as its being returned from this feature, but without it I get errors.

lib/layer/featureFormats.mjs
```diff
@@ -6,6 +6,51 @@ The featureFormats module exports utility methods to process the response data f
...
+@property {integer} cluster.max_size The length of features in the largest cluster feature.
+@property {Object} layer.params
+@property {Array} layer.params.fields Array of field names for additional feature properties.
+@returns {Array} Array of OpenLayers Feature objects.
+*/
+export function cluster(layer, features) {
+  layer.cluster.max_size = 1;
+
+  return features.map((vals, i) => {
+    const geometry = new ol.geom.Point(vals.shift());
```
RobAndrewHurst: What is the reasoning of using shift here?

Because this will mutate the array by removing the element from the array.

Is this what we want?

Image

dbauszus-glx: Could probably be done without a shift. I only moved the cluster function so that the order is alphabetical. Do you want to open a ticket to investigate the cluster function execution? i haven't tested this in a long time.

dbauszus-glx:
@simon-leech We don't want to go down this road. Duplicate geometries in a h3 table is bad data schema design. This is not something we want to encourage and support by making the config more complex, dealing with required filters, making the selection awkward, and add unnecessary parsing in the featureFormats method. MVT or vector should be used in this case which woorks. The wkth3 featureFormat is designed as opinioated format to be most performant and making best use of resources [bandwidth and storage], duplicate geometries is not.

simon-leech:
@dbauszus-glx Yeah completely agree no problems! 

If you could update the docs for me to explain the need for the qID to be a valid hex code and then also the usage of the geom property would be grand and I can re-test after that 👍

dbauszus-glx:
@simon-leech I updated the docs and provided a config example.

simon-leech (approved):
Happy with this now!

Issue #2508, opened 2025-11-18
Consent plugin [Cookie/Data]


dbauszus-glx:
There should be a consent logic in the mapp.ui library to use the toast element.
The plugin must NOT use localstorage but update the user property.

dbauszus-glx:
The consent message must be flexible in regards to data access. It must be optional to brick / log out the user if consent is not given. The consent must be stored in the user object. The consent must be revoked if the terms change. Consent must be re-requested if not stored in the user object.

simon-leech:
I think the one bit we need to add to the core is the ability to pass a mask flag that will mask the rest of the application while the toast is open, so nothing can be accessed while this is open. 

The rest of these requirements can be handled on a case by case basis in a plugin that uses a toast, no need to build this into the core

dbauszus-glx:
The linked PR uses the confirm dialog and stores the mapp.user.consent flag in the indexdb 'user' store.

@simon-leech The mask flag is not required with the confirm dialog.

PR #2577, opened 2026-01-07
feature(Toast): Add a mask to the toast element

simon-leech:
## Description
This PR introduces a `mask` flag to the `toast` element. 
When set, this will add a mask across the rest of the body that prevents interactions until the `toast` element is removed. 
The `mask` class has been added to the `common.css` file as it could be used for other things. 

## GitHub Issue

[Provide the link to the relevant GitHub issue it addresses.](https://github.com/GEOLYTIX/xyz/issues/2508)

## Type of Change

## How have you tested this?
Tested this locally by including the `mask` flag.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist


RobAndrewHurst (changes requested):
We need to give this a different name. Because this isn't a toast. This is more a banner.

image

A toast is something that appears as a brief notification, which can appear at the bottom of the page or in either one of the top corners.

simon-leech:
@RobAndrewHurst understood! 
Is it ok if we do that in a separate PR though as this one was specifically just for a mask and would like to keep them tidy? 

Also, I wouldn't want to break any calls to toast so do you think we can have a separate ticket to address that where we rename and deprecate calls to toast with a warning?

RobAndrewHurst:
To be fair I can't remember this going in. And I remember having a discussion about this with @dbauszus-glx & @jfitzpatrick17 that this isn't a toast. And in previous discussions when it was just a plugin :/ But oh well we will sort it.

dbauszus-glx:
A toast should not have a mask flag. A confirm dialog element should be used for user consent.

PR #2578, opened 2026-01-07
Drawer vs layer css review

cityremade:
This PR has the following changes:

* ui/elements/drawer - now supports override for default classList. It makes it possible to use drawer component without `.drawer` class, non-breaking.
* ui/layers/view - now uses drawer without `.drawer` class when creating layer in the layers panel
* ui/layers/listview - creates a layer group without `.drawer` class
* new stylesheet `_layer` added to store rules for a drawer with `.layer` class
* `_layerview` stylesheet cleaned up to store only distinction between a standalone layer and a layer group
* new background colour added to both light and dark palettes, new colour to cut between secondary and tertiary bases
* `_drawer` and `_layerview` stylesheets reviewed and cleaned up for separation
* rules on `.header` and `.meta` applied with existing classes (eg .no-select) and redundant rules removed
* most common border radius of 3px now stored as root variable so that all relevant containers can be adjusted at once.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -26,7 +26,8 @@
       "!package.json",
       "!jsdoc_mapp.json",
       "!jsdoc_xyz.json",
-      "!jsdoc_test.json"
+      "!jsdoc_test.json",
+      "!**/.vscode"
     ]
   },
   "formatter": {
--- a/lib/location/get.mjs
+++ b/lib/location/get.mjs
@@ -72,7 +72,11 @@ export async function get(location, list = location.layer.mapview.locations) {
   // Assign locale key to location
   location.locale ??= mapview.locale.key;
 
-  if (!(await getInfoj(location))) return;
+  if (!(await getInfoj(location))) {
+    locations.delete(location.hook);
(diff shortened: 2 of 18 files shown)


cityremade:
Warning: expected products analytics errors as some selectors rely on closest `.drawer`.

lib/ui/layers/view.mjs
```diff
@@ -142,11 +142,9 @@ function viewConfig(layer) {
     });
   }
 
-  // layer.classList is the legacy configuration for layer.viewConfig.classList
```
simon-leech: I think we need to keep this nullish assignment to prevent any issues. 
As if you provide `layer.viewConfig.classList` currently - it has `layer-view` added to it. 
But now  it will be ignored it seems?

dbauszus-glx:
Screenshot for reference. It does look busy.
image

cityremade:
this is not quartenary, this is a colour that cuts through secondary and tertiary and should not be used freely.
I've kept these separate for now as both may have some further work on it.

PR #2580, opened 2026-01-08
Layer order core plugin;

cityremade:
This PR adds layer order plugin to core plugins.

I have looked into adding it as a `ui/layers` method but after consideration it wouldn't sit well as it needs access to mapview and locale which are not typical arguments in this area. Also that way it would have to be called before listview is created and this is not actual UI behaviour since it is merely manipulating underlying objects.

So I decided to add as a plugin since it takes two arguments configuration and mapview objects which is an actual plugin logic.

### DIFF
--- a/lib/ui/utils/dataview.mjs
+++ b/lib/ui/utils/dataview.mjs
@@ -16,6 +16,7 @@ const dataview = {
   Toolbar,
   toolbar: {
     update,
+    viewport,
   },
 };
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx (changes requested):
This issue should be addressed in either the mapview decorator or in the addLayer method.

We don't do plugins for bug fixes.

cityremade:
This is a feature to alter existing order of reusable configuration.

The bug would be order not guaranteed after all templates in the workspace have been processed.

simon-leech (approved):

PR #2587, opened 2026-01-13
fix(SAML): Redirect saml acs

dbauszus-glx:
Get redirect location from _redirect cookie.

### DIFF
--- a/mod/utils/redirect.js
+++ b/mod/utils/redirect.js
@@ -1,6 +1,15 @@
 /**
+## /utils/redirect
+
+The redirect utility module exports a function to set a cookie with a redirect location in the response header.
+
+@module /utils/redirect
+*/
+
+/**
 @function setRedirect
-@description Validates, sanitizes, and stores a redirect URL in a secure cookie.
+@description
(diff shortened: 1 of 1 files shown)


AlexanderGeere:
@RobAndrewHurst I like the addition of the setRedirect function, definitely a useful addition for further endpoints. Do we still have to do the redirect after the SAML Login, does it not redirect to the supplied RelayState? 

Otherwise its looking good to me, with the exception of the issue.

RobAndrewHurst:
> @RobAndrewHurst I like the addition of the setRedirect function, definitely a useful addition for further endpoints. Do we still have to do the redirect after the SAML Login, does it not redirect to the supplied RelayState? 
> 
> 
> 
> Otherwise its looking good to me, with the exception of the issue. 

No the idp won't control that.

simon-leech (approved):

AlexanderGeere (commented):
When you dont have access to requested layer, nothing loads in mapp: 
screenshot-2026-01-15_16-10-18

Not sure if this should maybe be addressed in a different issue though

PR #2589, opened 2026-01-14
tests(getLayer / getLocale): Additional tests for getLayer and getLocale; Reorder Operations in getLayer

simon-leech:
## Description

This PR adds additional tests for `getLayer` and `getLocale`. 
3 tests have been added to each test suite. 

obj with role and template[s] with role
obj with role and template[s] without role
obj without role and templates[s] with role

The order of operations in the `getLayer` has been updated to match `getLocale` - this is as follows: 
``` js
// The roles property maybe assigned from a template. Templates must be merged prior to the role check.
  layer = await mergeTemplates(layer, params.user?.roles);

  //If the user is an admin we don't need to check roles
  if (!Roles.check(layer, params.user?.roles)) {
    return new Error('Role access denied.');
  }

  layer = Roles.objMerge(layer, params.user?.roles);
```

## Type of Change
- ✅ Testing

## How have you tested this?
Running the tests.

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added

### DIFF
--- a/mod/workspace/getLayer.js
+++ b/mod/workspace/getLayer.js
@@ -94,12 +94,10 @@ export default async function getLayer(params, locale) {
   // The roles property maybe assigned from a template. Templates must be merged prior to the role check.
   layer = await mergeTemplates(layer, params.user?.roles);
 
-  //If the user is an admin we don't need to check roles
-  if (!Roles.check(layer, params.user?.roles)) {
-    return new Error('Role access denied.');
+  // The mergeTemplates method returned an Error.
+  if (layer instanceof Error) {
+    return layer;
   }
-
-  layer = Roles.objMerge(layer, params.user?.roles);
 
   // Assign layer key as name with no existing name on layer object.
   layer.name ??= layer.key;
--- a/mod/workspace/getLocale.js
+++ b/mod/workspace/getLocale.js
@@ -95,13 +95,6 @@ export default async function getLocale(params, parentLocale) {
     return locale;
   }
 
-  //If the user is an admin we don't need to check roles
-  if (!Roles.check(locale, params.user?.roles)) {
-    return new Error('Role access denied.');
-  }
-
-  locale = Roles.objMerge(locale, params.user?.roles);
-
   locale.workspace = workspace.key;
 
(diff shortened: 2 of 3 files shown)


simon-leech:
> The getLocale method checks whether mergeTemplates returns an Error. When is this the case? Should the same check be performed in the getLayer method?
> 
> ```js
>   // The roles property maybe assigned from a template. Templates must be merged prior to the role check.
>   locale = await mergeTemplates(locale, params.user?.roles);
> 
>   // The mergeTemplates method returned an Error.
>   if (locale instanceof Error) {
>     return locale;
>   }
> 
>   //If the user is an admin we don't need to check roles
>   if (!Roles.check(locale, params.user?.roles)) {
>     return new Error('Role access denied.');
>   }
> 
>   locale = Roles.objMerge(locale, params.user?.roles);
> ```
> 
> The mergeTemplates method is only called from getLocale and getLayer, in each instance the call is followed by a check for Roles and the Roles.objMerge execution. Instead of having this code in the getLayer and getLocale modules the Role check and objMerge should just be added to the mergeTemplates method.


The `getLocale` will return an error if the user doesn't have permissions for that locale. There is actually a test for this that fails if the error check is removed. 

I've moved the duplicate code into the mergeTemplates method and all tests are passing - but we need to do some testing on an application level.

simon-leech:
> Isn't `obj = Roles.objMerge(obj, roles);` required before at the end of the mergeTemplates method?

If i include that then one of the tests fails as we get duplicated `infoj` entries. 
I'm guessing its because it tries to merge multiple times. 

I wonder if that line is actually doing anything in the `getLayer` and `getLocale` currently as the `mergeTemplates` has a `Roles.merge` in it anyway?

dbauszus-glx:
@simon-leech Good call. Let's take it out and see whether the test pass.

simon-leech:
I removed it in my last commit and all the tests are still passing so I think all good! 

Just need to do a bit of application testing to ensure no unexpected changes occurred.

dbauszus-glx (approved):

PR #2591, opened 2026-01-14
fix(layer/groupmeta): layer.groupmeta should be overwritten in group.meta  element.

dbauszus-glx:
Instead of adding groupmeta from each layer the group.meta element should be re-rendered.


cityremade:
by the looks of it it settles at the last one that was added to the group.
Is this desired behaviour?

dbauszus-glx:
@cityremade Yes. That is expected behaviour.

cityremade (approved):

PR #2593, opened 2026-01-14
feat(ui.elements.draw): Provide dialog property for draw panel.

### LINKED ISSUE #2592: Drawing panel interface
The drawing panel could have addtional nested drawers [eg. circle configuration].

Drawer nested in drawer are not easily visualised.

It is possible to popout the drawing panel drawer into a dialog.

It should also be possible to have a button instead of the drawer which opens the panel in a dialog [like filter panel].

This should also work within the location view.

dbauszus-glx:
It should be possible to display the draw panel in a dialog.

This draft PR copies script from the filter panel.

The showDialog logic wants to be cleaned up and there should not be duplicate code in the filter and draw panel modules.

This must be tested in a location view and on mobile interfaces.

This can be tested on `bugs_testing/draw/workspace.json`

Location Panel: 
screenshot-2026-02-04_12-30-20

Layer Panel:
screenshot-2026-02-04_12-30-30


AlexanderGeere:
@dbauszus-glx 
I moved the `drawDialog` and `openDialog` to a ui/utils module called panelDialog. 
Filter and draw panels now call the `panelDialog` function to create the dialog button for the panel.

AlexanderGeere:
@dbauszus-glx 

I had to add a little extra, but the panel can now be called into a dialog from a location as well.

RobAndrewHurst (approved):
Happy for this to go in

dbauszus-glx:
We want to merge https://github.com/GEOLYTIX/xyz/pull/2629 first.

simon-leech:
I resolved the merge conflict after the other PR was merged - testing this out now 👍

simon-leech (dismissed):
**1. Previously Filter displayed as just a button.** 
Image

I now see the label `Filter` for no reason as the button has the label. 
Image


**2. Also the `panelOrder` seems to have been changed without any configuration changes? This shouldn't happen.** 

**3. Duplicated options in the dialog** 
 ``` json 
    {
      "label": "HELLO",
      "display": true,
      "field": "field_a",
      "fieldfx": "ST_asGeoJson(field_a)",
      "edit": true, 
      "draw": {
        "polygon": true, 
        "circle": true,
        "dialog": true
      },
      "style": {
        "strokeColor": "#BB0A21",
        "fillColor": "#BB0A21",
        "fillOpacity": 0.3
      },
```

Image

**4. Duplicated drawer arrow in edit options** 
_Note this is using a plugin so may need to be fixed there (isoline_here)_ 
Image

**Comment by AlexanderGeere** [comment]
@simon-leech 
1. The drawer was previously create by an `if...else` i have added an early return so the label should now be gone.
2. I added the expected `data-id`s to the buttons so the panelOrder shold be correct now.
3. Added an edit check on the entry method as this was causing the duplication.
4. The caret is now conditionally added if its not there already

**Review by simon-leech — APPROVED** [review]
I'm happy from my testing that this is working as expected 💯

**Comment by dbauszus-glx** [comment]
This works pretty well. I don't understand the caret lookup. @AlexanderGeere Can you add a comment and let us know how to test this?

```js
  const caret = params.header?.template?.find?.((el) => el.includes('caret'))
    ? ''
    : mapp.utils
        .html`<div class="notranslate material-symbols-outlined caret"></div>`;
```

I changed the panelDialog to use a params object with the layer a property of the params.

I am not entirely sure how the setView property works. How can I test this.
```js
  //Setup a dialog.view where the panel needs a reference to a specific element.
  if (params.setView) {
    panelParams.dialog.view = panelParams.dialog.node.querySelector(
      `#${params.setView}`,
    );
  }
```

AlexanderGeere:
@dbauszus-glx 
> ```js
>   const caret = params.header?.template?.find?.((el) => el.includes('caret'))
>     ? ''
>     : mapp.utils
>         .html`<div class="notranslate material-symbols-outlined caret"></div>`;
> ```
> 
The `isoline_here` plugin creates a dropdown which adds its own caret to the header, without this you would see two carets, 
I will add a comment saying as much. 

> ```js
>   //Setup a dialog.view where the panel needs a reference to a specific element.
>   if (params.setView) {
>     panelParams.dialog.view = panelParams.dialog.node.querySelector(
>       `#${params.setView}`,
>     );
>   }
> ```

This was necessary for the filter panel, as the 'cards' are appended to an html element and we need to know the view element to do so
screenshot-2026-02-10_14-39-05

cityremade:
right now drawer is only created as a button with a dialog in `layer.view` containers.

I can't show circle settings alone in the dialog or just isoline settings
I thought we would have drawer module return a button,
that way any drawer could optionally be a dialog

An example for layer.draw object:

```json
"draw": { 
     "point": true, // this is just a button 
     "circle": { 
           "drawer": false, // this works, gets rid of the expandable container for circle drawing
           "dialog": true // this doesn't do anything
      }
}
```


image
All drawing tools in one go

dbauszus-glx:
@cityremade is the sample configuration missing the dialog:true flag in the draw object? Without this I see the draw panel as a drawer. The screenshot seems to look like:

```
"draw": {
  "dialog": true,
  "point": {},
  "circle": {
    "drawer": false
  }
},
```

What is the expected result for circle.dialog:true should there be a dialog button inside a dialog?

cityremade:
I would like to be able to make any drawer a button with dialog, not just those elements of the view that support it. 

Use case: make isoline settings or circle creation into a button+dialog instead of showing it in the layer view at all times.

This would go a level deeper than layer.draw.drawer{} , eg. layer.draw.circle.drawer{} or layer.draw.isoline_here.drawer{},
similar to the work done on drawer flags.

This could go in next iteration as well.

dbauszus-glx:
@cityremade Valid points. The method to create a dialog instead of a drawer should be universal and not limited to layer panel.

AlexanderGeere:
@cityremade / @dbauszus-glx 
I have renamed the module to `drawerDialog` and moved the call to the drawer ui element module. 

I have also added a `dialog: layer.<element>.dialog` to the drawers in the framework, i did this to all the ones that have popout. 

So you should be able to do `dialog: true` on any of the framework elements that are drawers e.g. dataviews, style, etc. 

This can be seen on `bugs_testing/draw/workspace.json`

dbauszus-glx:
@AlexanderGeere 

I don't think the mapp.ui.utils.drawerDialog(params) utility method is needed.

I added the panelDialog and openDialog method to ui.elements.drawer to call the ui.elements.dialog and return the toggle button.

I just commented out the mapp.ui.utils.panelDialog in the layer.panels.draw module and added a config block for layer.draw.dialog to allow for configuration in the dialog object is provided as an object.

```js
  if (layer.draw.dialog) {
    layer.draw.dialog = {
      ...layer.draw.dialog,
      btn_label: 'Open Drawing Dialog',
      title: 'Drawing',
      icon: 'new_label',
    };
  }

  layer.draw.data_id ??= `${layer.key}-draw-drawer`;

  // mapp.ui.utils.panelDialog({
  //   layer,
  //   key: 'draw',
  //   name: 'Draw',
  //   icon: layer.draw.dialog?.btnIcon || 'new_label',
  //   data_id: 'draw-drawer',
  // });
```

dbauszus-glx:
@AlexanderGeere I commented the setView property as I wasn't sure how to test this.

I also noticed that the object assign was the wrong way aroound. params.dialog must be spread into the defaults. Otherwise the custom config won't be applied.

https://github.com/GEOLYTIX/xyz/pull/2593/commits/c81c4950e2b662defb71c7a1dde30b499a0d3a70

simon-leech (dismissed):
The location view caret has been moved accidentally. 

The close button should be on the end of the location view header. 
This was caused by this PR - https://github.com/GEOLYTIX/xyz/issues/2629

AlexanderGeere:
@dbauszus-glx 
I exported the function to be used in the geometry entry.

I removed the ui util as the function now lives in drawer. 

@simon-leech 
I have returned the caret to its previous position

dbauszus-glx:
@AlexanderGeere It would be much easier if the dialog properties are defined in the dialog object.

The drawDialog object should not be required. The `key` and `name` properties don't appear to be used.

```js
  if (layer.draw.dialog) {
    layer.draw.dialog = {
      ...layer.draw.dialog,
      btn_label: 'Open Drawing Dialog',
      title: 'Drawing',
      icon: 'new_label',
    };
  }

  layer.draw.data_id ??= `${layer.key}-draw-drawer`;

  const drawDialog = {
    key: 'draw',
    name: 'Draw',
    icon: layer.draw.dialog?.btnIcon || 'new_label',
  };
```

image

This allows me to provide custom properties in the workspace.

This is not possible if you hardcode the properties into drawer params.

```
        "draw": {
          "dialog": {
            "btn_label": "Add new Location",
            "icon": "thumb_up"
          },
          "point": {
          },
          "circle": {
            "drawer": false,
            "dialog": true
          }
        },
```

This was working in the initial drawDialog method.

```js
export function drawerDialog(params) {
  if (params.dialog === true) {
    params.dialog = {};
  }

  params.dialog.btn_label ??= `Open dialog`;
  params.dialog.btn_title ??= params.dialog.btn_label;

  params.dialog.btn = mapp.utils.html.node`<button
    class="wide flat action multi_hover"
    data-id=${params.data_id}
    title=${params.dialog.btn_title}
    onclick=${() => {
      // classList.toggle resolves as true when the class is added.
      if (params.dialog?.btn.classList.toggle('active')) {
        openDialog(params);
      } else {
        // The decorated dialog object has a close method.
        params.dialog.close();
      }
    }}>
    <span class="material-symbols-outlined notranslate">${params.dialog.icon}</span>
    ${params.dialog.btn_label}`;

  // Hide the dialog when the layer is hidden
  params.layer?.hideCallbacks?.push(() => {
    params.dialog.close?.()
  });

  //Show the dialog on layer display
  if (params.dialog.showOnLayerDisplay) {
    params.layer?.showCallbacks.push(() => {
      params.dialog.show?.()
    });
  }

  return params.dialog.btn;
}
```

image

Although I had the spread wrong. This must be spread into the default, not the other way.
```js
  if (layer.draw.dialog) {
    layer.draw.dialog = {
      btn_label: 'Open Drawing Dialog',
      title: 'Drawing',
      icon: 'new_label',
      ...layer.draw.dialog,
    };
  }
```

The use of the panelParams is awkward. I would just provide the params and not trying to lookup the params from the optional layer.

dbauszus-glx:
@AlexanderGeere can you check this commit that uses a dialog object instead of spreading the a drawerPanels object into the drawer params for the dialog creation.

https://github.com/GEOLYTIX/xyz/pull/2593/commits/1a0e099f909652cde36440d36b703244036cfff8

This should allow you to override any config in the workspace.

I used icon but happy to change to btnIcon if this is already used as a dialog property.

Creating an element with an id, then setting the id as setView in order to look up the id for the dialog property feels odd. Is it not possible to achieve the same by assigning the element as the view property in the dialog object?

dbauszus-glx:
Custom config must always be spread into the defaults object. Not the defaults assigned to the custom object. I fixed this for params.popout{} and params.dialog{}.

The params.data_id is used for the button and assigned with the dialog suffix for the dialog element itself.

`dialog.icon` is the default which will be assigned if no custom has been provided as either `dialog.icon_name` or `dialog.btnIcon`.

cityremade:
How is helper dialog positioning determined? Should it be moved to the right hand side  to prevent overlap with geometry settings dialogs?
Screenshot 2026-02-13 at 13 41 04

dbauszus-glx:
@cityremade We shouldn't worry about this now. Because then we find something else on the right hand side and keep moving items around.

Let's create a seperate ticket for dialog positioning where we can look at different scenarios rather than making this PR another magnitude more complex and touch even more modules.

cityremade (approved):
I'm happy with this, thanks.

simon-leech (approved):
This is brilliant!!!! 👍 
Really liking this! 

I pushed a tiny commit to change the default titles for the dialogs from a little bit of UAT

RobAndrewHurst (approved):

PR #2602, opened 2026-01-16
fix(dropdown/CSS): Enable selected option highlight for single select dropdown

dbauszus-glx:
The selectOnChange method did only toggle the selected class on the option element for params.multi dropdown elements.

This PR adds the selected class to the `params.options[selectedIndex]` element after removing the class from any options element.

This works since the selectedIndex options element will always be on. It is not possible to undo select by clicking on the selected element.

https://github.com/user-attachments/assets/4730a9e1-585f-42f6-9158-9feba9257b38

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -120,8 +120,13 @@ function selectOnChange(e, params) {
   }
 
   // Single select options can not be unselected.
-  params.options.forEach((option) => option.classList.remove('selected'));
+  params.options.forEach((option) => {
+    option.classList.remove('selected');
+    option.style.backgroundColor = 'var(--color-base-secondary)';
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
@simon-leech The yellow colour is the changed colour for the entry which applies to the whole of the input. Like editing the text in an textarea element, or adding an array item.

This is not to be confused with the selected element colour which is controlled in the dropdown element.

AlexanderGeere:
@dbauszus-glx we can change it to remove the yellow from the dropdown with two extra lines:
screenshot-2026-01-19_11-07-39

And then the dropdown look likes this:
screenshot-2026-01-19_11-09-06

Happy to leave it as is but just for simons concerns.

simon-leech (approved):

PR #2603, opened 2026-01-16
Nested locales w/ Nested Roles

RobAndrewHurst:
### Summary

This massive PR introduces support for nested locales w/ nested roles in the workspace, with a complete overhaul of how roles are combined across template and locale hierarchies. The core problem was that roles from sibling templates/locales were leaking into each other's nested children, producing incorrect compound roles (e.g., `demographics.brand_a` when `brand_a` should only exist under `stores`).

### What does the new world look like?

**Role Combination (`mod/utils/roles.js`)**
- New `combine()` export that dispatches to either `combineTemplateRoles` (for template contexts with `localeRole`/`templateRole`/`objRole`) or `combineLocaleRoles` (for nested locale parent-child relationships).
- `combineLocaleRoles` creates `parent.child` dot-notation role combinations only for child-specific roles not already present in the parent.
- `combineTemplateRoles` replaces the old `roleAssign` function from `mergeTemplates.js`, moved here for better separation of concerns.
- `fromObj` now also extracts single `role` string properties (not just `roles` objects).
- `check` handles empty `roles` objects gracefully.

**Template Merging (`mod/workspace/mergeTemplates.js`)**
- Replaced monolithic `objTemplate` with focused functions: `prepareTemplate`, `mergeObjectWithTemplate`, `processRecursiveTemplates`, and `getRoleContext`.
- Role context is now captured before iterating `templates[]` arrays, preventing sibling templates from polluting each other's role scope.
- When a template contributes nested sub-templates, the sub-templates are combined only with the parent template's roles (`nextTemplatesContext`), not the fully accumulated object.
- When a template contributes a `locales` array, its role context is stored as `localesRoleContext` on the merged object so `getLocale` can scope nested locale role combination appropriately.
- Role check moved into `mergeTemplates` so access control is enforced during template composition.

**Locale Resolution (`mod/workspace/getLocale.js`)**
- Refactored into `loadLocale`, `processRoles`, and `composeLocale` for clarity.
- `processRoles` uses `parentLocale.localesRoleContext` when available, ensuring nested locales are combined only with the template's roles that defined the `locales` array, not accumulated sibling template roles.
- `composeLocale` composes dot-notation role strings (`parentLocale.role.locale.role`) and merges parent into child using `structuredClone` to prevent mutation.
- Temporary properties (`localesRoleContext`, `src`, `template`, `templates`, `_type`) are cleaned up before returning.

**Workspace Caching (`mod/workspace/_workspace.js`)**
- New recursive `loadLocale` function that processes nested locales via comma-separated keys (e.g., `uk,stores,brand_a`).
- `cacheTemplates` now operates on a `structuredClone` of `cache.locales` to prevent cache mutation, and can return the locales clone via `params.locales`.
- `roles()` endpoint extracts roles from all cached locales individually and removes self-referential duplicate roles (e.g., `locale.locale`).
- `getNestedLocales` now uses `getLocale` with a parent locale clone for proper role combination, replacing the previous `getTemplate` approach.
- Root `locales()` endpoint skips comma-separated keys (nested locales) from the top-level listing.

**Locale Dropdowns (`public/views/_default.html`)**
- Replaced flat single-dropdown locale selector with recursive `localeDropdowns()` that creates cascading dependent dropdowns for each nesting level.
- Dropdowns are only rendered when there is more than one option at a level — single-entry levels are auto-selected.
- `resolveLocalePath()` walks through all nesting levels to build the full locale path upfront, then issues a single redirect if the URL doesn't include auto-selected single-entry levels. This ensures the correct merged locale is loaded without multiple redirects.

### What tests have been written?

- New workspace test: **"Sibling Templates with Nested Locales"** — verifies `brand_a`/`brand_b` roles only appear under `stores` and `uk.stores`, not under sibling templates like `demographics`.
- New mergeTemplates test: **"sibling templates should not leak roles into nested sub-templates"** — unit-level verification of scoped role context.
- Extended workspace nested roles tests covering: role-restricted locale access, anonymous access denial, hidden parent locales, inherited role layer access, and cross-locale isolation.
- New test fixtures: `stores_template.json`, `demographics_template.json`, `sibling_workspace.json`, and the full `nested_roles/` fixture set.

### DIFF
--- a/mod/workspace/getLocale.js
+++ b/mod/workspace/getLocale.js
@@ -196,7 +196,9 @@ async function composeLocale(locale, parentLocale, params, workspaceKey) {
       if (parentLocale.localesRoleContext?.roles) {
         const contextRoles = Object.keys(parentLocale.localesRoleContext.roles);
         // Use the longest role as it represents the most specific path.
-        const parentRole = contextRoles.sort((a, b) => b.length - a.length)[0];
+        const parentRole = contextRoles.toSorted(
+          (a, b) => b.length - a.length,
+        )[0];
         locale.role = `${parentRole}.${locale.role}`;
       } else {
         locale.role = `${parentLocale.role}.${locale.role}`;


simon-leech:
I've tested against non-nested roles and no impact on them, all working as expected which was expected based on the changes but wanted to double check! 

Will move onto nested next 🤞

dbauszus-glx:
I added the `germany` locale key to it's own locales array for the nested locales test and added a console.error to the if condition.

The newVisited const is not needed when adding the key to the set in the recurse iteration of obj.locales.

dbauszus-glx:
qualifiedRoles and parentRoles should be sets throughout to prevent duplicates. These should both be provided as defaults and not require an empty array argument from the workspace/roles function call.

I added `roles: {"*": true}` to the globalvista_template to test this condition in the traverseNestedLocales method.

I added the no_role_locale to the nested_roles/workspace. This is to test the `qualifiedRoles = parentRoles` assignment for the objRoles.length = 0 condition in the traverseNestedLocales method.

```
    "no_role_locale": {
      "layers": {
        "test_role_layer": {
          "role": "test"
        }
      }
    }
```

dbauszus-glx:
We need to test role access from an object and template in a locale and nested locale.

It is to be decided whether the role in an object should be considered.

https://github.com/GEOLYTIX/xyz/pull/2603/commits/3247f59225c9166f31a0a52c218d993b294da600

```
"LAYER_OBJ": {
  "meta": "test role access from an object.",
  "role": "OBJ_ROLE"
},
"LAYER_TEMPLATE": {
 "meta": "test role access from a template.",
 "template": "TEMPLATE_ROLE"
}
```

I updated the test workspace with the two layer in the `uk` locale and the `nested uk.globalvista` locale. Both roles `OBJ_ROLE` and `TEMPLATE_ROLE` are only caught in the `uk` locale. Not in the nested `uk.globalvista`

image

dbauszus-glx:
@RobAndrewHurst 

The _workspace/getNestedLocales method returns an array of nested locales. This must be part of the _workspace/cacheTemplates method. We need to iterate over locales and nested locales, and their layers. It shouldn't be required to traverse the locales thereafter for roles as all locales and their templates would have been cached.

I am aware that the process is not efficient and may take too long with many templates having to be retrieved through network fetches. This could be addressed in a seperate ticket. Any templates from src are fetch synchronous. The templates are marked and the process is repeated until any template src defined in a fetched template is loaded.

Thereafter we iterate through locale, nested locale, and layer async. This should be fast since no additional resources have to fetched through the network.

Any role logic should be handled in the mergeTemplates module. We may have to abandon the localeRole idea and just handle the `role:string` property of the [template] object. The nature of the template [ie locale, layer, infoj entry] should not matter.

RobAndrewHurst:
> I don't think that traversing the locales for roles in the workspace is practical or can work. Layer from templates in nested locales will not have been loaded.
> 
> I added a [commit](https://github.com/GEOLYTIX/xyz/pull/2603/commits/fd65bec593870d46728be9397945721a146f4eee) in which I cache all nested locales with all layers from templates.
> 
> This works fairly well. However the role assign in the template merge which needs to handle nested roles does not work. I have a feeling this is linked to the localeRole logic which doesn't work.
> 
> Image

image
The commit you changed now results in locales roles being repeated from children

RobAndrewHurst:
> I don't think that traversing the locales for roles in the workspace is practical or can work. Layer from templates in nested locales will not have been loaded.
> 
> I added a [commit](https://github.com/GEOLYTIX/xyz/pull/2603/commits/fd65bec593870d46728be9397945721a146f4eee) in which I cache all nested locales with all layers from templates.
> 
> This works fairly well. However the role assign in the template merge which needs to handle nested roles does not work. I have a feeling this is linked to the localeRole logic which doesn't work.
> 
> Image

Yeah, this must be happening in the mergeTemplates. I am looking at it now

dbauszus-glx:
I have updated the locales test to fail as expected.

A user with the`uk.coremarkets.brand_b` should also have the `uk` and `uk.coremarkets` roles.

The question is whether all nested locales should be returned?

Previously the logic was to only return top order locales and nested locales only when they are requested on a top order locale.

The reason for this being that we do not want to parse 100 of nested locales but only the locales nested in one locale if requested to do so.

I think we just need to skip the nested locales in this loop.

image

I updated the locales loop here. https://github.com/GEOLYTIX/xyz/pull/2603/commits/9a0f3d84079f07db56c9094000017ac160c8e381

dbauszus-glx:
The [locale] caching is too aggressive.

I updated the test to use TEMPLATES_ROLE instead of draw role for the role in a templates template object.

I removed some unused imports and console logs.

I added a structuredClone for the cache.locales.

If you break on the `nested locales w/ Nested Roles` test and then set a breakpoint on the workspace returned from the workspaceCache method you will notice that all the locales are cached.

image

image

RobAndrewHurst:
I have changed the `cacheLocale` to store processed layers in a new property `merged_layers` instead of overwriting the raw `layers` object. Also added in a `structuredClone` in `getLayer` module.

I also added logic to detect if a layer is inherited from a parent locale or template without being overridden.

dbauszus-glx (approved):
Any of my concerns have now been addressed in this PR.

dbauszus-glx (changes requested):
None of the current tests create a role to remove.

Is this part still needed or was this designed to fix a flaw in the role combination logic which has since been addressed in the role merge script itself.

```js
  const rolesToRemove = new Set();
  rolesSet.forEach((role) => {
    const parts = role.split('.');
    // Check if it's a double role like 'locale.locale'
    if (parts.length === 2 && parts[0] === parts[1]) {
      rolesToRemove.add(role);
    }
    // Also remove triple+ nesting of same role
    for (let i = 0; i < parts.length - 1; i++) {
      if (parts[i] === parts[i + 1]) {
        rolesToRemove.add(role);
        break;
      }
    }
  });

  rolesToRemove.forEach((role) => rolesSet.delete(role));
```

dbauszus-glx:
@RobAndrewHurst I removed the merged_layers property as the issues was mutable locales in the cached workspace. Caching the layers in a seperate property will still cache the layers for no purpose. An easier approach is to created a structuredClone of the locales, provide the clone as argument to the cacheLocale method and return the clone from the cacheTemplates method with the locales param true.

The cacheLocale method has been renamed to loadLocale because the locale is not cached but loaded in the provided locales clone.

I updated the documentation for the loadLocale method but struggle with the isDefined concept.

dbauszus-glx:
I formatted the indentation of the localeDropdowns method in the _default body script but struggle to wrap my head around the deeply nested conditions. I think this should be a documented ui elements method.

```js
    async function localeDropdowns(level, list) {
      if (!list || !list.length) return;

      // Construct the expected key for the current selection at this level.
      // For root (level 0), this is just the first segment.
      // For nested (level 1+), it is the cumulative path "A,B".
      const expectedKey = currentPath.slice(0, level + 1).join(',');

      // Find the selected entry object
      const selectedEntry = list.find(l => l.key === expectedKey);

      const dropdown = mapp.ui.elements.dropdown({
        data_id: `locale-dropdown-${level}`,
        // Display short name (last part of path)
        placeholder: selectedEntry
          ? (selectedEntry.name || selectedEntry.key).split(/[/,]/).pop()
          : 'Select Locale',
          entries: list.map(l => ({
          title: (l.name || l.key).split(/[/,]/).pop(),
          key: l.key
        })),
        callback: (e, entry) => {
          window.location.assign(`${mapp.host}?locale=${entry.key}`);
        }
      });

      layersPanel.appendChild(dropdown);

      // If we have a selection at this level, check if we need to show the next level
      if (selectedEntry) {
        // Fetch children for this selection
        const nextList = await mapp.utils.xhr(
          `${mapp.host}/api/workspace/locales?locale=${selectedEntry.key}`
        );

        if (nextList && nextList.length > 0) {
          await localeDropdowns(level + 1, nextList);
        }
      }
    }
```

dbauszus-glx:
I feel that isDefined condition adds unnecessary complexity which is extremely difficult to explain.

For example the germany locale has a layer with the TEMPLATES_ROLE. With the condition the role is not assigned to a nested locale. Without the condition we would get following roles the `germany.TEMPLATES_ROLE` and `germany.nested.TEMPLATES_ROLE` which I believe to be the right result. It should be possible to control whether a user has access to the TEMPLATES_ROLE layer only in the context of the nested locale, but not on the germany locale alone.

```js
      const currentKey = localeKey.split(',').pop();
      const localeDef = locales[currentKey];

      let isDefined = !!localeDef?.layers?.[layerKey];

      if (!isDefined) {
        const templateKey = localeDef?.template || (!localeDef && currentKey);

        if (templateKey) {
          const template = await getTemplate(templateKey);
          if (
            !(template instanceof Error) &&
            Object.hasOwn(template.layers || {}, layerKey)
          ) {
            isDefined = true;
          }
        }
      }

      if (!isDefined && Object.hasOwn(locale.layers, layerKey)) {
        return;
      }
```

dbauszus-glx:
I added documentation to the processRoles method.

I'd also like to document the checkRoles method which is only ever called from the processRoles method but I struggle to understand the use of the isLeaf property.

The not isLeaf condition is never hit in any of the tests.

image

All tests pass if I comment out the isLeaf condition. https://github.com/GEOLYTIX/xyz/pull/2603/commits/99d0e131652d408cd9b37b06b370ddac39792636

RobAndrewHurst:
> I added documentation to the processRoles method.
> 
> 
> 
> I'd also like to document the checkRoles method which is only ever called from the processRoles method but I struggle to understand the use of the isLeaf property.
> 
> 
> 
> The not isLeaf condition is never hit in any of the tests.
> 
> 
> 
> image
> 
> 
> 
> All tests pass if I comment out the isLeaf condition. https://github.com/GEOLYTIX/xyz/pull/2603/commits/99d0e131652d408cd9b37b06b370ddac39792636
> 
> 
> 
> 

The isLeaf just indicates we are at the end of the tree ie the leaf

RobAndrewHurst:
> I am trying to wrap my head around the parentLocale?.roles condition in the getLocale checkRoles method. How can this be tested?
> 
> ```js
> function checkRoles(locale, parentLocale, user) {
>   let validRolesObj = locale.roles;
> 
>   // If nested, we must exclude parent roles from valid roles
>   // to enforce that user has the specific nested role.
>   if (parentLocale?.roles) {
>     validRolesObj = { ...locale.roles };
>     Object.keys(parentLocale.roles).forEach((key) => delete validRolesObj[key]);
> 
>     // If validRolesObj is empty, it means no new roles were generated.
>     const hasSpecificRoles = Object.keys(validRolesObj).length > 0;
> 
>     if (!hasSpecificRoles) {
>       // No specific restrictions on child, so parent roles are enough.
>       validRolesObj = locale.roles;
>     }
>   }
> 
>   // Use Roles.check with the restricted set
>   return Roles.check({ roles: validRolesObj }, user?.roles);
> }
> ```
> 
> I noticed that a layer with an empty role was not returned on a public instance.
> 
> The free locale without a role should have the OSM layer without a role.
> 
> I fixed this with the following condition in the Roles.check() method.
> 
> Image
> I still need to look at the locales request. Currently the roles object is not created for a locale with an access role but no template[s].
> 
> Image
> Hence all these locales are publicly accessible.
> 
> This issue has been fixed here: [7f73710](https://github.com/GEOLYTIX/xyz/commit/7f73710a71b9d893c999dd7541110e517451736d)
> 
> ```
> {
>   "templates": {
>     "OSM": {
>       "display": true,
>       "format": "tiles",
>       "URI": "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png"
>     },
>     "france": {
>       "role": "france",
>       "layers": {
>         "OSM_france": {
>           "template": "OSM"
>         }
>       }
>     },
>     "germany": {
>       "role": "france",
>       "layers": {
>         "OSM_france": {
>           "template": "OSM"
>         }
>       }
>     }
>   },
>   "locales": {
>     "free": {
>       "layers": {
>         "OSM": {
>           "template": "OSM"
>         }
>       }
>     },
>     "europe": {
>       "role": "europe",
>       "locales": [
>         "france",
>         "germany"
>       ],
>       "layers": {
>         "OSM": {
>           "template": "OSM"
>         }
>       }
>     },
>     "uk": {
>       "role": "uk",
>       "layers": {
>         "OSM": {
>           "template": "OSM"
>         }
>       }
>     }
>   }
> }
> ```

Is there a test for the publicly available locales?

dbauszus-glx:
I can't get my head around the dependent locales dropdown. How is that supposed to work? Currently a call for nested locales is sent for a locale without nested locales. 

If I select the europe locale with my user having the europe.france role i get the france locale in the dropdown.

Another call is made for locales nested in the france locale despite that locale not having any nested locales, accessible or otherwise.

dbauszus-glx:
I can now hit this block when I request the europe.france locale but still don't get what is supposed to happen.

the parentLocale has `roles: {europe: true}` the locale has `roles:{france: true, europe.france: true}`

it is attempted to remove the europe role from the france locale but the role doesn't exist on that locale anyways.

```js
  if (parentLocale?.roles) {

    Object.keys(parentLocale.roles).forEach((key) => {
      delete validRolesObj[key]
    });

    // If validRolesObj is empty, it means no new roles were generated.
    const hasSpecificRoles = Object.keys(validRolesObj).length > 0;

    if (!hasSpecificRoles) {
      // No specific restrictions on child, so parent roles are enough.
      validRolesObj = locale.roles;
    }
  }
```

Every works as expected when I comment out the custom checkRoles method in favour of the Roles.check method

https://github.com/GEOLYTIX/xyz/pull/2603/commits/bfd0e9190e175453eda38c71e233831fa7919917

RobAndrewHurst:
@simon-leech I will check this out

RobAndrewHurst:
@simon-leech This is now fixed by ensuring that the template's parent's role context is kept before we merge the object. This prevents the sibling pollution

RobAndrewHurst:
@simon-leech 
image

dbauszus-glx:
The negated role check was impossible difficult to read or debug with duplicate regex in a method called only from the second part of an or condition inside an array filter method.

https://github.com/GEOLYTIX/xyz/pull/2603/commits/8ca3273161670a5417813b9d9bdc56d5f4e1588b

dbauszus-glx:
I am completely lost with resolveLocalePath and localeDropdowns logic.

The currentPath is defined outside the resolveLocalePath method which does somethig.

What is the expected outcome?

My user has the europe.france role which should provide access to the europe locale and the nested europe france locale.

If I select Europe in the dropdown I get the Europe.France nested locale by way of six requests to the locales endpoint.

image

What I would expect is to get the Europe locale with the option to select the nested France locale.

I would like to discuss the expected logic first before digging into code and then define the logic from the way the code works. We should write the code after we define the logic.

The unchanged _default script allows me to do all these things without the need of nested methods and duplicate requests.

https://github.com/user-attachments/assets/36da7126-9a43-41e4-84df-a38c34cca9ba

dbauszus-glx (changes requested):
Please see comments in regards to the nested locales logic in the _default script.

dbauszus-glx:
The roles.fromObj method documentation wasn't clear. This method modifies a Set which is passed recursively. I renamed the method as setInObj, updated tests, and documentation here.
https://github.com/GEOLYTIX/xyz/pull/2603/commits/15f695dd1bfd9ea7b924af516c0f834d2b5c41f5

RobAndrewHurst:
@dbauszus-glx I have refined the logic into something a little bit less complex 😅 

> I am completely lost with resolveLocalePath and localeDropdowns logic.
> 
> The currentPath is defined outside the resolveLocalePath method which does somethig.
> 
> What is the expected outcome?
> 
> My user has the europe.france role which should provide access to the europe locale and the nested europe france locale.
> 
> If I select Europe in the dropdown I get the Europe.France nested locale by way of six requests to the locales endpoint.
> 
> image
> What I would expect is to get the Europe locale with the option to select the nested France locale.
> 
> I would like to discuss the expected logic first before digging into code and then define the logic from the way the code works. We should write the code after we define the logic.
> 
> The unchanged _default script allows me to do all these things without the need of nested methods and duplicate requests.
> 
>  Screen.Recording.2026-02-16.163713.mp4

I have refined the logic as it ws very complex and hard to follow! The two functions `resolveLocalePath` & `localeDropdowns` into one recursive function that will. The path construction has also been simplified as it actually just needed to be a one liner.

The redirects are a difficult issue to solve, because we need to get the initial locale. And then get the children for the user and then direct to the first or only child returned from the nested locales. So it's a call to the initial locale, then it's potential child.

But have a look now and let me know if things are easier to follow

dbauszus-glx:
@RobAndrewHurst  The locales interface is too complex to be put into the _default script.

I created a draft elements module for testing.
```js
    const localesInterface = await mapp.ui.elements.locales(locale, locales);

    // TODO: must be tested with void
    layersPanel.append(localesInterface);
```

The locales dropdown can have the first key from the keys array.

I don't iterate yet but that should be easy to implement.

We just need to step through the keys array to get the locales on each level and append a dropdown for that level.

We can decide whether to add the root to each nestedLocales dropdown to step back.

The locales dropdown logic must not update the location and reload the view.

https://github.com/user-attachments/assets/219abbfa-d6eb-41b2-81a2-5b1875eed153

RobAndrewHurst:
> The locales dropdown logic must not update the location and reload the view.

@dbauszus-glx I am not too sure how this is possible without reloading the mapview object. I think it's something we should work to in the future, but currently it's not possible.

This does relate to a previous issue we had in the past where we wanted to reload the mapview object instead of having to redirect on locale switch.

dbauszus-glx:
@RobAndrewHurst We do not currently reload the view. This is fine.

I unnest the recursive buildDropdown method from the locales method which is exported. This was hard to debug with some properties defined outside the closure of the nested async method. All required params (elements, locale, level, locales) should be passed through the recursive method.

The locales should not be queried again on first pass [level === 0].

An entry should be unshifted to the locales to step back a level.

https://github.com/user-attachments/assets/ef9090a9-960f-4654-b704-a5bba59f553b

A nested locale as keys array property. Excessive path string modification and comparison should not be required. The level param and the locale.keys array property should be enough to determine the current level.

dbauszus-glx:
In order to select the entries in the locales and nested locales dropdowns it is possible to slice the locale.keys array by level.

https://github.com/GEOLYTIX/xyz/pull/2603/commits/5b934fdad0fe2bee25f60f3972f4d6be216934f7

image

simon-leech:
Just done some testing on existing `roles` object and happy that's not been affected by this so backwards compatibility remains ✅ 

Onto more `role` testing now 🫡

cityremade (approved):

PR #2605, opened 2026-01-19
fix(node-pg): Don't re-run queries that timeout

RobAndrewHurst:
### Summary

This PR addresses the issue of long running queries which timeout result in an instance being bogged down. 

### Changes
This PR checks for the error code '57014' being returned from the pg server which indicates that the query timeout and should not be re-run.

### DIFF
--- a/mod/utils/dbs.js
+++ b/mod/utils/dbs.js
@@ -99,6 +99,9 @@ async function clientQuery(pool, query, variables, timeout) {
         variables,
       });
 
+      // Statement timeout
+      if (err.code === '57014') return err;
+
       retryCount++;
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Tested by setting my local `STATEMENT_TIMEOUT` env variable to "100". 

Before this PR, retry = 3. 
On this PR, just ran it once and then stops.

dbauszus-glx (approved):
Nice one. Tested this with SQL delay query which takes longer than the timeout.

```json
    "pg_sleep": {
      "template": "select pg_sleep(%{seconds});",
      "dbs": "NEON"
    }
```

PR #2623, opened 2026-01-21
fix(paramString): Special Character Encoding

simon-leech:
## Description

I've added a new test that highlights this issue. 
Providing a parameter to the `paramString` with special characters is not handled properly. 
e.g. `text: "TEST & MORE"`

I haven't fixed it as need some help on that! 

**We probably need to set up the front-end tests to run on the test runner too? As we they don't currently**

## Type of Change

- ✅ Testing

## How have you tested this?

Added tests to show its failing.

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ New Tests Added

### DIFF
--- a/lib/utils/paramString.mjs
+++ b/lib/utils/paramString.mjs
@@ -58,7 +58,7 @@ function paramString(params) {
         return `${entry[0]}=${encodeURIComponent(JSON.stringify(entry[1]))}`;
       }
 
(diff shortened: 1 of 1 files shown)


RobAndrewHurst:
The `paramString` utility has been updated to use `encodeURIComponent` for parameter values and not the entire param. 

This ensures that special characters like &, which have reserved meanings in URL parameters, are correctly encoded (e.g., & becomes %26).

Issue #2620, opened 2026-01-21
Gazetteer: Unable to search via dataset


simon-leech:
**Issue**
In v4.20.2 you are using to search using a `datasets` array. 

You won't hit a breakpoint inside the `onLoad` method currently.
<img width="512" height="218" alt="Image" src="https://github.com/user-attachments/assets/fc1a9424-92f4-4c2c-9407-e2ff2bd03a53" />

I think this is to do with this PR - #2548  which removes the `xhr.onLoad` method. 

**Sample Configuration** 
``` json
 "gazetteer": {
      "placeholder": "Search for a place eg. Chester",
      "datasets": [
        {
          "layer": "layer",
          "table": "schema.table",
          "qterm": "name",
          "leading_wildcard": true,
          "limit": 5,
          "no_result": null
        }
      ]
    },
```

**Solution Required**
- Fix the bug so searching works
- Add a test to prevent this happening again.

PR #2624, opened 2026-01-22
fix(xhr): Restore call to provided onLoad function

AlexanderGeere:
## Description
This PR restores the call to the provided onLoad function.

This was removed here #2548.

This was preventing the gazetteer dataset calls.

## GitHub Issue
#2620 

## Type of Change
- ✅ Testing

## How have you tested this?
tested on `bugs_testing/filters/all_filters.json` and wiht new tests.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/utils/gazetteer.mjs
+++ b/lib/utils/gazetteer.mjs
@@ -137,7 +137,7 @@ function search(term, dataset) {
         return;
       }
 
-      // Ensure that response if a flat array.
+      // Ensure that response is a flat array.
       [response].flat().forEach((row) => addRow(dataset, layer, row));
     })
     .catch((err) => {


RobAndrewHurst:
@AlexanderGeere The debounce logic is quite working as expected. I can see some requests still getting through

AlexanderGeere:
> @AlexanderGeere The debounce logic is quite working as expected. I can see some requests still getting through

@RobAndrewHurst I assigned debounce false to the params in the debounce handler to keep the reference to params. 
I tested on `bugs_testing/plugins/h3_layer.json` and the network tab looks the same as patch now as far as i can tell

RobAndrewHurst:
> I'm seeing duplicated responses for the same dataset search

Yeah that was the issue I was pointing out in the stand up. @AlexanderGeere found it when it fixed the datasets being added as rows to the results.

dbauszus-glx:
I removed the legacy dataset.onload method.

dbauszus-glx:
I did not like how the xhr module handled debounce. Using a config decision method to call handler method which calls the xhr method recursively without the option change the global key.

The debounce method will now be called at the end of the xhr utility method and passes the XMLHttpRequest object.

The request will be sent if there is nothing to debounce. The debounce method returns false.

If the debounce property is a string then this will be assigned as the key for the request.

If the debounce property is a number then this will be assigned as the delay for the request timeout.

The request will be sent after a timeout or the xhr will be rejected if the timeout is debounced.

The dataset layer key will be used for the debounce.

Dataset gazetteer requests should be cached.

dbauszus-glx:
I didn't bother with the test since the mapp [frontend] test logic will soon change.

PR #2625, opened 2026-01-29
try xhr and warn on failure to get location.

dbauszus-glx:
In the xhr utility it is clearer to read that the function should short circuit on debounce rather than checking the negation and xhr.send if not debounce.

The xhr.send should be tried as otherwise the error passed to the reject promise method in the xhr.onload method is not handled.

The call to the xhr util in the location/get getFeatureResponse method should be tried and an error returned.

The location get method should remove the location from locations set as well as the location.hook. The user should be alerted if the location get method fails.

The .vscode folder should be excluded from biome format checks.

### DIFF
--- a/lib/mapp.mjs
+++ b/lib/mapp.mjs
@@ -27,7 +27,7 @@ import utils from './utils/_utils.mjs';
 
 hooks.parse();
 
-const ol_current = 10.6; //25/06/2025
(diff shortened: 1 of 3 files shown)


dbauszus-glx:
@RobAndrewHurst 

On success:
image

on error:
image

because we are awaiting the promise to be resolved in the try...catch.

PR #2629, opened 2026-02-02
Drawer options

cityremade:
Removed repetitions across modules:
* checks now moved to drawer ui element.
* caret icon now created in one go on expandable elements, always in the top right corner.
* data-id attributes passed into all element types.

Removes checks on drawer options for the following modules:

* layers/view
* elements/drawing
* panels/draw
* panels/filter
* panels/reports
* panels/style
* locations/infoj
* locations/view

Drawer options retained: `drawer: undefined`, `drawer: false`, `drawer: null`. 
Not great but it has possibly spread by now.

### DIFF
--- a/lib/ui/elements/drawer.mjs
+++ b/lib/ui/elements/drawer.mjs
@@ -31,7 +31,7 @@ export default function drawer(params) {
       class="${'drawer ' + (params.class || '')}"
       data-id=${params.data_id}>
       <div class="header">${params.header}</div>
-      ${params.content}`;
+      <div class="content">${params.content}`;
 
     return params.drawer;
   }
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
I added the layer.drawer config to the layer.viewConfig object and removed a comment fragment `//}` froom layers/view.

I clean up reports module to work with regular links provided as href property. The icon should also be configurable. HTML fragments are provided in an array instead of creating a node to append nodes.

I test layer.view, location.view, drawing, styles, dataviews, and report which all work as expected.

Filter however do not work.

I can add and remove a filter in the default filter drawer.

https://github.com/user-attachments/assets/8109f226-1715-4675-8c2c-7378a4c0f807

With the drawer: false option the filter element is not displayed.

https://github.com/user-attachments/assets/2f09654e-695a-41ac-aa4f-4c612d728cba

Can you also look at the weird underline of the flat button used to clear the filter?

cityremade:
The problem with hidden drawers was line 89:
Screenshot 2026-02-06 at 11 31 21

Each full-on drawer has a `.content` element in it which was called in filters being added to their containing drawer.
Simplified drawer containers (false/null) didn't have that. For these elements content now wrapped in `.content`.

PR #2633, opened 2026-02-03
CSS - search input rules (patch)

cityremade:
Safari searchbox css fix but on patch.

This gazetteer/searchbox problem was detected on MacOS Safari.

The click on the list item wouldn't fire the callback function - i.e. the map wouldn't respond to the selected search.

It would merely make the `<ul>` element disappear as the first rule to execute would be CSS display 'block' to 'none' switch.

Here's the culprit:

Screenshot 2026-02-03 at 14 40 28

The search callback would never happen as the click would first switch display property and then the `<li>` target element would no longer be accessible in the document. So no click event callback afterwards.

Safari is less forgiving of these workflows and would execute css first. Yet once an element hits "display:none" it cannot be accessed by the browser.

I have removed the display: none / display: block switch from `<ul>` elements in searchbox lists.
This is not needed as the lists are cleared and filled in accordance with returned search results.
Empty `<ul>` takes no space anyway while it is still accessible in the DOM.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -26,7 +26,8 @@
       "!package.json",
       "!jsdoc_mapp.json",
       "!jsdoc_xyz.json",
-      "!jsdoc_test.json"
+      "!jsdoc_test.json",
+      "!**/.vscode"
     ]
   },
   "formatter": {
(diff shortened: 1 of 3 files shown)


cityremade:
this may need scoping out - when should the entries be cleared?

- after item is clicked and the search callback runs?
- after input is out of focus?
- else?

This was not a case before as the list would be switched to display: none.

dbauszus-glx:
The searchbox results should not be displayed when out of focus. However the display can be preserved with the hover condition on the li element. https://github.com/GEOLYTIX/xyz/pull/2639

PR #2636, opened 2026-02-03
Disable location view edits with toggleLocationViewEdits false

dbauszus-glx:
Location view edits will be disabled with `layer.toggleLocationViewEdits=false`. The toggle to enable edits will not be added to the location view header.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -26,7 +26,8 @@
       "!package.json",
       "!jsdoc_mapp.json",
       "!jsdoc_xyz.json",
-      "!jsdoc_test.json"
+      "!jsdoc_test.json",
+      "!**/.vscode"
     ]
   },
   "formatter": {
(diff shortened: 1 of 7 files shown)


AlexanderGeere:
@dbauszus-glx 
If you have edit true on entries in the infoj, why would you want to disable editing?

AlexanderGeere:
> @dbauszus-glx If you have edit true on entries in the infoj, why would you want to disable editing?
>

This would simplify the edit role that simon has been talking about, just a template that has `toggleLocationViewEdits: false`.

dbauszus-glx:
@AlexanderGeere That's the reason. Been talking with @simon-leech yesterday and it should be easier to control editing on the layer without the need to alter individual infoj entry edit config.

simon-leech:
All good with me too 👍

PR #2638, opened 2026-02-04
Filter and legend should updated each other

dbauszus-glx:
The ui/layers/legends/utils addFilter and removeFilter must be async to await the creation of the filter element. Otherwise the filter element will be replaced with an [object/promise] object.

The call to the mapp.ui.elements.pills() method does not acknowledge existing pills. Current filter values should be created as pills on creation.

The legend should be recreated in the timeout of the ui/layers/filters applyFilter method.

We need to understand why the display property is removed at the end of the ui/layers/legends/categorized method. I commented this out for now as the provided comment is meaningless.

```
layer.style.legend.style.removeProperty('display');
```

### DIFF
--- a/lib/ui/elements/dropdown.mjs
+++ b/lib/ui/elements/dropdown.mjs
@@ -257,7 +257,10 @@ function onSearchInput(e, params) {
       e.target.value = '';
       e.target.dispatchEvent(new Event('blur'));
 
-      params.callback?.(e, [...params.selectedOptions]);
+      // The pills method will fire a callback method for the option.
+      if (!params.pills) {
+        params.callback?.(e, [...params.selectedOptions]);
+      }
     }
   });
 }


dbauszus-glx:
I fixed the assignment for the cat.disabled property in the createLegend method. The array findIndex did not take into account that the index of the first value is 0. This works with array.includes condition.

The legend will now be redrawn in the timeout of the applyFilter method.

@cityremade can you have a look at the call for the mapp.ui.elements.pills in the filter_in method. Existing filter values should be created as pills.

simon-leech (changes requested):
This isn't working quite right - initial filter isn't being set for me in the filter pills.
This is my filter configuration 
``` json
  "filter": {
        "type": "in",
        "distinct": true,
        "dropdown_pills": true,
        "dropdown_search": true
      },
```

dbauszus-glx:
> This isn't working quite right - initial filter isn't being set for me in the filter pills. This is my filter configuration
> 
> ```json
>   "filter": {
>         "type": "in",
>         "distinct": true,
>         "dropdown_pills": true,
>         "dropdown_search": true
>       },
> ```

This can not work. Categorical legend filter are "ni" type. 

The pills are only checked with the `filter.searchbox` option. The dropdown_pills is a boolean provided to the elements/dropdown method with `if (filter.dropdown || filter.dropdown_pills || filter.dropdown_search)`.

image

dbauszus-glx:
The filter method needs to support in and ni type filter. Only ni type filter are linked to categorical legend filter.

I don't think it is required to check the length. The property value should be undefined or array, possibly empty. Either should work if provided as value for the pills property.

https://github.com/GEOLYTIX/xyz/pull/2638/commits/fad4b7310c5d6c9490066a02baf5acaf4f31f94c

dbauszus-glx:
@simon-leech what should be the difference between `searchbox: true` and `dropdown_pills: true, dropdown_search: true`?

simon-leech:
@dbauszus-glx Understood on `ni` vs `in`. 

There should be no difference between pills being created for a `filter.current` on load for `searchbox`, or for `dropdown_pills` with `dropdown_search`.

dbauszus-glx:
The callback from onSearchInput must not be fired after the callback from the pills element is fired.
https://github.com/GEOLYTIX/xyz/pull/2638/commits/6e093060d67359c0b868befa44eb3bda7d915176

PR #2639, opened 2026-02-04
fix(Searchbox): Display seachbox on hover condition

dbauszus-glx:
The searchbox should be displayed as block on hover.

This is to ensure that a blur out of focus event doesn't hide the searchbox. Eg. on click.


cityremade:
ok search now fires after selection.

simon-leech:
Weirdly I don't have an approval button on GitHub at the moment, not sure if a GitHub bug but this is working fine for me so I'll merge it into patch 👍

PR #2642, opened 2026-02-09
fix(Layer/Decorate): GetExtent may return array of [null,null,null,null]

simon-leech:
# Pull Request Template

## Description

This PR addresses an issue where the getExtent may return an array of null values. 
The checks have been moved into the `getExtent` function instead of after the `getExtent()` method is called.
Thus, the check can now be on if the response is false (meaning no extent to zoom to). 

## Type of Change

## How have you tested this?
Tested locally when the extent returned is null. 
This can be replicated by adding a default filter to a layer that does not have any data for the filter value, and then trying to zoom to it.

## Testing Checklist
- ✅ Updated Existing Tests
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/layer/decorate.mjs
+++ b/lib/layer/decorate.mjs
@@ -143,11 +143,9 @@ async function getExtent() {
     layer.featureSet instanceof Set
   ) {
     ids = [...layer.featureSet];
-  } else if (layer.L && layer.format !== 'mvt') {
-    // return extent from Openlayers source.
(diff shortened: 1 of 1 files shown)


simon-leech:
@cityremade Sonarcloud prefers the use of `Array.includes` so I've left it as that!

AlexanderGeere (approved):

PR #2644, opened 2026-02-12
Add more colours to the default Listview

simon-leech:
By default the listview ends at J. 
There is no specific reason for this, and we are getting lots of client feedback that they want to be able to select more locations in their usage. 
I've added some more colours to take it up to Z (so 26 options now). 
This is none breaking and just increases the options.

### DIFF
--- a/lib/mapview/addLayer.mjs
+++ b/lib/mapview/addLayer.mjs
@@ -26,7 +26,7 @@ A renderComplete event will be assigned to the Openlayers mapview.Map{}. This pr
 
 @returns {array} The array of decorated layers is returned.
 */
-export default async function addLayer(layers) {
+export default async function addLayer(layers = []) {
   // A single JSON layer is provided.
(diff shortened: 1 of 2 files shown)


cityremade:
Here's the new branding palette, how about using these plus more relevant additions to complete the list ?

yellow #FEE376
peach #FFC38D
lime #BBF77C
teal #97F2C7
blue #76D8EE
purple #AEBFFE
pink #FFADDE
orange #F79646

image (2)

cityremade:
what do you think?

[colours.html](https://github.com/user-attachments/files/25265555/colours.html)

dbauszus-glx:
> what do you think?
> 
> [colours.html](https://github.com/user-attachments/files/25265555/colours.html)

image

+1

cityremade:
since we got 26 selections I'd drop those deep brown shades.

PR #2647, opened 2026-02-12
Call mapview.changeEnd method after adding layers.

dbauszus-glx:
The changeEnd event method of a layer may not be called if there is no map to render. This would lead to a zoom restricted layer not being disabled on initial load. The changeEnd method is debounced and can be called after or before the renderComplete event.


RobAndrewHurst (approved):
Nice one!

simon-leech:
Finallyyyyy managed to reproduce this bug! Nice 👍

PR #2656, opened 2026-02-17
report link element

### LINKED ISSUE #2655: link/report element with dialog property option
There should be a ui/elements/report module with a method to generate a report / link element.

A link element will either be a default link to be opened in a blank target [although it should be possible to control the target in the configuration] or a href created from the hooks and provided template property.

The ui/layers/panels/reports module generated the link element like so:
```js
    report.key = key;

    report.host = layer.mapview.host;

    report.href ??= `${report.host}/view?${mapp.utils.paramString({
      lat: mapp.hooks.current?.lat,
      lng: mapp.hooks.current?.lng,
      template: report.template,
      z: mapp.hooks.current?.z,
    })}`;

    report.icon ??= 'summarize';

    reports.push(mapp.utils.html.node`<a
      class="link-with-img"
      target="_blank"
      href="${report.href}">
      <span class="notranslate material-symbols-outlined">${report.icon}</span>
      <span>${report.title || report.key}`);
```

The ui/locations/entries/link module which is used for type:link and type:report entries creates the element like so:
```js
  if (entry.report) {
    // Assign URL path for report.
    entry.url ??= `${entry.location.layer.mapview.host}/view?`;

    // Assign URL params for report.
    Object.assign(entry.params, {
      id: entry.location.id,
      layer: entry.location.layer.key,
      locale: entry.location.layer.mapview.locale.key,
      template: entry.report.template,
    });

    // Assign entry.label for link text.
    entry.label ??= `${entry.report.label || mapp.dictionary.report}`;
    // Assign entry.icon_name to report icon if not defined.
    entry.icon_name ??= 'description';
  }

  if (!entry.url) {
    console.warn(`An entry.url must be defined for the URL path.`);
    return;
  }

  // Construct href from URL + params.
  const href = entry.url + mapp.utils.paramString(entry.params);

  entry.data_id ??= 'link';
  entry.label ??= `${mapp.dictionary.link}`;

  entry.icon_name ??= 'open_in_new';
  entry.icon_class ??= '';
  const icon_class = `notranslate material-symbols-outlined ${entry.icon_class}`;

  entry.link_class ??= '';
  const link_class = `link-with-img ${entry.link_class}`;

  const node = mapp.utils.html.node`<div class="val">
    <a 
      data-id=${entry.data_id}
      target="_blank" 
      class=${link_class}
      href=${href}>
      <span 
        style=${entry.icon_style || ''}
        class=${icon_class}>${entry.icon_name}</span>
      <span>${entry.label}</span>`;
```

Both methods should create the link element through a element module.

The element module should check for a dialog:{} object property. If configured the link should be opened in an iframe in a dialog.

It should be possible to define the size, mode and header in the dialog object config. Modal should be the default.

dbauszus-glx:
The idea is to have one element module that creates a link element which can be configured to open in a dialog.

There should not be duplicate code for link / report in location and layer panel. The element should look and behave the same in the layer and location view.

### DIFF
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -43,7 +43,10 @@ export default function reports(layer) {
 
     report.layer = layer;
 
-    const el = mapp.ui.elements.reportLink(report);
+    // Provide the report as params property for the reportLink method if href is not explicit.
+    const params = report.href ? report : { report };
+
+    const el = mapp.ui.elements.reportLink(params);
 
     reports.push(el);
   }
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -43,7 +43,7 @@ export default function reports(layer) {
 
     report.layer = layer;
 
-    const el = mapp.ui.elements.reportLink({ report });
+    const el = mapp.ui.elements.reportLink(report);
 
     reports.push(el);
   }
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -39,24 +39,13 @@ export default function reports(layer) {
 
     report.key = key;
 
-    report.host = layer.mapview.host;
-
-    report.href ??= `${report.host}/view?${mapp.utils.paramString({
-      lat: mapp.hooks.current?.lat,
-      lng: mapp.hooks.current?.lng,
-      locale: layer.mapview.locale.key,
-      template: report.template,
-      z: mapp.hooks.current?.z,
-    })}`;
+    report.layer = layer;
 
     report.icon ??= 'summarize';
 
-    reports.push(mapp.utils.html.node`<a
-      class="link-with-img"
-      target="_blank"
-      href="${report.href}">
-      <span class="notranslate material-symbols-outlined">${report.icon}</span>
-      <span>${report.title || report.key}`);
+    const el = mapp.ui.elements.reportLink(report);
+
+    reports.push(el);
   }
 
   if (layer.reports.dialog) {
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -39,6 +39,8 @@ export default function reports(layer) {
 
     report.key = key;
 
+    report.label ??= key;
+
     report.layer = layer;
 
     report.icon ??= 'summarize';
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -43,7 +43,7 @@ export default function reports(layer) {
 
     report.icon ??= 'summarize';
 
-    const el = mapp.ui.elements.reportLink(report);
+    const el = mapp.ui.elements.reportLink({ report });
 
     reports.push(el);
   }
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -41,8 +41,6 @@ export default function reports(layer) {
 
     report.layer = layer;
 
-    report.icon ??= 'summarize';
-
     const el = mapp.ui.elements.reportLink({ report });
 
     reports.push(el);
--- a/lib/ui/layers/panels/reports.mjs
+++ b/lib/ui/layers/panels/reports.mjs
@@ -34,19 +34,22 @@ Specifying `layer.reports.drawer: false` will prevent a drawer from being made f
 export default function reports(layer) {
   // Create chkbox controls for each dataview entry.
   const reports = [];
-  for (const [key, report] of Object.entries(layer.reports)) {
+  for (const [key, entry] of Object.entries(layer.reports)) {
     if (typeof report === 'boolean') continue;
 
-    report.key = key;
+    entry.key = key;
 
-    report.label ??= key;
+    entry.label ??= key;
 
-    report.layer = layer;
+    entry.layer = layer;
 
-    // Provide the report as params property for the reportLink method if href is not explicit.
-    const params = report.href ? report : { report };
+    if (entry.template) {
+      entry.report = {
+        template: entry.template,
+      };
+    }
 
-    const el = mapp.ui.elements.reportLink(params);
+    const el = mapp.ui.elements.reportLink(entry);
 
     reports.push(el);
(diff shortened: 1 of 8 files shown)


cityremade:
This is what I got rn. 
What I don't like here is one module handling 2 types `link` and `report`, either is taken in by `link entry` module.
`template` used to be nested inside `report`, should this be retained both in layer and entry report definition?

_removed outdated definition_

simon-leech:
@cityremade Template must be retained in the report object as it currently is. 
This PR shouldn't cause any breaking changes to the existing configuration.

simon-leech:
Also, the `entry.params` object must still correctly pass those values into the report. 
The example below would add `&key=report_key` to the URL.
e.g. 
``` json
 {
      "type": "report",
      "label": "Report",
      "report": {
        "template": "report"
      },
      "params": {
        "key": "report_key"
      }
    },
```

cityremade:
```json
{
  "reports": {
    "layer_report": {
      "label": "Layer Select",
      "data_id": "layer_report",
      "template": "custom"
    },
    "layer_report2": {
      "label": "Layer Select 2",
      "data_id": "layer_report2",
      "template": "custom"
    }
  },
  "infoj": [
    {
      "type": "report",
      "data_id": "location_report",
      "report": {
        "label": "Location Select",
        "template": "custom"
      },
      "params": {
        "key": "params_report_key"
      }
    },
    {
      "type": "link",
      "url": "https://google.com",
      "label": "My Link"
    }
  ]
}
```

AlexanderGeere (changes requested):
Working as expected so far. 

I nullish assigned the key to the label as this is what was happening before. 

I was just wondering if we could maybe make the iframe link and the regular one look the same in the reports panel, 

Image

The top one is the iframe option, I think the icon is enough of a delineation.

I think  the link without the outline looks better 

This can be tested on: `bugs_testing/drawer/workspace.json`

simon-leech (dismissed):
The report icon has changed from the default page icon to a open in new tab - this is wrong, this PR shouldn't affect existing report set up at all. 

Do we need the `const label = [].find(Boolean)` or could we use nullish assignment to the params.label for example? 

I see a TODO about icon_class, this should be retained as it lets us configure the icon_class when needed, e.g. to add border for example.

cityremade:
When report is configured to display in an iframe the module returns a button. Button => open modal.

Regular report or link to report view will be a link / anchor tag.

Label array find is meant to find a label it can display as there's some inconsistency whether it's label, title or report.label etc. This finds the boolean or falls back to dictionary value if none is set. Order in the array matters. First one non-falsy is used as label.

dbauszus-glx:
It should be possible to define a href link for the the reports panel. 

```
        "reports": {
          "test": {
            "title": "test",
            "href": "github.com"
          },
          "geolytix": {
            "href": "https://geolytix.com"
          }
        }
```

This works if the lib/ui/layers/panels/reports provides the report entry as argument for the reportLink method.

Either the reportLink panels/report method need to account for the provision of the href property.

https://github.com/GEOLYTIX/xyz/pull/2656/commits/6d0ea3dc0c7ca3b6100cbf05cc27aa44a1b1e734

I think this can be done like so:
```js
    // Provide the report as params property for the reportLink method if href is not explicit.
    const params = report.href ? report : { report };

    const el = mapp.ui.elements.reportLink(params);
```
https://github.com/GEOLYTIX/xyz/pull/2656/commits/f845a0a86231abba4ae6a5885cec42ff4348cc65

simon-leech (commented):
The dialog link for a report, and the report link when opening in a new tab should be the same I think. 
At the moment one has an outline and isn't bolded and one is.

cityremade:
@simon-leech I want the button to show the dialog. This would be incorrect use of anchor tag to do something else that open a url in the same window or a new tab. Dialog opens on a plain click.

simon-leech:
We shouldn't be deciding that different elements have outlines and some don't. 
The report whether in a dialog or new tab should look how report currently does. 
It should be possible to configure using the icon_class and outline to make it look how you have it currently @cityremade - but this shouldn't be decided.

Otherwise its impossible to remove the outline, it should be off by default and configurable on.

AlexanderGeere (approved):

dbauszus-glx:
It must also be possible to assign the dialog flag to a link. https://github.com/GEOLYTIX/xyz/pull/2656/commits/95ecf3ad74da1b057d70e600ef4ef620f08f646c

image

dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2656/changes/1734546ce27795d52fe043516dfab0318fa31905

Instead of:
```js
  if (params.url) {
    params.href = params.url;
  }
```
use:
```js
params.href ??= params.url; //href will be undefined if url is undefined.
```

Assign onclick if needed and not return a conditional method inside an inline method in template that returns null or prevents default.
```js
  if (params.dialog) {
    params.header ??= params.label;
    anchor.onclick = (e) => {
      e.preventDefault();
      iFrameDialog(params);
    };
  }
```

Assign defaults to props instead of using const this = that depending with multiple checks on the same condition.
```js
  params.link_class ??= '';
  params.link_class += ' link-with-img';
```

Or
```js
params.label ??= params.report?.label || params.title || mapp.dictionary.link;
```
instead of
```js
  const label = [
    params.report?.label,
    params.label,
    params.title,
    mapp.dictionary.link,
  ].find(Boolean);
```

Target can be _blank if you prevent the default in the click no need to check the condition again to assign target null. It should be possible to configure params.target just like any other params property.

lib/ui/layers/panels/reports.mjs
```diff
@@ -34,29 +34,24 @@ Specifying `layer.reports.drawer: false` will prevent a drawer from being made f
 export default function reports(layer) {
   // Create chkbox controls for each dataview entry.
   const reports = [];
-  for (const [key, report] of Object.entries(layer.reports)) {
+  for (const [key, entry] of Object.entries(layer.reports)) {
     if (typeof report === 'boolean') continue;
```
RobAndrewHurst: This will now never work because report was renamed to entry

RobAndrewHurst (changes requested):

RobAndrewHurst (dismissed):

RobAndrewHurst (approved):

simon-leech (approved):

PR #2662, opened 2026-02-19
Add consent plugin

### LINKED ISSUE #2508: Consent plugin [Cookie/Data]

dbauszus-glx:
This PR adds a core plugin which awaits a confirm dialog to store consent in the indexedDB user store.

The consent text must be configured.

### DIFF
--- a/lib/plugins/_plugins.mjs
+++ b/lib/plugins/_plugins.mjs
@@ -6,6 +6,7 @@ The module exports a collection of core mapp plugins.
 @module /plugins
 
 @requires module:/plugins/admin
+@requires module:/plugins/consent
 @requires module:/plugins/feature_info
 @requires module:/plugins/fullscreen
 @requires module:/plugins/layer_order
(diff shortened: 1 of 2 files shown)


simon-leech:
@dbauszus-glx Would it be possible to allow a query to populate the content rather than just a text property? 

The consent may need to be re-approved following a new version of the text which would be stored in a db so a query would be required

dbauszus-glx:
@simon-leech Yes. Possible but out of scope for the initial implementation.

We also want to support templates instead of queries.

And the confirm dialog does not currently support html content. So this will be another patch. Because once we support the content to be retrieved from a src or query we also want to support html content.

The scope for this PR is to store the content on the user object.

AlexanderGeere (approved):
Working as expected, checked on a public instance as well

RobAndrewHurst (approved):

PR #2664, opened 2026-02-20
fix(logger): Refactor logger to use shared database connection pool

RobAndrewHurst:
The logger module was creating a connection pool each time it logged to the db.

## What has been done?

- Removed the dedicated pg.Pool from the logger module, which was creating a separate database connection pool alongside the shared one managed by mod/utils/dbs.js.
- The PostgreSQL logger now uses a dynamic import('./dbs.js') to reuse the existing shared pool, avoiding the circular dependency (dbs.js imports logger.js).
- Table name sanitization moved from per-log-call to initialization time.
- Logging queries now benefit from the retry logic and connection management already built into dbs.js.


AlexanderGeere (approved):
Logs are reaching the db as before

Issue #2600, opened 2026-01-16
Re-run options query for location view drop-down


jfitzpatrick17:
If a distinct value is added in a location update, the query for distinct values is not re-run on the location update. 

The way it should be handled is by dependents.

Anytime a field entry is a dependent then it should populate the options.
Or a flag - anytime a location is updated.

May already exist

dbauszus-glx:
This issue will be addressed in the conditional check of the ui/locations/entries/text textedit method. The query populates the entry.edit.options[] array. If the array has values the query will not be run again.

```js
    // If options is an array and contains values, we can create a dropdown.
    if (entry.edit.options.length) {
      // Create dropdown from options.
      textoptions(entry);
    }
    // If options is empty array, we need to query the table to populate it.
    else {
      // We can query a particular template or Query distinct field values from the layer table.
      mapp.utils
        .xhr(
          `${entry.location.layer.mapview.host}/api/query?` +
            mapp.utils.paramString({
              field: entry.json_field || entry.jsonb_field || entry.field,
              filter: entry.location.layer.filter?.current,
              id: entry.location.id,
              key: entry.jsonb_key || entry.json_key,
              layer: entry.location.layer.key,
              locale: entry.location.layer.mapview.locale.key,
              table: entry.location.layer.tableCurrent(),
              template:
                entry.edit.query ||
                (entry.jsonb_field || entry.json_field
                  ? 'distinct_values_json'
                  : 'distinct_values'),
            }),
        )
```

The easiest option would be to reverse the condition and add a flag, eg. update: true.

```js
if (!entry.edit.options.length || entry.edit.update) {
  // Run query
} else {
  textoptions(entry);
}
```

The ghastly template condition `template: entry.edit.query || (entry.jsonb_field || entry.json_field? 'distinct_values_json' : 'distinct_values')` wants to be cleaned up as well.

dbauszus-glx:
This has been resolved in the linked patch PR.

PR #2671, opened 2026-02-20
fix(text entry): Rerun the distinct query for options on a text entry

AlexanderGeere:
## Description
The distinct query will be re run if an entry using the same field as the dropdown is updated. 

This is achieved by an update flag being set on every entry that uses the same field in the infoj. 

## GitHub Issue
#2600 

## Type of Change

## How have you tested this?
Tested locally on `bugs_testing/editing/options_dropdown.json`

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/location/decorate.mjs
+++ b/lib/location/decorate.mjs
@@ -36,6 +36,7 @@ A decorated mapp location.
 @property {function} remove {@link module:/location/decorate~remove}
 @property {function} syncFields {@link module:/location/decorate~syncFields}
 @property {function} update {@link module:/location/decorate~update}
+@property {string} [updateTemplate='location_update'] The query template for the location update post query.
 @property {HTMLElement} [view] The location view displayed in the location listview in the default Mapp application view.
 */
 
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
@AlexanderGeere 

I unnest the onkeyup event method and added documentation for this.

I added the `if (!entry.edit.options.length || entry.edit.update)` condition to the textoptions method and made that method async.

It is easier to read when the response is awaited and the textoptions was called after the query anyways or else if there is no query.

dbauszus-glx (dismissed):
I don't fully understand the edit.update flag. Why does this flag has to be set or deleted?

I always just defined an empty options array `"options": []` which will trigger the query to run when the textoptions method is executed because the array has no length.

AlexanderGeere:
@dbauszus-glx 

I removed the update flag and instead just empty the options.

RobAndrewHurst (approved):

PR #2676, opened 2026-02-24
check getTemplateObject return

### LINKED ISSUE #2693: String templates from src are not cached
Currently only template objects from src are cached.

The getTemplate module is unnecessary complex with duplicate condition logic and unnecessary dependencies.

dbauszus-glx:
The getTemplate module and methods where unnecessary complex with duplicate conditions and unnecessary dependencies.

This PR cleans up the module structure and documentation.

Apart from modules, any template with a src will be cached in the workspace templates.

Any error conmdition returns a formatted new Error object.


dbauszus-glx:
https://github.com/GEOLYTIX/xyz/pull/2676/commits/6391cc7090343dc5064ff31363aac3222e2a408b

The envReplace utils method is called in the mergeTemplates and doesn't do anything in the getTemplate method. The template.src string will already have variables substituted.

Calling the getTemplateObject method with the template.src string to check whether the template is cached is overkill. This check can easily be performed in the getTemplate method itself.

cityremade (approved):

RobAndrewHurst (changes requested):
This seems to break nested roles in templates.

I have a locale that also has a templates array. In each entry in the template array there is a role in the template.

Where I would normally see uk.demographics.stores. I am just seeing uk in the roles api.

You can test this with `bugs_testing/roles/si_test_feb/workspace.json`

RobAndrewHurst:
Added a runtime-level regression test for nested template roles in 
tests/mod/workspace/_workspace.test.mjs (nested role should still resolve nested locale path).

This test exercises the actual workspace endpoint flow (key: 'locale') for nested locale resolution (uk,brand_a_template) with a user that only has the nested role (uk.stores.brand_a). 

It asserts that access should be granted (200) and that OSM_BRAND_A is present in locale.layers.
Current behavior returns Role access denied. 

This test intentionally fails and demonstrates the bug in a production-like path (not just isolated merge/unit logic).

dbauszus-glx:
@RobAndrewHurst This PR does not touch role evaluation and the test fails on the minor as well.

dbauszus-glx:
@RobAndrewHurst I removed the test which fails on the patch branch and created a separate ticket for this test. https://github.com/GEOLYTIX/xyz/pull/2703

dbauszus-glx:
This doesn't need to be a minor PR. I created a patch PR in which the envReplace is restored. https://github.com/GEOLYTIX/xyz/pull/2706

PR #2677, opened 2026-02-24
fix(Gazetteer): Check debounce of request

simon-leech:
## Description
The gazetteer was displaying records with no data attached, and was also showing outdated data. 

This PR makes the XHR return object with debounce true, and the gazetteer then check for a debounce value.

## Type of Change

## How have you tested this?
Test locally using a datasets search and typing quickly, you will see multiple empty records and outdated data appearing.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine


simon-leech:
Good point @dbauszus-glx  - pushed a fix to that !

PR #2684, opened 2026-02-25
fix(Gazetteer): Must use no response or debounced, not optional chain

simon-leech:
## Description
In this PR - the initial issue was addressed. At the end of the PR I made a mistake and converted the check to an optional chain. 
This was wrong 
The line `if(!response?.debounce)` does not do the same thing as `if(!response || response?.debounce` as we need to check for either no response, or for a debounce value of true. 

## Type of Change

## How have you tested this?
Tested locally.


dbauszus-glx:
Thanks for checking this. I read this wrong in initial PR and wasn't able to debounce with breakpoint.

PR #2692, opened 2026-03-02
Layer view and drawer cleanup

### LINKED ISSUE #2574: Separate drawer and layer-view css classes
Drawer used as a layer view item needs its own class for better readability.

This prevents style inheritance and content blending when drawers are used as hierarchical containers.

Review on drawer vs layerview stylesheets to eradicate redundant rules.

cityremade:
Introduced layout changes

* Updated default palette - new base colour introduced for content layout readability. 
* Equivalent colour added to dark mode with addition to palette documentation.
* `.no-select` class now applied on drawer headers and layer meta elements - previously each had a set of no-select rules in respective stylesheets. Redundant rules removed.
* drawer and layerview stylesheets cleaned up - redundant rules moved to lower level drawer stylesheet, improved notation and order for nested rules.
* location view - hierarchy of colours to follow the layer view logic, that is: base -> secondary -> tertiary. Previously it would follow base -> tertiary -> secondary which caused readability clashes against inputs and dropdowns.
* dropdown background now base-tertiary to contrast correctly against drawers.
* layer view stylesheet - css selectors detached from .drawer class.
* location view - aligned padding, now header has the same padding on both sides. Selectors detached from .drawer class.
* dark mode palette - colours for secondary and tertiary switched over, previously tertiary would be darker than secondary which didn't work at all times.
* drawers - expanded drawers now have a flat look with a border.
* drawers - collapsed drawers have box shadow which is a better indication that the element has inner content.
* layer group - font in primary colour for distinction.

### DIFF
--- a/lib/ui/elements/drawer.mjs
+++ b/lib/ui/elements/drawer.mjs
@@ -35,7 +35,7 @@ export function drawer(params) {
     params.drawer = mapp.utils.html.node`<div
       class="${'drawer ' + (params.class || '')}"
       data-id=${params.data_id}>
-      <div class="header">${params.header}</div>
+      <div class="header no-select">${params.header}</div>
       <div class="content">${params.content}`;
 
     return params.drawer;
@@ -68,7 +68,7 @@ export function drawer(params) {
     data-id=${params.data_id}
     class=${params.class}>
     <div
-      class="header"
+      class="header no-select"
       onclick=${onClick}>
       ${params.header}
       ${caret}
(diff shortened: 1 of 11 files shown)


dbauszus-glx (approved):
I like it. @cityremade Thanks for cleaning the css. Very good to use the no-select class on the header.

Image

simon-leech (changes requested):
I'm struggling a bit with these changes from a UI perspective. 
1. This looks really messy, why is the Dataviews and style white background but Add new locations beige and Filtering beige?

Image

2. The new group colour looks the same as a default hyperlink - this is confusing to me.
Image

dbauszus-glx:
I agree that the hyperlink colour can be confusing for the group header. But I am also a believer in using underlines for hyperlinks.

cityremade:
on hold.

PR #2700, opened 2026-03-05
Review legend methods for distributed and categorical themes

### LINKED ISSUE #2611: Distribution Count: Ability to show 0 values
In the `distribution:count` thematics, we automatically remove all legend options with a value of 0. 
We have been asked if its possible to show them, as its valid to have them in the legend, but still want the counts in brackets after them: 
e.g .
A [10]
B [5]
C [0] (currently hidden). 

This shouldn't be a default as we want them hidden by default. 
Rather I think a new flag `showZeroValues` could be added to allow this on a theme-by-theme basis.

### LINKED ISSUE #2705: Legend/filtering bug noticed on v4.20.8
Note: There was a fix to legends in v4.20.4 (https://github.com/GEOLYTIX/xyz/pull/2628https://github.com/)

Steps to recreate:
Turn on Retail Places
Toggle off a type e.g. City Centre
Apply a filter e.g. type - urban centre
Reset all filters
City Centre will appear in the legend as a category, but the polygons aren't displaying on the mapp

Link to Asana card with full details:  https://app.asana.com/1/1206078063984783/project/1209445032828999/task/1213544739765727?focus=true

dbauszus-glx:
I removed the distributedLegend method which didn't seem to do anything.

I renamed the createLegend into catElement to better explaine what this method does.

The catElement method will now show cat elements without a count if the `legend.showEmptyCat` flag is set.

image

The themeLegendSwitch method is too specific for an element method. This has been moved the legend utils as this method is only ever used in the context of a legend.

The legend utils filterAdd and filterRemove methods don't need to be exported as they are only used by the catToggle method which is exported but not used anywhere. The duplicate code catToggle method can be removed from the graduated legends module. The exported catToggle method works by setting the filterOnly flag in the graduated theme object.

Element methods should return an element to be pushed into a grid rather than relying on the grid being assigned / emptied somewhere else.

The alignContents class left was never used in the css and created a messy class assignment.

This is horrible to read:
image

I prefer a more verbose expression which is self explanatory.
image

The themeLegend method has been cleaned up to create the legend.classList property to be assigned to the theme.

The scroll property should be controlled by the nowrap class assigned with the flag. There shouldn't be an inline scoll style when the class is already assigned.

The class can also be directly assigned in the layout property.

```
"legend": {
  "showEmptyCat": true,
  "layout": "flex nowrap"
},
```


AlexanderGeere (changes requested):
If you have a category that has no elements, none of the counts show:
Image

The same instance on current patch:
Image

This works as expected, on this PR, when you add showEmptyCat to the legend:
Image

I tested this on `bugs_testing/draw/workspace.json`

dbauszus-glx:
@AlexanderGeere The issue was called by undefined entries in the grid array where the catElement method short circuits. This has been fixed by filtering the array map method for undefined entries.

dbauszus-glx:
I put this back into draft to resolve #2705 here as well.

dbauszus-glx:
An in filter set in the filterpanel will now be removed when a ni filter on the same field is set through the legend label toggle.

A ni filter set through the legend will be removed if a ni filter is set through the filter panel.

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #2706, opened 2026-03-10
Review getTemplate mod

### LINKED ISSUE #2693: String templates from src are not cached

dbauszus-glx:
The getTemplate module methods was unnecessary complex with duplicate conditions.

This PR cleans up the module structure and documentation.

Apart from modules, any template with a src will be cached in the workspace templates.

Any error condition returns a formatted new Error object.


mod/workspace/getTemplate.js
```diff
@@ -179,39 +152,3 @@ async function moduleTemplate(template, response) {
...
-@returns {Promise<Object|Error>} JSON Template
-*/
-async function cacheTemplate(workspace, template, response = {}) {
-  Object.assign(template, response);
-
-  if (template.src) {
-    workspace.templates[template.key || template.src] = template;
-
-    // file src templates should not be cached.
-    if (!template.src.startsWith('file:')) {
```
RobAndrewHurst, line 211: @dbauszus-glx If we have removed this does this mean that even templates from the file provider are also cached?

mod/workspace/getTemplate.js
```diff
@@ -31,123 +31,96 @@ import workspaceCache from './cache.js';
...
+  if (!Object.hasOwn(getFrom, method)) {
+    // Unable to determine getFrom method.
+    return new Error(`Unknown getFrom method: ${template.src}`);
+  }
 
-    response = await getFrom[method](template.src);
+  const response = await getFrom[method](template.src);
 
-    if (response instanceof Error) {
-      template.err = response;
```
RobAndrewHurst, line 96: Do we not use this in other logic to return what the actual error is for debugging purposes?

dbauszus-glx: We do. That's the reason why. In mergeTemplates module, the objTemplate method calls the getTemplate method and assigns the error returned to the template.err array. These are then printed in the addLayer method in the mapp api. 
image

mod/workspace/getTemplate.js
dbauszus-glx: Yes. All templates, but modules are cached. With a setup where some templates are local [file], some from a cdn, others from public https, maybe others from a database it would be confusing to understand why some update immediately and others don't. the logic to update the workspace by controlling the cache time should be universal. Also using the file system still has associated costs.

dbauszus-glx:
@RobAndrewHurst The template.key will now be assigned from the template provided as string to the method.

https://github.com/GEOLYTIX/xyz/pull/2706/changes/e33b5ff4324a166cd83b33cb402875e5c5990697

RobAndrewHurst:
@dbauszus-glx Nice one!

RobAndrewHurst (approved):
Nice one! I am happy for this to go in

AlexanderGeere (approved):

PR #2707, opened 2026-03-13
Add promoteDisplay method

### LINKED ISSUE #2695: Promote layer zIndex in show method
Some layers should be promoted on top of other visible layers if shown. This could be enabled with a flag for the show callbacks method.

This must not be the default behaviour since we want to avoid a basemap being drawn on top of all other layers.

dbauszus-glx:
This PR adds the promoteDisplay method to the lib/layer/decorate module.

> Called from the layer.show() method the promoteDisplay method will shortcircuit if the layer does not have the promoteDisplay flag.
> 
> Otherwise the method will iterate through the mapview.layers{} to capture the highest zIndex of any displayed layer.
> 
> The captured zIndex will be increased by 1 and set on the Openlayers layer.L to ensure that this layer is rendered on top.


jfitzpatrick17:
Does this PR introduce back-end logic only? What would this look like to the end user or does this need to be specc'd out?

dbauszus-glx:
@jfitzpatrick17 A layer shown with the promoteDisplay will be displayed on top of other layers already displayed.

simon-leech:
@dbauszus-glx from reviewing the code it looks like this only applies to currently displayed layers right? 

So if I have 4 layers 
A,B,C,D. 

I promote B above A but C and D have higher zIndexes anyway and then turned on C, B wouldn't be on top anymore? 

Also there's no way to control B's zIndex once you promote it, if B originally had a zIndex of 12 it would be removed by promoting it and the only way to return it to original would be a refresh of the tab? 

I think this needs some more scoping as to me this needs a full User Interface in which we show layers in zIndex order (highest to lowest) and let users reorder as they wish?

dbauszus-glx:
@simon-leech The zIndex property is not changed by setting the zIndex on the Openlayers layer object.

We can get the zIndex from the L object instead of the layer property.

Layer without the property flag will not be drawn on top if shown.

If you have 3 layers and want to make sure that either is drawn on top if toggled then you need to set the property on the three layers. This can not be a default flag as base layers shouldn't be drawn on top. Most polygon layers should not be drawn on top neither.

A user interface is not part of the scope for this logic. We do not want to faff around what button should sit when where for a feature that should be controlled natively by toggling the display of a layer.

dbauszus-glx:
@simon-leech I have updated the method to not rely on the layer.zIndex but use L.getZIndex() too get the current zIndex from the Openlayers object.

RobAndrewHurst (approved):
Happy with this!

AlexanderGeere (approved):

PR #2708, opened 2026-03-13
unnest featureCallback method

### LINKED ISSUE #2279: Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true.
Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true.

https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently

This warning is thrown by successive requests to render a layer when the highlighted feature changes as well as by the forEachFeatureAtPixel query itself to a lesser extend.

Image

dbauszus-glx:
This PR improves the performance and documentation of the pointerMove method in the highlight interaction module.

The featureCallback method has been unnested from the pointerMove method which is called in quick succession by the mouse move event.

The pointerMove method clears the longClick state rather than the timeout which makes the longClick unusable since it is impossible to hold down the mouse button and keep the device perfectly still. Panning is somewhat debounced by a pixel tolerance. The longClick must be reset when the map is panned.

The pointerMove identifies a candidate which is not in the stored candidateKeys set. Only the size of the set has to be compared with the length of the candidates object keys to determine if the set is the same. The areSetsEqual method can therefore be removed from the mapp utils.

The clear method does not have to be called if the highlight feature is on the same layer. This prevents the same layer to be rendered twice in quick succession. Once to remove highlight, then to add a different highlight feature to the render.

The multiple readbacks warning should only be issued if the candidates change. Moving the cursor over the same feature should not issue this warning anymore. Then again this is just a warning. There are links in the issue explaining the nature of these warnings.


AlexanderGeere (approved):

RobAndrewHurst (approved):
LGTM

PR #2710, opened 2026-03-15
feature(Location/Pin): Infoj Entry flag `locationHeader` for showing value in Location Header

### LINKED ISSUE #2711: Location View Header from infoj entry value
Clients have said that the default labels 'A', 'B' etc are generic and don't help them understand which location is selected.
It has been asked for the location header (not the pin) to have a label for the location e.g. Scotland (B) instead of just B.

simon-leech:
## Description
This PR adds a new flag to the `infoj` `entry` of `locationHeader`. 
When this flag is set, the value of that infoj entry will be used to apply to the header when the location is selected.
This helps users understand which locations are selected, rather than just using 'A' for example.
image

## Type of Change

## How have you tested this?
Tested locally both with and without the flag.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ Ran locally on my machine


dbauszus-glx:
Is there a ticket for this?

simon-leech:
> Is there a ticket for this?

@dbauszus-glx Sorry forgot to add - linked it now :)

AlexanderGeere (approved):
Works as expected

dbauszus-glx (changes requested):
This shouldn't come from the pin entry.

The pin label comes from the location.record.

It feels weird that the pin should have a different value to be assigned to the location view header.

This will become even more complex when we control where this value comes from. Where might end up with a geometry query for the pin location, and then another query for the pinLabel which is only used in the location.view.

The location.view method wants to be revised before pushing more complex logic in. In a similar fashion to the layer view. Here we made the mistake to push in more and more custom logic until we needed a complex method just to clean up the custom logic.

simon-leech:
@dbauszus-glx  As discussed the confusion was my fault due to calling the flag `pinLabel`. 

I have renamed the flag to be `locationHeader`. When this is set on an infoj entry that has a value, the value will be shown in the location view header

dbauszus-glx (approved):

RobAndrewHurst (approved):
Nice one!

PR #2718, opened 2026-03-18
Return feature.length from locationCount method.

### LINKED ISSUE #2709: Unnecessary call of the feature count method
The feature_count method should not be called from the changeEnd event if there is no material change to the layer's features.

```js
  const feature_count = await mapp.utils.xhr(
    `${layer.mapview.host}/api/query?${paramString}`,
  );
```

dbauszus-glx:
This PR unnests and documents the vector format layer setSource method.

The length of the layer.Features array should be a representation of the location count. A query everytime the viewport changes should not be required.

An MVT format layer will not have features unless a wkt_properties lookup is configured.

MVT layer must use a viewport for the locationCount query as otherwise the query would long running and potentially timeout.

### DIFF
--- a/lib/layer/format/vector.mjs
+++ b/lib/layer/format/vector.mjs
@@ -66,52 +66,8 @@ export default function vector(layer) {
 
   clusterConfig(layer);
 
-  layer.setSource = async (features) => {
-    // The layer datasource is empty.
-    if (features === null) {
-      layer.L.setSource(null);
-      mapp.layer.featureFields.reset(layer);
-      mapp.layer.featureFields.process(layer);
-      return;
-    }
-
-    if (!features) return;
-
-    layer.Features = await mapp.layer.featureFormats[layer.featureFormat](
-      layer,
-      [features].flat(),
-    );
-
-    layer.source = new ol.source.Vector({
-      features: layer.Features,
-    });
-
-    delete layer.xhr;
-
-    if (layer.fade) mapp.layer.fade(layer);
-
-    // Geojson is a cluster layer.
-    if (layer.cluster?.distance) {
-      // Create cluster source.
-
-      layer.cluster.source = new ol.source.Cluster({
-        distance: layer.cluster.distance,
-        source: layer.source,
-      });
-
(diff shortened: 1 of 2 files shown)


simon-leech (commented):
@dbauszus-glx  What are the requirements on this one ? 


I'm a bit confused as it seems to have changed default behaviour. 
At the moment the filters can search across the whole dataset. 
e.g on a point layer I can filter the whole dataset. 

Using the same layer on this PR, it seems to only apply to what's in the viewport ?

dbauszus-glx:
@simon-leech 

The requirement is to not query the location count if not changed or required. 

In this example if have a wkt point layer. The location coount will always be the same as the features on this layer which won't change unless the layer is reloaded, eg when a filter is set.

https://github.com/user-attachments/assets/38ac49d9-93b4-485a-bcdd-d250a057112a

Now imagine there are many layer witth many features and many filters. Long running queries executed at the same time could send the database connections into a tail spin once they start timing out.

The default behaviour filter is not changed. The location count returns the length of features instead of executing a query.

AlexanderGeere (commented):
@dbauszus-glx 
On patch when i use a viewport true filter i get:
Image

On this PR i get it:
Image

On the PR it starts at 6, i apply the filter then jumps to 27. 

This can be seen on `bugs_testing/filter/all_filters.json`

dbauszus-glx:
@AlexanderGeere I am unable to reproduce this issue. The dataset you describe is a WKT dataset which does not make use of the viewport.

There are 27 features on the dataset. I apply a filter, the dataset is queried and the 'filtered' features count updates to the number of features returned from the wkt query with the filter.

https://github.com/user-attachments/assets/eaa492b1-57b7-4b55-a1dd-71d2c1d46bd5

AlexanderGeere (approved):
Added a filter.viewport check to the shortcircuit.

This is working as expexted

simon-leech (approved):
Working as expected to me - compared to an application on v4.20.8 and applied the same filters with the same outputs.

PR #2722, opened 2026-03-20
Add table_schema query and checks

### LINKED ISSUE #2513: Layer [table] schema discovery
We would like to introduce a feature where a user can modify an existing theme or create a new theme.

This will require us to have a role restricted schema for a layer as defined by the infoj.entries.

Currently requests for layer data, eg. wkt, geojson, mvt do not restrict which fields are accessible.

I can send a wkt request with any fields as fields param.

```
http://localhost:3000/api/query?geom=geom_p_4326&layer=retailpoints&locale=locale&srid=4326&table=geodata.uk_glx_open_retail_points&template=wkt&fields=retailer,store_name,open_date
```

The mod/query/layerQuery method should check which fields are accessible to my user.

This needs to be tested by providing a layer json with an infoj array that includes fields which are role restricted.

A layer schema query template should be added that parses the role restricted info array and only returns the fields available to my user.

Any layer query method should return an access error if fields are requested that are not available to a user.

dbauszus-glx:
This PR adds a table_schema query to the XYZ core queries.

The smallest definition of a layer [template] is with a table and dbs. Although the workspace fallback will be used if not explicit in the template.

```json
  "templates": {
    "foo": {
      "dbs": "MAPP",
      "table": "geodata.uk_glx_geodata_oa_metrics_2021"
    }
  },
```

The templateTables method has been updated to allow for the table name to pass without the schema.

The schema and table params can now be defined for the table_schema query.

```
http://localhost:3000/api/query?template=table_schema&table_schema=geodata&table=uk_glx_geodata_oa_metrics_2021&layer=foo
```

The schema is returned as an array:
```
[
  {
    "column_name": "id",
    "data_type": "integer",
    "udt_name": "int4",
    "character_maximum_length": null,
    "column_default": null,
    "is_nullable": "NO"
  },
  {
    "column_name": "oa_code",
    "data_type": "text",
    "udt_name": "text",
    "character_maximum_length": null,
    "column_default": null,
    "is_nullable": "YES"
  },
  {
    "column_name": "lsoa_id",
    "data_type": "integer",
    "udt_name": "int4",
    "character_maximum_length": null,
    "column_default": null,
    "is_nullable": "YES"
  },
```

image

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -215,6 +215,8 @@ async function layerQuery(req, res) {
         )`;
   }
 
+  if (checkFieldsParam(req, res) instanceof Error) return;
+
   await fieldsMap(req, res);
 
   await infojMap(req, res);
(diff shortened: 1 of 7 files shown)


dbauszus-glx:
The query module structure has been revised. The executeQuery method was unnecessary complex. This method has now been resolved into the query method.

The checkFieldsParam method has been added to the layerQuery method to ensure that any field in the fields param is referenced in the layer object of a layer query.

The methods to replace variables in query template strings with string or substitute variables have now been unnested and documented.

Request params are now consistently assigned to the template object if not defined on the template `??=`.

All params, and properties should now be documented.

The dbs check should be made before compiling the query [string] in the getQueryFromTemplate method.

mod/query.js
```diff
@@ -310,6 +385,46 @@ async function fieldsMap(req, res) {
...
+@description
+The recursive method parses all properties in an object and calls itself if the property value is an object.
+
+String values of object properties with the key matching the prop argument will be added to the set argument.
+
+@param {object} obj Object to parse for property values.
+@param {string} prop The property key.
+@param {set} set The set to which the property values should be added.
+*/
+function objPropValueSet(obj, prop, set) {
```
RobAndrewHurst, line 400: This function is called `objPropValueSet` but in the JSDOC above it's `checkFieldsParam`? Does the documentation still make sense for the logic of the function?

dbauszus-glx: This is a mistake. The method was initially nested, and undocumented inside checkFieldsParam. Should be fixed now.

RobAndrewHurst (approved):

AlexanderGeere (approved):

PR #2724, opened 2026-03-25
fix(acl): Release client after everything

### LINKED ISSUE #2723: Connection leak in ACL module causes total backend lockup
## What's the Craic?

A missing `client.release()` in the error path of `mod/user/acl.js` causes database connections to leak on every failed query. After 10 failures (the default pool max), the ACL connection pool is permanently exhausted. Since the ACL pool is used for auth checks on every authenticated request, the entire backend becomes unresponsive and can only be recovered by restarting the web server.

## Hoekom?

In `mod/user/acl.js`, the `acl()` function checks out a client from the pool but only releases it on the success path:

```javascript
async function acl(q, arr) {
  try {
    const client = await pool.connect();
    const { rows } = await client.query(/* ... */);
    client.release();  // only reached on success
    return rows;
  } catch (err) {
    console.error(err);
    return err;  // client is never released
  }
}
```

If `client.query()` throws (e.g. `23502 not_null_violation`, or any other PostgreSQL error), execution jumps to the `catch` block and the client is never returned to the pool.

## Reproduction

1. Trigger any query error through the ACL path (e.g. a constraint violation, invalid input, or connection blip).
2. Repeat ~10 times.
3. All subsequent requests to authenticated endpoints will hang indefinitely.
4. Only a process restart recovers the application.

## Fix

Declare `client` before the `try` block and release it in a `finally` block to guarantee cleanup regardless of success or failure. Also add `connectionTimeoutMillis` as a safety net so requests fail fast rather than hang forever if the pool is ever exhausted.

```javascript
pool = new Pool({
  connectionString: connection[0],
  connectionTimeoutMillis: 5000,
});

async function acl(q, arr) {
  let client;
  try {
    client = await pool.connect();

    const { rows } = await client.query(
      q.replace(/acl_table/g, acl_table).replace(/acl_schema/g, acl_schema),
      arr,
    );

    return rows;
  } catch (err) {
    console.error(err);
    return err;
  } finally {
    client?.release();
  }
}
```

RobAndrewHurst:
### Summary

If client.query() throws an error the client is never released. It's permanently leaked from the pool. The ACL pool has no explicit max config, so it defaults to 10. After 10 query errors, all 10 connections are checked out and never returned. Every subsequent pool.connect() waits forever (no`connectionTimeoutMillis` set either, so it defaults to infinite).

Since the ACL is used for auth checks on every authenticated request, once the ACL pool is exhausted, every single request to the app hangs indefinitely.

This would create a lockup of unreleased clients in the backend.

### Fix

```diff
--- a/mod/user/acl.js
+++ b/mod/user/acl.js
@@ -32,18 +32,20 @@
 
       pool = new Pool({
         connectionString: connection[0],
+        connectionTimeoutMillis: 5000,
       });
 
       return acl;
     })();
 
 async function acl(q, arr) {
+  let client;
   try {
-    const client = await pool.connect();
+    client = await pool.connect();
 
     const { rows } = await client.query(
       q.replace(/acl_table/g, acl_table).replace(/acl_schema/g, acl_schema),
       arr,
     );
 
-    client.release();
-
     return rows;
   } catch (err) {
     console.error(err);
     return err;
+  } finally {
+    client?.release();
   }
 }
```

### DIFF
--- a/mod/user/acl.js
+++ b/mod/user/acl.js
@@ -34,6 +34,7 @@ export default !connection?.[1]
 
       pool = new Pool({
         connectionString: connection[0],
+        connectionTimeoutMillis: 5000,
       });
 
       return acl;
@@ -41,27 +42,28 @@ export default !connection?.[1]
 
 /**
 @function acl
- 
+
 @description
 The acl method will connect to pg pool and query the ACL with a provided query template. The `/acl_table/` and `/acl_schema/` in the query template will be replaced with values provided as `PRIVATE` or `PUBLIC` environment variable.
- 
+
 @param {string} q Query template.
 @param {array} arr Parameters to be substrituted in query template.
 */
 async function acl(q, arr) {
+  let client;
   try {
-    const client = await pool.connect();
+    client = await pool.connect();
 
     const { rows } = await client.query(
       q.replace(/acl_table/g, acl_table).replace(/acl_schema/g, acl_schema),
       arr,
     );
 
-    client.release();
-
     return rows;
   } catch (err) {
     console.error(err);
     return err;
+  } finally {
+    client?.release();
   }
 }
--- a/mod/user/acl.js
+++ b/mod/user/acl.js
@@ -62,7 +62,7 @@ async function acl(q, arr) {
     return rows;
   } catch (err) {
     console.error(err);
-    return err;
+    return new Error('Failed to connect to ACL.');
   } finally {
     client?.release();
   }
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
I changed the Error message returned which will be displayed in the login view.

image

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #2728, opened 2026-03-27
Enhancements to confirm and alert dialogs

### LINKED ISSUE #2726: HTML content for alert / confirm dialog elements
At the moment `confirm` dialog supports only custom text.
It stores content value which is html layout of the element but we can only configure text and current content parameter wraps the buttons which resolve user action.

Image
<br/>
<br/>
Would be good to take out confirm buttons to sit outside content so that I can pass some html layout when we have some more information for the user. 

<br/>

Also how about scraping that header navy both in confirm and alert? It was a good idea at the time but now our dialogs have evolved plain white.

Image

<br/>
<br/>
Here's example, custom popup I am making right now

Buttons aligned without full width, plain white header which can support an icon for readability.

Image

<br/>
<br/>

<h3>To do</h3>
 
* true / false buttons to sit outside params.content, ideally in a flexbox aligned to end
* simplify alert / confirm header - plain white.

cityremade:
Introduced changes

* alert/confirm css - removed rules for inverted background and font colour, removed header border. Header now plain.
* removed redundant h4 tag nested inside `header > h2`. Now each dialog has h2 tag only, generated by dialog method.
* buttons now outlined and inside a flexbox aligned right.
* text now aligned left, previously centered which would make it look like a poem.
* removed fixed `z-index: 1005` from confirm/alert buttons, it did nothing.
* new property `icon` - header icons now added for readability. Set `params.icon` to use any icon in the header for better readability and clear context. If unset, fallback to defaults.
* new property `innerContent` - allows adding custom html fragment instead of text only as well as separation of custom html and functional buttons.
* cancel/ok buttons now outside customisable dialog content.
* minor scaling effect on buttons in active state, it gives users response that the button has been clicked.
* updated docs.

Screenshot 2026-03-27 at 16 11 20

### DIFF
--- a/lib/ui/elements/alert.mjs
+++ b/lib/ui/elements/alert.mjs
@@ -30,7 +30,7 @@ mapp.ui.elements.alert({
 @property {string} [params.data_id='alert'] The data-id attribute value for the dialog.
 @property {string} [params.title] Text to display in the alert header. Defaults to 'Information'.
 @property {string} [params.text] Text to display in the alert content. 
-@property {string} [params.innerContent] custom HTML fragment to display in the dialog, optional. Overrides text.
+@property {HTMLElement} [params.innerContent] custom HTML fragment to display in the dialog, optional. Overrides text.
 @property {string} [params.icon] Custom Material Symbols icon name to use in the header for message clarity. Defaults to 'info'.
 @returns {HTMLElement} alert The alert element.
 */
--- a/lib/ui/elements/alert.mjs
+++ b/lib/ui/elements/alert.mjs
@@ -30,7 +30,7 @@ mapp.ui.elements.alert({
 @property {string} [params.data_id='alert'] The data-id attribute value for the dialog.
 @property {string} [params.title] Text to display in the alert header. Defaults to 'Information'.
 @property {string} [params.text] Text to display in the alert content. 
-@property {HTMLElement} [params.innerContent] custom HTML fragment to display in the dialog, optional. Overrides text.
+@property {HTMLElement} [params.innerContent] custom HTML fragment to display in the dialog, optional. Overrides `text`.
 @property {string} [params.icon] Custom Material Symbols icon name to use in the header for message clarity. Defaults to 'info'. Set empty string `icon: ''` to skip the symbol entirely.
 @returns {HTMLElement} alert The alert element.
 */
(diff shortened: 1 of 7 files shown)


dbauszus-glx (changes requested):
The idea of the alert is to be a substittute for the system alert dialog.

```js
alert('This is an alert.');
```

Image

I added a fallback check for the params being provided as string.

```js
mapp.ui.elements.alert('This is an alert.');
```

The padding looks weird in the ui alert element. Padding should be in line with header [inlduing the icon]. Here it looks like it is neither inline with icon, nor the text but just randomly defined. The padding for the header, content, footer button seems to be all over the place.

Image

It would be much easier and cleaner to use the dialog padding rather than nested paddings and margins.

Image

Are this media screen rules needed?

```css
@media only screen and (max-width: 768px) {
  dialog.alert-confirm {
    min-width: 350px;
    max-width: 70%;
  }
}
```

For the confirm dialog I would only pack into the promise what needs to be in the promise and not just return a promise from the confirm method with all the logic packed into the promise method itself.

None of these documented properties are actually used in the method itself but only in the undoocumented promise method which is returned from the confirm method.

Image

The padding issue is even more prominent when I decided to null the icon by providing an empty string.

Image

cityremade:
Changes

* support for empty string icon - this skips creation of the icon span tag,
* confirm promise is now click event based. This allows the html element being created outside of the promise,
* simplified padding - removed all inner padding rules,
* removed media queries - didn't do anything anymore,
* `no-select` class added on alert and confirm elements, redundant rules for no select removed from their stylesheet,
* margin set on basic p tag only - this is important as this prevents 'wall of text' effect.

AlexanderGeere:
@cityremade 
This is looking very nice! i just had a question. Im not sure how invested we are but i noticed on firefox the text does this:
screenshot-2026-04-08_12-10-27

This can be fixed by changing the,
screenshot-2026-04-08_12-11-27
`white-space` property to `wrap` instead of `pre`. 
Would that be alright? 

You can check this out here **`bugs_testing/drawer/workspace.json`** Theres a `alert test` button on the layer panel. 

(This does also happen on current patch, i just figured while we are here we could maybe look at this)

cityremade:
@AlexanderGeere sorted. text-wrap and white-space were in conflict.

AlexanderGeere (approved):

dbauszus-glx (approved):

RobAndrewHurst (approved):

PR #2730, opened 2026-03-30
Add keyorder to layer.view module

### LINKED ISSUE #2727: layer.view panelOrder with custom dataset.id
The panelOrder config is used to arrange the order of the content array for the layer.view.

```js
  // Create content from layer view panels and plugins
  layer.viewConfig.content = Object.keys(layer)
    .map((key) => mapp.ui.layers?.panels?.[key]?.(layer))
    .filter((panel) => typeof panel !== 'undefined');

  if (Array.isArray(layer.viewConfig.panelOrder)) {
    // Sort the content array according to the data-id in the panelOrder array.
    layer.viewConfig.content.sort((a, b) => {
      return layer.viewConfig.panelOrder.findIndex(
        (chk) => chk === a.dataset?.id,
      ) < layer.viewConfig.panelOrder.findIndex((chk) => chk === b.dataset?.id)
        ? 1
        : -1;
    });
  }
```

This doesn't seem to work with a custom dataset.id generated from a plugin panel method.

```json
{
  "locale": {
    "plugins": [
      "${PLUGINS}/v4.13.0/geojson_download.js"
    ],
    "layers": {
      "OSM": {
        "display": true,
        "format": "tiles",
        "attribution": {
        },
        "panelOrder": ["OSM-geojson-download-drawer", "meta"],
        "meta": "foo",
        "geojson_download": {
          "foo": {
            "query": "bar"
          }
        }
      }
    }
  }
}
```

The geojson_download plugin in this example creates a drawer element with the dataset.id = "OSM-geojson-download-drawer". This element should be ordered before the "meta" element.

Image

dbauszus-glx:
I believe that the panelOrder is wrong and misleading. A much cleaner approach would be to use a keyOrder array which does exctly that.

```js
    layer.viewConfig.content = Object.keys(layer)
      .sort((a, b) => {
        return layer.viewConfig.keyOrder.indexOf(a) - layer.viewConfig.keyOrder.indexOf(b);
      })
      .map((key) => mapp.ui.layers?.panels?.[key]?.(layer))
      .filter((panel) => typeof panel !== 'undefined');
```

### DIFF
--- a/lib/ui/layers/view.mjs
+++ b/lib/ui/layers/view.mjs
@@ -108,20 +108,31 @@ function viewConfig(layer) {
     ];
   }
 
-  // Create content from layer view panels and plugins
-  layer.viewConfig.content = Object.keys(layer)
-    .map((key) => mapp.ui.layers?.panels?.[key]?.(layer))
-    .filter((panel) => typeof panel !== 'undefined');
+  if (layer.viewConfig.keyOrder) {
 
-  if (Array.isArray(layer.viewConfig.panelOrder)) {
-    // Sort the content array according to the data-id in the panelOrder array.
-    layer.viewConfig.content.sort((a, b) => {
-      return layer.viewConfig.panelOrder.findIndex(
-        (chk) => chk === a.dataset?.id,
-      ) < layer.viewConfig.panelOrder.findIndex((chk) => chk === b.dataset?.id)
-        ? 1
-        : -1;
-    });
+    // Create content from layer view panel methods which may be extended through plugins.
+    layer.viewConfig.content = Object.keys(layer)
+      .sort((a, b) => {
+        return layer.viewConfig.keyOrder.indexOf(a) - layer.viewConfig.keyOrder.indexOf(b);
+      })
+      .map((key) => mapp.ui.layers?.panels?.[key]?.(layer))
+      .filter((panel) => typeof panel !== 'undefined');
+
+  } else {
+
+    // Create content from layer view panels and plugins
+    layer.viewConfig.content = Object.keys(layer)
+      .map((key) => mapp.ui.layers?.panels?.[key]?.(layer))
+      .filter((panel) => typeof panel !== 'undefined');
+
+    if (Array.isArray(layer.viewConfig.panelOrder)) {
+      // Sort the content array according to the data-id in the panelOrder array.
+      layer.viewConfig.content.sort((a, b) => {
+        const aIndex = layer.viewConfig.panelOrder.findIndex((chk) => chk === a.dataset?.id);
+        const bIndex = layer.viewConfig.panelOrder.findIndex((chk) => chk === b.dataset?.id);
+        return aIndex < bIndex ? 1 : -1;
+      });
+    }
   }
 
   // layer.classList is the legacy configuration for layer.viewConfig.classList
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
We removed the panelOrder default and warn on panelOrder config.

keyOrder should be default going forward and will always be applied first to the order of keys.

dbauszus-glx:
I keep the wrong order since legacy configs may depend on this but fixed the indexOf and typeof undefined checks.

RobAndrewHurst (approved):
LGTM

AlexanderGeere (dismissed):
Im seeing strange behaviour on **`bugs_testing/drawer/workspace.json`.** 

On patch it looks like this:
Image

On the PR it looks like this:
Image

There is no `panelOrder` or `keyOrder` in the config.

dbauszus-glx:
@AlexanderGeere Thanks for finding this. Looks like there is an issue with filtering undefined entries from the array map.

image

dbauszus-glx:
Massive face palm. 🤦🏻

Check obviously should be for `undefined` not 'undefined'.

@AlexanderGeere can you check this again.

cityremade (changes requested):
When testing on demo sites setup which has a mix of built-in and plugin generated elements I am not getting the ordering as expected:

```json
"viewConfig": {
    "panelOrder": [
      "sites-geojson-download-drawer", // this is generated by plugin `v4.13.0/geojson_download.js`
      "draw-drawer",
      "layer-toggle", // this is generated by plugin `v4.8.0/layer_toggle_button.js`
      "filter-drawer",
      "dataviews-drawer",
      "style-drawer",
      "meta"
    ]
  }
```
Visible order: 
meta
style
filtering
geojson
drawing
dataviews
layer toggle button

Image

AlexanderGeere:
> Massive face palm. 🤦🏻
> 
> Check obviously should be for `undefined` not 'undefined'.
> 
> @AlexanderGeere can you check this again.

Nice One! this has fixed it `keyOrder` is working as expected as far as I can tell. Per @cityremade change request the `panelOrder` does not seem to work as expected. I put a test example here: `bugs_testing/drawer/workspace.json`

dbauszus-glx:
@cityremade The panelOrder is wrong, but has been for a long time. Existing configurations may depend on the inverted order hence we cannot change this and introduced the keyOrder which is more reliable and in the correct order.

AlexanderGeere (approved):

simon-leech (changes requested):
Correctly see the warning for legacy panelOrder, swapped to use keyOrder fine. 

Can we make it so if you use keyOrder and not all keys are provided, any that aren't part of the keyOrder array go to the bottom by default?

dbauszus-glx:
An item not in the keyOrder array will have an index of -1.

Subtracting -1 is actually adding 1 and hence the item was ordered first.

The sort method now returns -1 if b does not have an index in the keyOrder.

@simon-leech

cityremade:
```json
"keyOrder": [
      "meta",
      "draw",
      "filter",
      "layer_toggle_button", // plugin
      "geojson_download", // plugin
      "style",
      "dataviews"
    ]
```

meta listed at first index shows up last and cluster panel not included shows up at the top.
Plugins ordered correctly.

Screenshot 2026-04-09 at 12 28 34

cityremade (approved):
working as expected ✅

AlexanderGeere (approved):

dbauszus-glx:
@simon-leech nicely done

PR #2743, opened 2026-04-07
fix(Entry/Type:Layer) -Show Panel on Show(); Hide on Hide()

### LINKED ISSUE #2742: Type:Layer Style Panel should be displayed on show and hidden on hide
Image

There is currently a UI issue whereby when a `type:layer` entry is added to the location-view. All properties of the style are immediately accessible for the user. 
This leads to confusion where the user could toggle a Display Labels checkbox and nothing happen as the overarching layer is not actually turned. 

To resolve this - we should only show the style panel when the layer is on, and hide it when its off (controlled by the checkbox).

simon-leech:
## Description
This PR tweaks the display of the style panel for a type layer entry

Initially, style panel is hidden
In show - display it
In hide - hide it

## Type of Change

## How have you tested this?
Tested against a type layer which is not display true but returns a style panel

### DIFF
--- a/lib/ui/locations/entries/layer.mjs
+++ b/lib/ui/locations/entries/layer.mjs
@@ -219,8 +219,8 @@ async function decorateLayer(entry) {
     entry.style.default = { ...entry.location?.style, ...entry.style.default };
   }
 
-  // Append style panel to layer panel, if style is available.
-  entry.style.panel && entry.panel.append(entry.style.panel);
+  // Append style panel to layer panel, if style is available and layer is already on.
+  entry.style.panel && entry.display && entry.panel.append(entry.style.panel);
 
   entry.location.removeCallbacks.push(() => {
     entry.remove();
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):

dbauszus-glx (changes requested):
This needs a little more work. At current this doesn't work with my layer entry that has a panel with an opacity slider and label key.

Image

I turn on the layer which displays but the panel doesn't.

Image

This should not be controlled by adding and removing elements from the container. If you want to control the display you should only toggle the display property.

Please note that the entry.style.panel creation is a stop gap. It should be possible to create other panel, eg meta, draw and the creation of the view should probably be managed by the layer.view module and with little duplicate code in the layer entry. We only kept this back until some requests to have a different panel then the style panel in the location entry.

simon-leech:
@dbauszus-glx I've updated this to do the following: 

Initially, style panel is hidden
In show - display it
In hide - hide it

dbauszus-glx (approved):
This is now working as expected for me.

AlexanderGeere (approved):

cityremade (approved):

PR #2744, opened 2026-04-07
Turbo Repo with Application Split

dbauszus-glx:
This is a draft PR to test express deployments to vercel.

The server.js script is a copy of the express.js file which exports the express app.

https://vercel.com/docs/frameworks/backend/express

The static routes have been removed since vercel will not use these.

My test vercel.json has the `npm i` installCommand since the devDependencies will need to be updated and a rewrite for the public files.

Public files must be included in order to be accessible through the node process which executes the express app.

```
{
  "version": 2,
  "installCommand": "npm i",
  "trailingSlash": false,
  "functions": {
    "server.js": {
      "includeFiles": "public/**"
    }
  },
  "headers": [
    {
      "source": "/(.*)",
      "headers": [
        {
          "key": "Cache-Control",
          "value": "no-cache"
        },
        {
          "key": "Access-Control-Allow-Headers",
          "value": "Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With"
        }
      ]
    }
  ],
  "rewrites": [
    {
      "source": "/public/(.*)",
      "destination": "/$1"
    }
  ]
}
```


dbauszus-glx:
With the xyz app moved into the apps directory we can add a build for just this app.

Here it is important that the workspace in the root public directory is referenced with `../..` to step out of apps/xyz.

```
{
  "version": 2,
  "trailingSlash": false,
  "builds": [
    {
      "src": "/apps/xyz/server.js",
      "use": "@vercel/node",
      "config": {
        "includeFiles": "public/**"
      }
    }
  ],
  "routes": [
    {
      "src": "/(.*)",
      "dest": "/apps/xyz/server.js"
    }
  ],
  "env": {
    "WORKSPACE": "file:../../public/workspace.json",
    "TITLE": "XYZ-SERVER"
  }
}
```

I added the mapp app to the vercelignore as this app does not have to built, nor deployed. The built mapp library is referenced in the public folder.

image

.vercelignore
```diff
@@ -1,12 +1,12 @@
...
 
-env
 package-lock.json
 pnpm-lock.yaml
 **/node_modules
 
 *.md
-express.js
+
+server.js
```
RobAndrewHurst: Don't we want this if we are using express in vercel?

dbauszus-glx: I removed this from the vercelignore as the root server script is required.

simon-leech:
@RobAndrewHurst possibly silly question but what / how do I test this PR? 😅 
Adding tests or running the app or what?

dbauszus-glx:
@simon-leech Don't worry about testing this for now. There are still quite a few things that need to be resolved, including documentation in regards to the dependency resolution.

dbauszus-glx:
@RobAndrewHurst I want to keep this in draft until I am happy with the dependency resolution on my test vercel deployments. Thereafter I want to revise the markdown readme[s] in this regard.

RobAndrewHurst:
@dbauszus-glx I need this in ASAP for the mapp-next stuff. So the documentation and stuff will only come later on in the major. We need to start getting this in some real testing

RobAndrewHurst:
> @RobAndrewHurst I want to keep this in draft until I am happy with the dependency resolution on my test vercel deployments. Thereafter I want to revise the markdown readme[s] in this regard.

This is going to take too long

dbauszus-glx:
@RobAndrewHurst In that case we can merge it in to start patching. Not much difference between rushed or no testing.

PR #2746, opened 2026-04-07
fix(distributed_theme.legend): Assign `filterOnly` to filter only on the map

### LINKED ISSUE #2745: Type:Distributed - Toggling individual options in the theme disables all values
When you use a type distributed theme and toggle one of the options in the theme, it is correctly hidden in the map. 
However, the whole legend is toggled to disabled which is wrong.

AlexanderGeere:
## Description
Distributed theme values will never match what is in the `filter.current` so in this PR we set the filterOnly flag true for all distributed themes. This change will make it so the filter is only visual and does not interact with the `filter.current` of the layer:
screenshot-2026-04-08_11-39-16

The disabled flag for the category selected is also nullish checked on the filter as the the `filterAdd` function sets the flag so the property should not be overwritten from the filter. 

## GitHub Issue
#2745

## Type of Change

## How have you tested this?

> [!IMPORTANT]
> Tested locally on `bugs_testing/styles/distributed_themes.json`

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/legends/distributed.mjs
+++ b/lib/ui/layers/legends/distributed.mjs
@@ -20,6 +20,7 @@ The distributedTheme method creates and returns a distributed theme legend for t
 @returns {HTMLElement} The distributed theme legend element.
 */
 export default function distributedTheme(layer) {
+  //Apply filterOnly so the filtering action on the theme is only visual.
   layer.style.theme.filterOnly = true;
   const theme = layer.style.theme;
 
--- a/lib/ui/layers/legends/distributed.mjs
+++ b/lib/ui/layers/legends/distributed.mjs
@@ -20,6 +20,7 @@ The distributedTheme method creates and returns a distributed theme legend for t
 @returns {HTMLElement} The distributed theme legend element.
 */
 export default function distributedTheme(layer) {
+  layer.style.theme.filterOnly = true;
   const theme = layer.style.theme;
 
   themeLegend(theme);
--- a/lib/ui/layers/legends/utils.mjs
+++ b/lib/ui/layers/legends/utils.mjs
@@ -29,7 +29,7 @@ export function catElement(cat, theme, layer) {
 
   // Check whether cat is in current filter.
   cat.disabled = layer.filter?.current[cat.field]?.ni?.some(
-    (el) => el === cat.value || cat.values.includes(el),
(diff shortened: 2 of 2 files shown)


simon-leech (approved):
Working for me - thank you! 👍

dbauszus-glx (commented):
Can this work to push an array [cat.values] into an array?

```
layer.filter.current[cat.field].ni.push(cat.keys || cat.value || cat.values);
```

dbauszus-glx (dismissed):
The distributed theme legend was not meant to use filter. The assignment is pseudo random. Dependent on which feature is met first. Any filter should only be applied to the style but not create a NI filter. 

Let;s say you have two categories which groups the features that you see in the mapview. You set a filter on one category and then change mapview. Now more features are loaded and added to the category groups. You now have a situation in which a category is disabled with some features added to a filter but other features are shown because they were added later.

dbauszus-glx:
This fails as I expected.

The retail points layer has a viewport with just two categories at the beginning. I set the filter on Tesco then zoom out other values will be added to the first cat entry but never to the filter. You end up with a layer filter that is different to the display.

https://github.com/user-attachments/assets/ce32867b-ddad-4ac7-b617-7f0dde663236

A distributed theme filter must not be a layer filter but only visual, ie set the style to null.

AlexanderGeere:
> Can this work to push an array [cat.values] into an array?
> 
> ```
> layer.filter.current[cat.field].ni.push(cat.keys || cat.value || cat.values);
> ```

Yes a couple lines down the array gets flattened, But i will change this anyway so the filter is only visual as per your change request

AlexanderGeere:
@dbauszus-glx I set the `filterOnly` property to true on all distributed themes, let me know how you feel about that. I think its best as then the null styling gets applied through existing logic. 

I also nullish check the filter as the `filterAdd` function sets the disabled flag and `filterOnly` themes will obviously not be in the current filter.

simon-leech (approved):
Happy with this - still working for me 🥇

dbauszus-glx:
Does this need a nullish assignment?
```
cat.disabled ??= layer.filter?.current[cat.field]?.ni?.includes(cat.value);
```

dbauszus-glx (approved):
working for me as expected.

PR #2748, opened 2026-04-09
fix(Location/dataview entry): Shortcircuit on entry.show existing already

### LINKED ISSUE #2747: Dataview entry setData method called thrice with dynamic flag on update
A [tabulator] dataview entry with the dynamic flag will be created twice and the setData method called three times. This works but is inefficient and will cause an error exception in the case of a tabulator dataview which must be awaited before the data can be set.

From the updateCallback method of the location update method, the infoj is processed and the mapp.ui.Dataview(entry) [58] method is called on an existing dataview entry. This will call the create method of the entry to re-create and subsequently set the data on the newly created dataview. But before the data can be set the create method will be called again from the entry.show method [69].

Image

Subsequently the setData method is called three times and errs because the Tabulator has already been recreated by the time the setData method is called first.

The first two setData methods are triggered by the create methods. The third time it is triggered by the dynamic flag update after the create.

Image

We would expect a dataview to only be created once, and the data to be set once when a location is updated.

simon-leech:
## Description

This PR fixes an issue where the dataview component is recreated multiple times during dynamic location updates. 
It introduces a short-circuit check using `!entry.show` to verify if the dataview has already been initialised before attempting to decorate it again. 
This ensures it is only created once, and prevents redundant setData calls and exceptions.

``` js
// BEFORE 
// Decorate the dataview entry.
if (mapp.ui.Dataview(entry) instanceof Error) return;

// AFTER 
// Decorate the dataview entry, only if the show method does not yet exist.
// Once decorated, the show method exists so we don't need to redecorate on subsequent dataview entry updates.
if (!entry.show && mapp.ui.Dataview(entry) instanceof Error) return;
```

A check in Dataview.mjs has also been added to check for entry.show too. 
## Type of Change

## How have you tested this?
Tested locally and off the PR the warnings and errors fire, on this PR they no longer do.

## Testing Checklist
- ✅ Ran locally on my machine

### DIFF
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -154,7 +154,7 @@ function show() {
 
     this.update instanceof Function && this.update();
   } else if (this.dynamic) {
-    this.create();
+    typeof this.show !== 'function' && this.create();
 
     this.update instanceof Function && this.update();
   } else if (this.reload) {
--- a/lib/ui/locations/entries/dataview.mjs
+++ b/lib/ui/locations/entries/dataview.mjs
@@ -54,10 +54,11 @@ export default function dataview(entry) {
   // A dataview must have a HTMLElement target
   if (entry.target instanceof HTMLElement === false) return;
 
-  // Decorate the dataview entry.
-  if (mapp.ui.Dataview(entry) instanceof Error) return;
+  // Decorate the dataview entry, only if the show method does not yet exist.
+  // Once decorated, the show method exists so we don't need to redecorate on subsequent dataview entry updates.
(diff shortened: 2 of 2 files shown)


AlexanderGeere:
@simon-leech I added another show check in `Dataview.mjs` to conditionally call create if show already exists and now the setData is only called once on update.

dbauszus-glx (approved):
@simon-leech nicely done

dbauszus-glx (approved):

cityremade (approved):

RobAndrewHurst (approved):

PR #2753, opened 2026-04-13
Revise iteration of layer.style.elements and check for labels and hovers

### LINKED ISSUE #2752: TypeError: Cannot read properties of undefined (reading 'minZoom')
This issue is caused where we have setLabel definition but the label is not defined in the labels.

Image

The label will then be undefined and the changeEnd method will check for the minZoom of the undefined label.

### LINKED ISSUE #2758: Label - Zoom Restricted Label Checkbox is mistakenly enabled when swapping theme
A layer with a single `label` checkbox that has a `minZoom` flag will be mistakenly enabled when the theme is changed. 

Configuration for the label for example
``` json
 "label": {
      "display": false,
      "field": "store_name",
      "declutter": true,
      "strokeColor": "#ffffff",
      "strokeWidth": 2,
      "minZoom": 11
    },
```

When you swap between themes, the checkbox is enabled but clicking it obviously won't do anything.

dbauszus-glx:
The setHover and setLabels values need to be checked against the properties in the style.labels or style.hovers object. If the object does not have the own property the object will be set to undefined which cause a typeerror when the changemethod checks for the min or maxZoom properties of undefined.

The filter/filter/map/flat/filter iteration was impossible to debug and is not very performant since under the hood js will create nested generator methods to yield results.

### DIFF
--- a/lib/ui/elements/layerStyle.mjs
+++ b/lib/ui/elements/layerStyle.mjs
@@ -180,13 +180,21 @@ function label(layer) {
 
   if (layer.style.label.hidden) return;
 
+  if (layer.style.labelCheckbox) {
+    return layer.style.labelCheckbox;
+  }
+
+  // The title will only be shown in dropdown for multiple labels option.
+  let label = layer.style.labels && mapp.dictionary.layer_style_display_labels;
+
+  // Assign label from single label object title.
+  label ??=
+    layer.style.label.title || mapp.dictionary.layer_style_display_labels;
+
   layer.style.labelCheckbox = mapp.ui.elements.chkbox({
     checked: !!layer.style.label.display,
     data_id: `labelCheckbox-${layer.key}`,
-    label:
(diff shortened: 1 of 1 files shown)


cityremade (approved):

simon-leech (approved):
This is working as expected but I did find another style bug related to `minZoom` on a label that I wrote on this ticket as assuming we want to address that in a separate ticket- 
https://github.com/GEOLYTIX/xyz/issues/2758

dbauszus-glx:
I addressed the issues in regards to label zoom checks with multiple label and themes.

The checkbox should be returned if it already exists. The changeEndCallback for the zoomCheck should not be pushed multiple times.

layer.reload does not trigger changeEndCallbacks, the zoomLabel method should be called after the reload.

simon-leech (approved):
This is working for me and the minZoom issue resolved too

RobAndrewHurst (approved):

PR #2755, opened 2026-04-13
fix(Entry/Dataview): Check on `entry.display` and `entry.create`

simon-leech:
The previous fix meant that dvs with a target of location were not being rendered into the location view. 
In this PR we: 
- Check on if entry.display does not exist but create does. 
- This means the dv is already decorated, so just return the checkbox to be used by the user.

Added bugs_testing\dataviews\dv_all_options.json - which has all 16 combinations of table / chart and the flags - if we test against this we should cover all use cases 🤞
image

### DIFF
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -154,7 +154,9 @@ async function show() {
 
     this.update instanceof Function && this.update();
   } else if (this.dynamic) {
+    // The create method will set existing data but will not query data.
     this.create instanceof Function && (await this.create());
+    // The update method will query data.
     this.update instanceof Function && this.update();
   } else if (this.reload) {
     this.update instanceof Function && this.update();
--- a/lib/ui/Dataview.mjs
+++ b/lib/ui/Dataview.mjs
@@ -154,7 +154,7 @@ async function show() {
 
     this.update instanceof Function && this.update();
   } else if (this.dynamic) {
-    this.create instanceof Function && await this.create();
+    this.create instanceof Function && (await this.create());
     this.update instanceof Function && this.update();
   } else if (this.reload) {
     this.update instanceof Function && this.update();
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
The setData is called twice like this.

First from the decorator, and then from the show method.

Image

@simon-leech Let me have a look at a solution for this while I have this open.

dbauszus-glx:
I think there was some confusion between show and display properties.

The show method will be assigned by the decorator hence there should never be the condition where a dataview doesn't have a show method but has a create method.

The dataview entry method should return the checkbox if the dataview has been decorated but should not be displayed.

The dataview only needs to be decorated if it hasn't yet been decorated ie `!entry.create`.

```js
  // The dataview has been decorated but should not be displayed.
  if (!entry.display && entry.create instanceof Function) {
    return mapp.utils.html.node`
    ${entry.chkbox || ''}
    ${entry.locationViewTarget || ''}`;
  }

  // The dataview doesn't have a create method and needs to be decorated.
  if (!entry.create && mapp.ui.Dataview(entry) instanceof Error) return;
```

The show method should be async.

The creation which which will set existing data must be awaited for dynamic dataviews before calling the update method which will query data.

```js
async function show() {
  this.display = true;

  if (this.create === undefined) {
    this.create = () => {
      mapp.ui.utils[this.dataview].create(this);
    };
    this.create();

    this.update instanceof Function && this.update();
  } else if (this.dynamic) {
    this.create instanceof Function && (await this.create());
    this.update instanceof Function && this.update();
  } else if (this.reload) {
    this.update instanceof Function && this.update();
  }
  //...
}
```

dbauszus-glx:
The are an infinite possible dataviews and targets hence the logic test matrix must be indepedent on the target and type of dataview.

We need to test a dataview entry with dynamic and reload properties. The dataview will be re-created with dynamic flag but only updated with reload flag.

Dataview with display on and off. The dataview will be decorated but not shown with display off.

On update a dataview should not be decorated again and only only be shown if display is true.

The create and update logic is nested in the dataview show method.

simon-leech:
This looks good to me now too - thank you @dbauszus-glx !

dbauszus-glx (approved):

AlexanderGeere (approved):

PR #2756, opened 2026-04-14
Show disabled cat in legend

### LINKED ISSUE #2751: Disabled categories not shown in theme
The recent changes in this PR https://github.com/GEOLYTIX/xyz/pull/2700/changes - have broke a lot of the functionality of styling. 

**Issues**
1. A categorized theme with distribution count must be off unless the layer is on. It should be hidden when the layer display is toggled. 
This is as not all the theme cats may be present in the data so we only show the right ones.
2. Clicking on a cat in theme should disable it, not remove it from the legend entirely. 
It is currently impossible to re-instate that cat without refreshing the tab.

### LINKED ISSUE #2754: categorized theme - TypeError: Cannot read properties of null (reading 'icon')
Setting the filterOnly flag on a theme will set the style to null to prevent a feature from being styled without altering the layer filter object.

This will throw an error exception where the categorized theme method looks for an icon in the cat.style object.

Image

dbauszus-glx:
A cat which has been disabled by user input should be shown in the legend. Otherwise it will be impossible to enable the cat by user input.

https://github.com/user-attachments/assets/739d0504-3962-4bb3-a70b-345a386b9114

There must be a conditional check on the style.icon to be returned from the categorized theme method. Otherwise an error will be thrown where the style is set to null.

### DIFF
--- a/lib/layer/themes/categorized.mjs
+++ b/lib/layer/themes/categorized.mjs
@@ -53,7 +53,7 @@ export default function categorized(theme, feature) {
 
         flat ||= Array.isArray(cat.style.icon);
 
-        return cat.style.icon;
+        return cat.style?.icon;
 
         // Filter out empty icon entries from map response.
       })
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):

simon-leech (approved):
This looks good to me thank you 👍

PR #2761, opened 2026-04-15
Use flag to control hover query cache

### LINKED ISSUE #2757: Hover - debounce and cache meaning data is not refreshed when fields updated?
If I have a `hover.query` assigned that returns features from the location, and I update the locations information, I would expect the hover to show the latest data the next time I hover over that feature. 

If i update the site_name and then hover over the feature again, the site_name should be updated.
``` sql
SELECT 
    concat('<div><b>', site_id, '</b> ',site_name, '</div>')
    AS label
    FROM ${table}
    WHERE ${qID} = %{id}
```

I think this is due to the debounce / cache but not quite sure.

dbauszus-glx:
A flag should be used to control whether a hover query should be cached.

The query will not be cached with the dynamic flag.

```
          "hover": {
            "query": "hover_query",
            "display": true,
            "dynamic": true
          }
```

### DIFF
--- a/lib/layer/featureHover.mjs
+++ b/lib/layer/featureHover.mjs
@@ -69,7 +69,7 @@ export default function featureHover(feature, layer) {
   mapp.utils
     .xhr({
       url: `${layer.mapview.host}/api/query?${paramString}`,
-      cache: true,
+      cache: !layer.style.hover.dynamic,
       debounce: { key: 'hover', delay: 100 },
     })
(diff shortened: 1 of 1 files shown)


cityremade (approved):

simon-leech (approved):
Working well for me - tested changing data that's in the hover and dynamic doing what is expected

AlexanderGeere (approved):

PR #2766, opened 2026-04-16
Merge direct layer role payloads into resolved workspace objects

RobAndrewHurst:
## What happened? 

This change fixes a gap in workspace object composition where a `roles` object defined directly on a layer could control access, but its matching role payload was not merged into the final resolved layer.

Previously:

- direct layer `roles` were preserved during server-side composition
- `Roles.check()` could allow or deny access based on those roles
- but the matching role payload was never merged into the resolved object
- then `removeRoles()` stripped the raw `roles` object from the API response

That meant users with a matching role would not receive the expected role-specific layer overrides, such as `name`, `filter.current`, or other mergeable properties.

This PR applies the direct object role payload before the workspace API strips `roles` from the response.

## Problem

Given a layer like:

```json
{
  "name": "Base name",
  "roles": {
    "KEY_BRANDS": {
      "name": "KEY BRANDS",
      "filter": {
        "current": {
          "key_brands_layer": {
            "in": [true]
          }
        }
      }
    }
  }
}
```

and a user with role `KEY_BRANDS`, the expected resolved layer should include:

```json
{
  "name": "KEY BRANDS",
  "filter": {
    "current": {
      "key_brands_layer": {
        "in": [true]
      }
    }
  }
}
```

Before this change, access could succeed, but the role payload was not merged, and the client received neither the merged values nor the raw `roles` object.

## Root Cause

`mergeTemplates()` already applied `Roles.objMerge()` to template objects during template preparation, but it did not apply the same behavior to the final resolved object itself.

A first pass at fixing this by calling `Roles.objMerge(obj, roles)` at the end caused a regression: template role payloads were merged twice, which duplicated merged array content like `infoj`.

## Solution

The final merge step now applies only the object's original direct `roles` payload, rather than re-merging all accumulated role data after template composition.

This preserves existing template behavior while enabling direct layer-level role payloads to be merged into the resolved object before response sanitization.

## Changes

- update `mod/workspace/mergeTemplates.js`
- capture the object's original direct `roles`
- after template processing and access checks, merge only those direct role payloads into the resolved object
- avoid reapplying template role payloads that were already merged earlier

## Tests

Added regression coverage for direct object role merging in:

- `tests/mod/workspace/mergeTemplates.test.mjs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**v4.22.4**
+**v4.22.5**
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 ![Codi Unit Tests](https://github.com/GEOLYTIX/xyz/actions/workflows/unit_tests.yml/badge.svg)
--- a/lib/mapp.mjs
+++ b/lib/mapp.mjs
@@ -52,5 +52,5 @@ globalThis.mapp = {
   Mapview,
   plugins,
   utils,
-  version: '4.22.4',
+  version: '4.22.5',
 };
--- a/mod/workspace/mergeTemplates.js
+++ b/mod/workspace/mergeTemplates.js
@@ -50,7 +50,7 @@ export default async function mergeTemplates(obj, roles) {
 
   obj.roles ??= {};
   //const directRoles = structuredClone(obj.roles);
-  
+
   if (obj.role) {
     obj.roles[obj.role] ??= true;
   }
--- a/mod/workspace/mergeTemplates.js
+++ b/mod/workspace/mergeTemplates.js
@@ -87,7 +87,10 @@ export default async function mergeTemplates(obj, roles) {
   // Apply only the original object's direct role payloads before the workspace
   // API strips roles from the response. Template role payloads have already
   // been merged during template preparation.
-  obj = merge(obj, Roles.objMerge({ roles: directRoles }, roles));
+  if (Object.keys(directRoles).length) {
+    const rolesObj = Roles.objMerge({ roles: directRoles }, roles);
+    obj = merge(obj, rolesObj);
+  }
 
   return obj;
 }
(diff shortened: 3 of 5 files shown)


dbauszus-glx (changes requested):
I think the problem is as simple as follows. When the template merge logic in regards to the obj.role and obj.roles was revised we moved the Roles.objMerge into the prepareTemplate method. This means that the merge was not applied to objects without or not in a template.

I don't fully understand why it is necessary to clone the directRoles object prior to template merges.

I added a condition to prevent merging an empty object into itself on every [template] object.

Setting a breakpoint I can see obj has the roles object.

Image

Wouldn't it work to just use:
```
obj = Roles.objMerge(obj, roles);
```

This could be done at the beginning of the mergeTemplates method instead of the structuredClone, or at the end of the process in case a template may add or alter the roles object.

@simon-leech It is important to understand that the roles object is legacy and will not work with nested roles.

RobAndrewHurst:
> I think the problem is as simple as follows. When the template merge logic in regards to the obj.role and obj.roles was revised we moved the Roles.objMerge into the prepareTemplate method. This means that the merge was not applied to objects without or not in a template.
> 
> I don't fully understand why it is necessary to clone the directRoles object prior to template merges.
> 
> I added a condition to prevent merging an empty object into itself on every [template] object.
> 
> Setting a breakpoint I can see obj has the roles object.
> 
> Image
> Wouldn't it work to just use:
> 
> ```
> obj = Roles.objMerge(obj, roles);
> ```
> 
> This could be done at the beginning of the mergeTemplates method instead of the structuredClone, or at the end of the process in case a template may add or alter the roles object.
> 
> @simon-leech It is important to understand that the roles object is legacy and will not work with nested roles.

I did try this, but I was getting duplicate entries in the tests

dbauszus-glx:
@RobAndrewHurst Can you test my commit where I comment the directRoles and do the Roles.objMerge prior to assign an empty roles object. Does this create duplicate entries?

I am running the provided tests which all pass. We need to create an additional test for the duplicate entries if you have a test case for this.

RobAndrewHurst:
@dbauszus-glx Nice one! Happy with this!

@simon-leech Can you please review this asap

simon-leech (approved):

AlexanderGeere (approved):

dbauszus-glx (approved):

PR #2775, opened 2026-04-21
use relative path in provider file module.

### LINKED ISSUE #2774: monorepo apps file paths
The provider/file method determines a relative file path from the position of the file module itself.

A relative path for a resource outside of the cloned repository can be provided by stepping out of the repository by prefixing the path with `../`, eg. "file:../xyz_resources".

This works since the path itself will be prefixed with `../../` when joined with the relative location of the provider module taken from the import.meta.

Image

With restructure of apps in a monorepo the relative location of the provider module within the repository has changed. The module is nested two steps deeper `apps/xyz`.

In order to maintain a local launch configuration in which the apps/xyz/server.js script is executed would require the path to be prefixed with `../../../../` which seems excessive.

A cleaner approach would be to use the current working directory of the process through the return from process.cwd().

This will require us to remove the prefix in the templates/_views.js module.

dbauszus-glx:
This PR ensures that current launch configurations with relative path definitions remain valid.

### DIFF
--- a/apps/xyz/mod/provider/file.js
+++ b/apps/xyz/mod/provider/file.js
@@ -38,13 +38,15 @@ export default async function file(ref) {
       throw new Error('Unauthorized');
 
     // Subtitutes {*} with xyzEnv.SRC_* key values.
-    const path = (ref.params?.url || ref).replace(
+    const relativePath = (ref.params?.url || ref).replace(
       /{(?!{)(.*?)}/g,
       (matched) => xyzEnv[`SRC_${matched.replace(/(^{)|(}$)/g, '')}`],
     );
 
+    const path = join(process.cwd(), relativePath)
+
     // Join the releative path with the import directory name to generate a path with platform specific separator as delimiter.
-    const file = readFileSync(join(import.meta.dirname, `../../${path}`));
+    const file = readFileSync(path);
 
     if (path.match(/\.json$/i)) {
       return JSON.parse(file, 'utf8');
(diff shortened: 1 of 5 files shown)


dbauszus-glx:
@RobAndrewHurst I put this in draft for now as I need to understand the relative paths in the failing tests.

image

RobAndrewHurst:
@dbauszus-glx Fixed. I removed the mocks didn't need to be in the file.test.mjs and fixed the view test

dbauszus-glx:
@RobAndrewHurst I still have 35 tests / 7 files failing.

image

RobAndrewHurst:
@dbauszus-glx 
image
They should all be passing. They pass in the actions and for me locally

dbauszus-glx:
@RobAndrewHurst The tests are still failing for me. There may be a different reason than file paths.

image

I feel the test expecting the wrong response.

image

dbauszus-glx:
@RobAndrewHurst Ignore previous comment.

This fails on file path. Looking for tests folder on root.
image

I am concerned that the template is an error instead of having an err property.
image

dbauszus-glx:
@RobAndrewHurst The vscode extension must have a different current working directory.

I do not find the files in the tests directory unless I specify the location within /apps/xyz
```
WORKSPACE: 'file:/apps/xyz/tests/assets/query_workspace.json',
```

RobAndrewHurst (approved):

PR #2779, opened 2026-04-22
Cloudfront fetch must return err on failure to fetch

### LINKED ISSUE #2778: cloudfront fetch fail
The cloudfront provider module will sign a url and fetch a ressource from the signed url to be returned from the module.

The signing process will not err if the key provided in the as KEY_CLOUDFRONT in the environment variables is valid and present in the root.

Fetch will not fail if the file / path is wrong.

Fetch will fail if the domain is wrong.

dbauszus-glx:
Rather than returning the async Fetch method from the cloudfront provide module method the fetch should be resolved inside the method.

The method must be tried to account for the fetch method failing.

The fetch response will be 403 if the signed url path is wrong or the file doesn't exist eg. `cloudfront:dev.geolytix.io/bogus.json`

The fetch method will crash if the domain is bogus eg. `cloudfront:bogus.io/file.json`

Resource parsing will crash if the resource cannot be parsed as assumed. eg. `cloudfront:dev.geolytix.io/text.json`

### DIFF
--- a/mod/provider/cloudfront.js
+++ b/mod/provider/cloudfront.js
@@ -36,9 +36,9 @@ The fetched will resource body will be parsed as JSON if the URL ends in `.json`
 export default cloudfront_signer ? cloudfront : null;
 
 async function cloudfront(req) {
-  const url = req.params?.url || req;
-
   try {
+    const url = req.params?.url || req;
+
     const signedURL = await cloudfront_signer(url);
 
     if (signedURL instanceof Error) {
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):
tested with a broken, not found and found urls. This PR prevents the process from crashing as expected.

cityremade (approved):

RobAndrewHurst (approved):

PR #2780, opened 2026-04-22
Add drawer `disableOnHide` flag to reference layer display toggle.

### LINKED ISSUE #2760: Theme and layer display
The display of legend elements may be dependent on the presence of features.

In the case of the distributed theme legend it will not be displayed until the layer is displayed. This looks odd since it is possible have only a legend in the style drawer which would be empty in this case.

Image

The legend will show once the layer display is toggled and will remain displayed after the layer display is toggled off.

Image

This behaviour is odd and deviates from the way other legends are displayed.

Neither should behaviour deviate for other layer view panels. The most accessible approach would be to collapse the layer view when the display is toggled off. This behavior could be flagged as optional in the layer.viewConfig

dbauszus-glx:
This PR introduces the 'disableOnHide' value for the drawer property.

The drawer property will is checked by the ui/elements/drawer method before the drawer element is assigned to the same prooperty.

Setting a string value eg. 'disableOnHide' will allow to add show and hide callback methods to the layer object.

This can be used to control whether a drawer eg style will be disabled and collapsed when the layer is hidden.

The complex popout logic should be in a seperate method to shortcircuit with the popout propertty falsy rather than the whole logic stuffed into an if condition block.

The styleParser has been updated to apply this new flag to all style objects, providing a consistent UI pattern where the style is collapsed and disabled when a layer is off.

### DIFF
--- a/lib/ui/elements/drawer.mjs
+++ b/lib/ui/elements/drawer.mjs
@@ -63,6 +63,17 @@ export function drawer(params) {
     : mapp.utils.html`
       <div class="notranslate material-symbols-outlined caret">`;
 
+  if (params.drawer === 'disable' && params.layer?.showCallbacks) {
+    params.class += params.layer.display ? '' : ' disabled';
+    params.layer.showCallbacks.push(() => {
+      params.drawer.classList.remove('disabled');
+    });
+    params.layer.hideCallbacks.push(() => {
+      params.drawer.classList.add('disabled');
+      params.drawer.classList.remove('expanded');
+    });
+  }
+
   params.drawer = mapp.utils.html.node`
   <div
     data-id=${params.data_id}
--- a/lib/ui/elements/drawer.mjs
+++ b/lib/ui/elements/drawer.mjs
@@ -16,7 +16,11 @@ The drawer method will create and return drawer element with a header and conten
 @property {string} params.data_id The data-id for drawer element.
 @property {HTML} params.header The header element[s].
 @property {HTML} params.content The content element[s] for the drawer.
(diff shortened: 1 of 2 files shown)


lib/ui/elements/drawer.mjs
```diff
@@ -63,6 +61,19 @@ export function drawer(params) {
...
+    params.layer.showCallbacks.push(() => {
+      params.drawer.classList.remove('disabled');
+    });
+    params.layer.hideCallbacks.push(() => {
+      params.drawer.classList.add('disabled');
+      params.drawer.classList.remove('expanded');
+    });
+  }
+
+  popout(params);
```
RobAndrewHurst: Why do we popout the drawer before we create it on line 77?

dbauszus-glx: I am not sure. I just moved the conditional if block into a function. Not sure about the origin of the block. @AlexanderGeere

AlexanderGeere: It doesnt work this way around, I switched it in my last commit. [ae12855](https://github.com/GEOLYTIX/xyz/pull/2780/commits/ae1285568fc03c93f438bee2d482a7ff83917e7c)

AlexanderGeere (approved):
I updated the documentation for the new disable option.

Working as expected otherwise

simon-leech (approved):
I'm happy with this! 

i changed the flag to `disableOnHide` so its clearer to configuration what it means. 
I also set this flag as the default in `styleParser` for all styles as I think this is better to have a consistent UI pattern. It can be overwritten passing a different string to cancel it out. 
I would rather this as its the same everywhere and UAT on v4.23.0 will include this new pattern across all layers.

PR #2783, opened 2026-04-22
Update gitignore to include all public files including mapp modules

### LINKED ISSUE #2782: Include library files in repository; Review .gitignore
v5 allows to deploy the xyz-server without the mapp library files.

The built library files should therefore be included in the repository.

dbauszus-glx:
The mapp files do not require to be built and can be included with other public files in the repository.


dbauszus-glx:
@RobAndrewHurst I am somewhat stumped by the json gitignore rules. I feel these are legacy since we block all json but then override any json I can think about in the repository.

```
/*.json
/*.jsonc
!tests/*.json
!tests/**
!package.json
!jsdoc*.json
!codi.json
!biome.json
!turbo.json
```

dbauszus-glx:
@RobAndrewHurst Can we disable code analysis for the built library files?

dbauszus-glx:
Removing the json rules did not have any effect on the files included in the repository.

I added the public folder to the sonarcloud exclusion and await a re-run of the scan.

dbauszus-glx:
The esbuild caching seems to be very aggressive. I noticed that changes in the lib files are not always considered and files are not recreated when deleted.

I delete the public/js/lib/mapp.js file and run build the deleted file is cached, no output.

I make a change to an input file and run build. The changed mapp file is written out and cached.

I revert the change and built. The file exist in cache and will not be updated.

RobAndrewHurst (approved):

PR #2784, opened 2026-04-22
v5 express middleware

### LINKED ISSUE #2777: v5 express middleware
The XYZ repository will be restructured as a monorepo with the v5 major release. This will allow for better dependency management as well as express middleware.

The api/api.js module which could be deployed as a Vercell serverless function did not make use of express routing and middleware.

The module was ported to /apps/xyz/api.js but with little change apart from the removal of the saml authentication into a seperate monorepo app.

The redirect for DIR paths set in the environment variables and validation of request parameters probably want to be regular express middleware to simplify the api module logic.

Image

Likewise does the validateRequestAuth method want to be middleware. This middleware should be able to validate token issued by SAML or other custom auth layer and handle the ACL lookup. Custom security layer should not require access to the ACL themselves.

The calling the requestRouter method at the end of the validateRequestAuth method is awkward.

Routing should be handled by the express router. Request validation and authorization should be handled in Express middleware.

dbauszus-glx:
Use middleware to validate request params and authentication.

The api module is no longer required since the express app can route to individual endpoint methods after requests pass through the middleware.


dbauszus-glx:
@RobAndrewHurst I addressed the sonarqube warning in regards to the array some/includes check for the user param only to realise that the check is bogus anyways since the param is already deleted a few lines earlier.

image

dbauszus-glx:
I am currently stuck with redirects to these type of dev tool calls:

image

For now I shortcircuit /.well-known requests in this middleware.

```js
router.use((req, res, next) => {

  if (/(?<=\/.well-known\/appspecific)/.test(req.url)) {
    return;
  }

  if (xyzEnv.DIR && req.url.length === 1) {
    res.setHeader('location', `${xyzEnv.DIR}`);
    return res.status(302).send();
  }
  next();
});
```

apps/xyz/server.js
```diff
@@ -133,14 +130,18 @@ router.post(
...
 router.get(`${xyzEnv.DIR}/view{/:template}`, view);
 
-router.get(`/`, view);
+router.get(`${xyzEnv.DIR}/`, view);
 
-const app = express();
-app.disable('x-powered-by');
 app.use(router);
 
+app.use(validateRequestParams);
```
markjose-hhl: I think I was expecting this to be on the XYZ router

apps/xyz/router.js
markjose-hhl, line 1: I was expecting a little more of this, I think when we talked about what the factory did, I had a bit more in my mind. I pictured this being an `xyzFactory` exposing a method called something like `createXyzMiddleware` and which received an array of middleware that was optional for implementations outside of XYZ.

apps/xyz/server.js
markjose-hhl, line 1: With this new pattern I wouldn't expect to see any of the `router` code in here, I'd see this in the `xyzFactory` where this is a simple Express Middleware server which injects XYZ and the base authentication implementation.

markjose-hhl (changes requested):

dbauszus-glx:
@markjose-hhl Splitting the router from server app is a good idea. I think there should be a router module which exports the router and a server module which imports the router. The router can then be imported in another monorepo app to be modified.

Alternatively the xyz routes can also be imported into a custom app as there are just a handful of route modules anyways.

apps/xyz/server.js
dbauszus-glx: Yes. Same for me. The middleware should be on the router, not the app.

apps/xyz/server.js
dbauszus-glx: Agree. There should be a router module exporting the router and a server module importing the router to use in a server app.

dbauszus-glx:
@AlexanderGeere 

The createRouter method breaks if I don't provide a middleware argument.

image

image

The condition should be on the length of the middleware argument.

```
  if (middleWare.length) {
    router.use(middleWare);
  }
```

I added the rootRedirect method to the middleware modules.

Is the export of the xyzRouter and the default app required?

```
export const xyzRouter = router;
export default app;
```

Wouldn't a custom server script import the createRouter method?

apps/xyz/router.js
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  router.use(limiter);
+
+  router.use(cookieParser());
+
+  router.use(`${xyzEnv.DIR}/public`, express.static(publicDir));
+  router.use(xyzEnv.DIR, express.static(publicDir));
+
+  if (middleWare.length) {
```
markjose-hhl: I'm being pedantic but `middleware` is all one word so no capital 'W'. We should also pluralise to indicate an array. Technically not a correct plural but indicative and accepted nomenclature.

markjose-hhl (approved):

dbauszus-glx:
The middleware cannot be applied to the router itself. Otherwise we could get a loop condition where the redirect to the login goes through the auth middleware and redirects to the login ad infinitum.

dbauszus-glx:
The auth-server test for the server script in the auth app should not be in the xyz app. I moved the test script to auth app but this will need a better a look at some point.

dbauszus-glx:
I comment out the check for signed requests to the provider module in the auth module for the time being. I believe this logic to be legacy but need to have another look at this next week.

The failing test in this regard are bogus since signed requests are only intended for provider endpoints.

apps/xyz/mod/user/auth.js
```diff
@@ -128,7 +127,7 @@ We compare the given signature to one calcualted from the key and the url.
 
 @returns {Object} returns an object containing whether or not the signature verification passed.
 */
-function keyVerification(req, res) {
+function signatureVerification(req, res) {
```
RobAndrewHurst: This is dead code as it's not referenced anywhere

dbauszus-glx: I am looking at this today. I believe this code to be mostly legacy. The logic should only apply to provider endpoints.

AlexanderGeere: We added this so you can authenticate and download files through the file endpoint, Its not in the providers directly as that would require circumventing the auth so we put as part of the auth. [#2453](https://github.com/GEOLYTIX/xyz/pull/2453)

dbauszus-glx: Correct. We couldn't circumvent then because we didn't use middleware for requests but piped everything through the API module.

dbauszus-glx:
I moved the signature verification for now into the auth app. this isn't used nor ready for production and i don't want to push this in just because until there is a good use case to test this with.

AlexanderGeere (approved):
I put the redirect back for missing DIR.

This is working as expected otherwise, i am able to login and logout with both the custom login and the normal xyz login

dbauszus-glx:
@AlexanderGeere 

the rootRedirect is provided as middleware.

image

This is now executing twice.

dbauszus-glx:
The shortcircuits in the validateRequestAuth middleware have now been removed they are not required by middleware.

login, logout, register routes go straight to their respective endpoints without passing through the validateRequestAuth middleware to shortcircuit.

AlexanderGeere:
@dbauszus-glx 

I added a bare `/` route which uses the middleware, so in the case when go to just `localhost:3000` you will get redirected to the `DIR` if specified.

PR #2785, opened 2026-04-22
DIR route and redirect

dbauszus-glx:
The redirect if the DIR is defined but not provided should happen in middleware.

It looks like the DIR route was accidentaly removed in the major branch.


RobAndrewHurst:
I have added in an optional environment variable `XYZ_CWD` that tells the XYZ app where to resolve workspace-relative files from.

Why it exists:
- The SAML server may be started from `apps/saml`.
- Some XYZ resources live at the workspace root, such as `public/`, `resources/`, and key files.
- Instead of changing `process.cwd()`, the SAML server sets `XYZ_CWD` to the workspace root.

In `apps/saml/server.js`:

```js
process.env.XYZ_CWD ??= workspaceRoot;
```

Then XYZ uses:

```js
xyzEnv.XYZ_CWD || process.cwd()
```

for file lookups.

Currently it affects:
- Static public files in `apps/xyz/server.js`
- `SECRET_KEY` and `SIGN_*` pem files in `processEnv.js`
- Local file provider resources in `mod/provider/file.js`

If `XYZ_ROOT` is not set, XYZ falls back to the old behavior and resolves files from `process.cwd()`. This preserves normal standalone usage while making SAML startup independent of the launch directory.

Example:

```sh
XYZ_ROOT=/home/rob-geo/Documents/geolytix.nosync/xyz pnpm dev:saml
```

With that value, XYZ resolves workspace files like:

```txt
/home/rob-geo/Documents/geolytix.nosync/xyz/public
/home/rob-geo/Documents/geolytix.nosync/xyz/resources
/home/rob-geo/Documents/geolytix.nosync/xyz/my-signing-key.pem
```

In the SAML app this is set automatically:

```js
const workspaceRoot = fileURLToPath(new URL('../..', import.meta.url));

process.env.XYZ_CWD ??= workspaceRoot;
```

So if `apps/saml/server.js` is here:

```txt
/home/rob-geo/Documents/geolytix.nosync/xyz/apps/saml/server.js
```

then `XYZ_CWD` becomes:

```txt
/home/rob-geo/Documents/geolytix.nosync/xyz

AlexanderGeere (commented):
Running this locally i have DIR set in my .env, But when i go to localhost in the browser its not redirecting to the DIR. That is to say i dont see the dir in the URL:

Image

my DIR is set to `latest`

dbauszus-glx:
@AlexanderGeere Which is the initial address `localhost:3000` or `localhost:3000/`?

AlexanderGeere:
> @AlexanderGeere Which is the initial address `localhost:3000` or `localhost:3000/`?

Its the same result for both

RobAndrewHurst:
@dbauszus-glx @AlexanderGeere I have resolved this. 

The .env was not being loaded by dotenv because it was looking for the .env in the root of the xyz app. I have changed this in the processEnv module to load it from the actual root

AlexanderGeere (commented):
I have pushed a change for the cloudfront certificate, i was getting a not found error in the console for the key.

This loads correctly now without the DIR, but if i supply a DIR i get:

Image

AlexanderGeere:
The DIR variable needs a leading `/`, i accidentally removed it from my env. This is working as expected otherwise.

dbauszus-glx:
I added a check to processEnv util to ensure that DIR starts with and doesn't end with a slash.

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #2790, opened 2026-04-29
Custom auth layer

### LINKED ISSUE #2791: Provide apps template for custom auth provider
The v5 release allows for apps to extend the [express] router with custom auth routes for login.

A template for a custom auth provider should be added as an app to the monorepoo to test the ACL lookup.

The saml condition should be removed.

A single environment variable should be enough for the auth routing eg `"AUTH_PATH": "/custom",`

The validateRequestAuth method should redirect login requests to the custom /login route.

The custom login route can be added to the router exported from the apps/xyz/server.js script.

dbauszus-glx:
This PR introduces a template for a custom authentication layer.

Instead of relying on the saml configuration in the api module validateRequestAuth method a check is made for the AUTH_PATH environment variable.

After setting the redirect the request will be redirected to either the custom or the default login route.

```js
  // PRIVATE instances require user auth for all requests.
  if (!req.params.user && xyzEnv.PRIVATE) {
    loginRedirect(req, res);

    if (xyzEnv.AUTH_PATH) {
      res.setHeader('location', `${xyzEnv.DIR}${xyzEnv.AUTH_PATH}/login`);
    } else {
      res.setHeader('location', `${xyzEnv.DIR}/api/user/login`);
    }

    res.status(302).send();
    return;
  }
```

The user cookie and redirect cookie will be set in the loginRedirect of the api module. The utils/redirect method has been removed.

The custom login auth logic is now relative straight forward. Redirected to the custom/login view a the form action sends the email to the custom/verify method. The user with a lookup flag is passed to the user/redirect module method which handles user lookup and redirect.

```js
app.get(`${xyzEnv.DIR}/custom/login`, custom_login);
app.post(
  `${xyzEnv.DIR}/custom/verify`,
  [express.urlencoded({ extended: true }), express.json({ limit: '5mb' })],
  custom_verify,
);

function custom_login(req, res) {
  const form = `<form method="POST" action="${xyzEnv.DIR}/custom/verify">
    <input type="text" name="username" placeholder="Username" required />
    <button type="submit">Login</button></form>`;

  res.send(form);
}

async function custom_verify(req, res) {
  const user = {
    email: req.body.username,
    lookup: true,
  };
  redirect(req, res, user);
}
```

An api/user/logout route was added.

The behaviour for parameterised routing for the login, logout, and register route have not been changed in this PR.


dbauszus-glx:
@RobAndrewHurst Do you want to have a crack of moving the auth-server script into it's own apps directory.

AlexanderGeere (commented):
testing wtih `xyz-keycloak` Login works as expected. I am unable to logout however.

Logging in works with a DIR specified as well

dbauszus-glx:
@AlexanderGeere Thanks. I wasn't expecting the logout to work as I was unable to test the login yet with keycloak.

RobAndrewHurst:
> @AlexanderGeere Thanks. I wasn't expecting the logout to work as I was unable to test the login yet with keycloak.

I have fixed the logout now, the SessionIndex and profile properties were removed from the user cookie. This is needed for the Idp to know what session to terminate.

RobAndrewHurst (approved):

AlexanderGeere (changes requested):
I am not able to logout. I tried with and without the SAML_SLO. 

I am using a DIR env variable and im using `pnpm dev:saml` to launch the app.

RobAndrewHurst:
**Implemented Changes**
- Preserved SAML identity fields from ACS into the signed user cookie:
  - `nameID`
  - `nameIDFormat`
  - `nameQualifier`
  - `sessionIndex`
  - `spNameQualifier`

- Updated SAML logout handling:
  - If the user cookie contains `sessionIndex`, `/saml/logout` calls `getLogoutUrlAsync()` and redirects to the IdP logout URL.
  - If there is no SAML session, logout falls back to clearing the local cookie.
  - Cookie decoding is safe when no cookie object exists.

- Refactored SAML handler for testability:
  - Added `createSamlHandler(strategy, redirectUser)`.
  - Lazily creates the real `SAML` client only for the default runtime handler.
  - Allows tests to inject a fake SAML strategy without requiring real IdP config or certs.

- Hardened the shared `redirect()` helper:
  - Rejects missing ACL users instead of signing the originally supplied user object.
  - Rejects blocked users.
  - Rejects explicitly unverified users.
  - Rejects explicitly unapproved users.
  - Handles missing ACL configuration.
  - Revalidates redirect-cookie values before redirecting.
  - Rejects protocol-relative redirects such as `//example.com`.
  - Falls back to `${DIR}/` for unsafe or malformed redirect-cookie values.

- Fixed async handling:
  - `custom_verify()` now awaits `redirect()`.
  - SAML ACS now awaits the injected/shared redirect helper.

- Fixed frontend auth links:
  - Exposes `AUTH_PATH` to the rendered page as `data-auth-path`.
  - Login plugin now uses `DIR + AUTH_PATH` when configured.
  - For SAML deployments, login/logout links resolve to:
    - `${DIR}/saml/login`
    - `${DIR}/saml/logout`
  - Without `AUTH_PATH`, existing query-flag behaviour remains:
    - `?login=true`
    - `?logout=true`

- Rebuilt the browser bundle:
  - `public/js/lib/mapp.js`
  - `public/js/lib/mapp.js.map`

**New Test Coverage**
Added focused regression tests for the new auth code.

`apps/xyz/tests/mod/user/redirect.test.mjs`
- Signs a user cookie after successful ACL lookup.
- Clears the redirect cookie after login.
- Redirects to a safe stored redirect.
- Rejects missing ACL users.
- Rejects blocked ACL users.
- Falls back to `${DIR}/` for unsafe redirect-cookie values.

`apps/xyz/tests/mod/saml.test.mjs`
- Verifies SAML logout redirects through the IdP when `sessionIndex` exists.
- Verifies logout falls back to local cookie clearing when no `sessionIndex` exists.
- Verifies ACS passes SAML identity fields into the shared redirect helper.

`apps/xyz/tests/mod/auth-server.test.mjs`
- Verifies custom auth routes are registered under `DIR`.
- Verifies the custom login form posts back under `DIR`.
- Verifies custom verify delegates to the shared redirect helper.

`apps/xyz/tests/plugins/login.test.mjs`
- Verifies login plugin uses `DIR + AUTH_PATH` for SAML logout.
- Verifies login plugin uses `DIR + AUTH_PATH` for SAML login.
- Verifies fallback query-flag behaviour when `AUTH_PATH` is not configured.

**Risk Reduced**
This change reduces risk around:
- Local-only logout leaving the IdP session active.
- Signing cookies for missing or blocked ACL users.
- Unsafe redirect-cookie values being trusted after authentication.
- SAML code being difficult to test due to real client construction at import time.
- `DIR`/subpath deployments generating incorrect login/logout URLs.

AlexanderGeere (approved):
Can now logout and login through the saml. 
Works with a DIR as well. 

Normal login also works.

dbauszus-glx:
I added a custom/logout method for testing.

I am ok for this getting merged but will create a ticket to eventually clean up the user evaluation. There shouldn't be a getUser and evaluate method in the fromACL module using the ACL import. Then the same in the cookie module  though only checking for blocked here. The same checks again in the auth module checkParamToken method. And now again in the redirect method.

This wants to be addressed with a review of the ACL method.

I created a seperate ticket for this. https://github.com/GEOLYTIX/xyz/issues/2803

AlexanderGeere (approved):
I made a couple tiny fixes, custom verify was customVerify and registerAuthRoutes didnt exist. 

I tested by setting auth_path to custom. Working as expected

PR #2796, opened 2026-04-30
Cloudfront xss

### LINKED ISSUE #2798: Handle provider JavaScript MIME types without reopening XSS risk
## Summary
SonarQube flags `mod/provider/_provider.js` for reflected XSS because provider responses can flow from remote/provider content into `res.send()`.

The existing mitigation constrains provider response content types and sets `X-Content-Type-Options: nosniff`, but plugins may be loaded as JavaScript modules from any provider. Those `.js` and `.mjs` resources need a browser-loadable JavaScript MIME type.

## Desired behaviour
- Keep `X-Content-Type-Options: nosniff` on provider responses.
- Do not trust arbitrary request `content_type` values for executable MIME types.
- Serve provider resources ending in `.js` or `.mjs` as `text/javascript` so dynamic plugin imports continue to work.
- Keep non-JavaScript provider strings as `text/plain` unless explicitly allowed as `application/json`.

## Acceptance criteria
- Plugin module URLs ending in `.js` or `.mjs` can be loaded from any provider.
- `content_type=text/javascript` on a non-JavaScript resource is ignored/falls back to `text/plain`.
- Unit tests cover both valid plugin MIME handling and rejection of arbitrary JavaScript content types.
- SonarQube XSS finding is reviewed after the change; if still flagged, document why the executable MIME type is derived from the resource URL rather than arbitrary user input.

RobAndrewHurst:
Updated the provider flow to reduce the XSS/reflection risk SonarCloud flagged.

Changes made:
- `mod/provider/_provider.js`
  - No longer reflects `req.params.provider` in error responses.
  - Handles provider `Error` responses with a generic `500` message.
  - Adds `X-Content-Type-Options: nosniff`.
  - Restricts response content types to `application/json` or `text/plain`.
  - Falls back to `text/plain` for unsafe or unknown `content_type` params.
  - Sends object responses via `res.json()`.

- `mod/provider/cloudfront.js`
  - Removed user-controlled URL from CloudFront error messages.

- Tests updated:
  - `tests/mod/provider/_provider.test.mjs`
  - `tests/mod/provider/cloudfront.test.mjs`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,11 +12,26 @@ jobs:
   sonarqube:
     name: SonarQube
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    env:
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Skip SonarQube for fork pull requests
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+        run: |
+          echo "Skipping SonarQube scan because this pull request comes from a fork."
+          echo "GitHub does not expose repository secrets, including SONAR_TOKEN, to fork pull_request workflows."
+      - name: Check SonarQube token
+        if: (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && env.SONAR_TOKEN == ''
+        run: |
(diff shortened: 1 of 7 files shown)


AlexanderGeere (changes requested):
I don't seem to be able to load plugins on this PR:

Image

Im testing with `bugs_testing/filter/all_filters.json` and using the cloudfront plugins url

RobAndrewHurst:
@AlexanderGeere This should be good now!

dbauszus-glx (approved):
Works for me. I unnested the declarations provider and allowedContentTypes from the function execution and merged the lookup method for the allowedContentTypes into the provider method.

I used ai to rewrite the module and metthod documentation thereafter.

I set the default contentype 'text/plain' first to be changed if needed  rather than checking first and assigning default if nothing else matches.

I was not able to test an object response from a real provider.

PR #2800, opened 2026-05-06
redact user data from logs

RobAndrewHurst:
Implemented a SonarCloud security fix for `mod/utils/logger.js`.

The logger no longer writes raw user-controlled values to any logging sink. Before output reaches the configured remote logger, `console.error`, or `console.log`, it is converted into a fixed safe marker:

- `Error` values log as `[Error]`
- arrays log as `[Array]`
- objects log as `[Object]`
- primitive/string values log as `[REDACTED]`

This addresses SonarCloud rule `jssecurity:S5145` by removing the tainted value flow into logging sinks.

I also added focused tests in `tests/mod/utils/logger.test.mjs` covering:

- object redaction before console logging
- string redaction before console logging
- error redaction before stderr logging
- redacted values being sent to the configured Logflare logger

### DIFF
--- a/mod/utils/logger.js
+++ b/mod/utils/logger.js
@@ -150,17 +150,17 @@ function postgresql() {
   }
 
   // Sanitize the params.table once at init to ensure no SQL injection.
-  const table = params.table.replace(/[^a-zA-Z0-9_.]/g, '');
+  const table = params.table.replaceAll(/[^a-zA-Z0-9_.]/g, '');
 
   return async (log, key) => {
     // Dynamic import to avoid circular dependency (dbs.js imports logger.js).
     const { default: dbs } = await import('./dbs.js');
 
     dbs[params.dbs](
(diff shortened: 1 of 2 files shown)


dbauszus-glx (changes requested):
I don't understand what the PR is trying to resolve.

I can hit the logger with a bit of work from the fromACL method.

The log value is a string which will be redacted since the log value is neither an object, nor array.

With this PR the `[REDACTED]` value will be logged instead of a string. This is useless and we could as well remove the logger module completely.

Looking at the linked sonarqube report it is suggested to remove CLRF characters.

Image

dbauszus-glx:
@RobAndrewHurst have a look at my last commit where I just remove the CLRF character as suggested by the sonarqube report.

image

This should resolve the ticket.

I have commented sanitizeLog method in the committ.

PR #2802, opened 2026-05-06
update number parsing smells

RobAndrewHurst:
# Pretty self explanatory


dbauszus-glx (approved):
Reads well, though impossible to fully test.

PR #2806, opened 2026-05-08
fix(query): Take copy of template to avoid mutation and add tests

### LINKED ISSUE #2805: DBS Connection: Core Templates given dbs connection in v4.23.0
In v4.23.0, a core template (`locationCount`) for example is given a `dbs` connection. 

In the `query.js` module, these two lines were in the code 
``` js
// Use layer dbs as fallback if template dbs is not defined.
  template.dbs ??= req.params.layer?.dbs;

  // Use workspace dbs as fallback if not explicit or from layer.
  template.dbs ??= req.params.workspace.dbs;
```

This has a few issues:

1. It meant that the second time a core template was called, it was using the original dbs connection set to the template. For example if locationCount was called from a layer with `dbs:A` first, then called again on layer with `dbs:B` - `dbs:A` would still be used. 
2. I am not sure how the second nullish assignment could ever be applied? 

This should not be a nullish assignment but always an assignment of - layer.dbs, workspace.dbs, template.dbs. 
This should support existing functionality while fixing the issue.

simon-leech:
## Description

This PR makes the query module take a copy of the `template` before mutating its properties. This is required as using when a `core` template was used across layers, it was always using the `dbs` connection of the first layer that called it. It meant that the second time a core template was called, it was using the original dbs connection set to the template. For example if locationCount was called from a layer with dbs:A first, then called again on layer with dbs:B - dbs:A would still be used.

This PR also rewrites
``` js
// Use layer dbs as fallback if template dbs is not defined.
  template.dbs ??= req.params.layer?.dbs;

  // Use workspace dbs as fallback if not explicit or from layer.
  template.dbs ??= req.params.workspace.dbs;
```
to 
``` js
template.dbs =
    req.params.layer?.dbs ||
    req.params.workspace.dbs ||
    template.dbs;
```

This PR also tidies up the test module, removing the unused `_query.test.mjs` and moving all tests to `query.test.mjs`. The invalid dbs connection test has not been deleted but moved into a describe with additional db connection tests.

## Type of Change

## How have you tested this?
Tested locally and added additional tests to cover this. 

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine


dbauszus-glx (commented):
The template in the query module should be a clone and modifications should not be stored in the template at rest.

dbauszus-glx (dismissed):
It must not be possible to provide the DBS as a param as this would be a significant security issue which we removed a while ago. Otherwise you could send a query to any DBS if you guess an identifier which itself is not sensitive.

RobAndrewHurst (approved):
Happy for this to go in!

AlexanderGeere (approved):
Working as expected

PR #2809, opened 2026-05-12
fastify.js server

dbauszus-glx:
draft to test fastify.js performance


dbauszus-glx:
the password which was accidentaly spilled has been reset.

PR #2812, opened 2026-05-12
Fastify perf

### LINKED ISSUE #2789: Fastify routing
It is worth testing a fastify app for routing. XYZ was initially written with fastify and only shifted to express for local testing due some request object compatibility. With the v5 major release we will deploy the router app to vercel as a serverless function hence compatibility should no longer be an issue. Fastify should perform superiour for large number of requests expected in multi tenant setup. XYZ workspace compilation dependends on JSON parsing which is much improved with fastify.

dbauszus-glx:
New clean branch for fastify test.

Tests are expected to fail as we only added a handful of routes for testing performance.

### DIFF
--- a/apps/xyz/fastify.js
+++ b/apps/xyz/fastify.js
@@ -77,21 +77,21 @@ if (process.versions.node.split('.')[0] < 22) {
 
 // app.use(cookieParser());
 
-import Fastify from 'fastify'
+import Fastify from 'fastify';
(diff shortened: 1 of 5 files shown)


dbauszus-glx:
Test for the locale endpoint with a workspace served from the file system showed that fastify and express latency and request throughput are vistually identical.

image

PR #2813, opened 2026-05-12
The dotenv dependency is required in the xyz-app

dbauszus-glx:
Deploying apps/xyz/server script will fail without dotenv in the apps package dependencies.


AlexanderGeere (approved):

dbauszus-glx:
This is no longer required with varlock. https://github.com/GEOLYTIX/xyz/pull/2822

PR #2814, opened 2026-05-12
Control whether entry options should be dynamic

### LINKED ISSUE #2810: Entry Edit Options Not Being Used when Provided
This bug looks to have been mistakenly introduced in this PR - https://github.com/GEOLYTIX/xyz/pull/2671

There is two scenarios with Edit Options we need to cater for (and have tests for): 

### Entry.Edit.Options Provided
In this scenario, the options are fixed, they should not be altered at any point.
``` json
{
"edit": {
"options": ["A","B","C"]
}
}
```

### Entry.Edit.Options Not Provided
In this scenario, the options are retrieved from the db using `distinct_values` query, they should be updated after saving changes too in case of new values (that's what PR 2671 addressed). 
``` json
{
"edit": {
"options": []
}
}
```

## Proposed Bug Location
I think the bug location is in `onkeyup`. This assumes that if `edit.options` is provided to just wipe it out so that it is re-ran from the db. 
However, this is not the correct approach for option 1 above. Instead I think we need to add `edit.optionsProvided` flag that is automatically attached to an entry when options are manually configured. Then this check in `onkeyup` can check for the presence of this flag and not wipe out the options if its present. 

``` js
function onkeyup(e, entry) {
  // Split target value as array if an arraySeparator has been provided.
  entry.newValue = entry.edit.arraySeparator
    ? e.target.value.split(entry.edit.arraySeparator)
    : e.target.value;

  entry.location.infoj.find((el) => {
    if (
      (entry.field === el.field || entry.json_field === el.json_field) &&
      el.edit?.options
    ) {
      //Remove options so the query is rerun.
      el.edit.options = [];
    }
  });

  entry.location.view?.dispatchEvent(
    new CustomEvent('valChange', { detail: entry }),
  );
}
```

dbauszus-glx:
The entry.options should not be reset on keyup.

A flag `edit.dyanamic = true` has been introduced in this PR. The options will be set to an empty array if used. This will force the options to be populated from a query.

### DIFF
--- a/lib/ui/locations/entries/text.mjs
+++ b/lib/ui/locations/entries/text.mjs
@@ -104,16 +104,6 @@ function onkeyup(e, entry) {
     ? e.target.value.split(entry.edit.arraySeparator)
     : e.target.value;
 
-  entry.location.infoj.find((el) => {
-    if (
-      (entry.field === el.field || entry.json_field === el.json_field) &&
-      el.edit?.options
-    ) {
-      //Remove options so the query is rerun.
-      el.edit.options = [];
-    }
-  });
-
   entry.location.view?.dispatchEvent(
     new CustomEvent('valChange', { detail: entry }),
   );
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
This is working for me! 

Happy with the pattern that by default things do not automatically refresh, but adding the `dynamic:true` flag allows them to. This follows the same logic for hovers and dataviews 👍 

Thanks for fixing the bug too with `edit.options` being wiped accidentally in `onkeyup`

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #2820, opened 2026-05-13
Log user.email instead of request params.

dbauszus-glx:
This PR reduces the risk of logging sensitive or user-controlled data across mailer, view, and fetch logging.

## Summary

- Removes the `mailer_body` logger key, which could include sensitive request body data and provides limited logging value.
- Updates the `mailer` logger key to read the recipient email address from the user object instead of the request body, avoiding logging user-controlled input.
- Removes the `view-req-url` logger key because it adds limited value and may expose user-controlled request URL data.
- Updates `fetch` logging in the `getFrom` module to log the fetched response URL and status instead of logging the request URL passed into the template module.
- Corrects the params/property documentation for the `fromACL` method.

## Rationale

Request-derived values such as body fields and request URLs can contain sensitive or user-controlled data. Logging these values increases risk without adding much operational value.

Request-level logging should happen at the edge/network layer, where all incoming requests can be captured consistently, including requests that never reach view handlers or other application methods.

For fetched resources, the response object already exposes the final `url` alongside the response `status`. This is a better logging source than the original request URL passed into the template module, and it avoids logging unsanitized user-controlled input.

### DIFF
--- a/mod/provider/getFrom.js
+++ b/mod/provider/getFrom.js
@@ -88,7 +88,7 @@ async function Https(url) {
   try {
     const response = await fetch(url);
 
-    logger(`${response.status} - ${url}`, 'fetch');
+    logger(`${response.status} - ${response.url}`, 'fetch');
 
     if (url.match(/\.json$/i)) {
       return await response.json();
--- a/mod/user/fromACL.js
+++ b/mod/user/fromACL.js
@@ -204,7 +204,7 @@ async function getUser(request) {
     return user;
   }
 
-  return await failedLogin(request);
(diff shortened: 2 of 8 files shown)


dbauszus-glx:
@RobAndrewHurst The test fails because of the mocking.

The row [user] returned from the acl must have an email as specified in the return of the SQL statement.

image

AlexanderGeere (approved):

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #2829, opened 2026-05-20
fix(mergeTemplates): Roles object creating dot notation roles mistakenly

### LINKED ISSUE #2828: `roles` object inside a template is mistakenly creating dot notation roles
A template with a `roles` object inside is mistakenly creating dot notation roles. 

If my locale has a `roles` object containing 
``` json
"roles": {
"Super": null,
"Standard": null,
"A": null 
}
```

and a template has the same thing inside it. The roles endpoint is returning `Super.Super`, `Super.Standard`, `Super.A` etc. 

This is wrong, a `roles` object should not create dot notations, only a `role`.

simon-leech:
## Description

Please include a summary of the changes, and the motivation for the change.

## Type of Change

- ✅ Testing

## How have you tested this?
Failing test provided for the problem.

## Testing Checklist
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

### DIFF
--- a/mod/utils/roles.js
+++ b/mod/utils/roles.js
@@ -240,7 +240,7 @@ export function combine(child, parent) {
   }
 
   // Handle simple locale-style role combination
-  combineLocaleRoles(child, parent);
+  combineObjRoles(child, parent);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


RobAndrewHurst:
Fixed role combination so `roles` object entries are not treated as dot-notation hierarchy sources.

The issue was caused by `combineLocaleRoles` using every key from `parent.roles` and `child.roles` to generate combinations. That meant a plain roles object like:

```json
{
  "roles": {
    "Super": null,
    "Standard": null,
    "A": null
  }
}
```

could incorrectly produce generated roles such as `Super.Super`, `Super.Standard`, `Super.A`, etc.

The implementation now only creates dot-notation roles when both parent and child have explicit `role` string properties. `roles` object keys are still collected as valid standalone roles, but they no longer create hierarchy combinations by themselves.

Test coverage added/updated:
- Added regression coverage for a template/locales workspace using only `roles` objects.
- Updated the utility test to assert that `roles` object keys do not combine.
- Updated nested role expectations to remove shortcut ancestor roles that are no longer generated by scoped role combination.

dbauszus-glx (approved):
This works well for me. I updated the documentation a little bit and changed name from locale to obj since this applies to all kinds of workspace objects like locale, layer, etc.

I tested this with:
```
{
  "templates": {
    "OSM-template": {
      "role": "test-OSM-template",
      "roles": {
        "OSM-layer": true
      },
      "format": "tiles",
      "URI": "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png"
    },
    "attribution-template": {
      "roles": {
        "attribution-property": "true"
      },
      "attribution": {
      }
    }
  },
  "locales": {
    "foo": {
      "roles": {
        "foo": true
      },
      "layers": {
        "OSM": {
          "template": "OSM-template",
          "templates": [
            "attribution-template"
          ],
          "display": true
        }
      }
    },
    "bar": {
      "role": "test-bar",
      "roles": {
        "bar": true
      },
      "layers": {
        "OSM": {
          "template": "OSM-template",
          "templates": [
            "attribution-template"
          ],
          "display": true
        }
      }
    }
  }
}
```

and got the expected result. 
```
["attribution-property","bar","foo","OSM-layer","test-bar","test-bar.test-OSM-template","test-OSM-template"]
```

AlexanderGeere (approved):

PR #2832, opened 2026-05-21
Mapview modules; this assignment, documentation, named methods

dbauszus-glx:
This PR removes the assignment of `const mapview = this` in several mapview modules.

Assigning this to a variable is an outdated JavaScript pattern that creates unnecessary complexity and reduces code readability.

In JavaScript, the value of this depends on how a function is called. Before ES6, developers commonly assigned this to a variable (like self, that, or foo) to preserve the context when calling functions where this would have a different value.

This pattern has several drawbacks:

It creates additional variables that clutter the scope
It can cause confusion about which context is being referenced
It makes the code harder to understand and maintain
It’s prone to errors if the wrong variable is used
Modern JavaScript provides better alternatives that eliminate the need for this pattern. Arrow functions automatically preserve the lexical this binding, meaning they inherit this from the enclosing scope. The Function.prototype.bind() method can also explicitly set the this value for regular functions.

The touchStart method has been unnested from the highlight interaction method.

The documentation for several modules and methods was either missing or all over the place.

### DIFF
--- a/lib/mapview/infotip.mjs
+++ b/lib/mapview/infotip.mjs
@@ -1,16 +1,21 @@
 /**
-## Mapview.infotip()
+## /mapview/infotip
 
 @module /mapview/infotip
-
-@param {Object} content
 */
 
 const infotip = {};
 
-export default function (content) {
-  const mapview = this;
+/**
+@function Infotip
 
+@description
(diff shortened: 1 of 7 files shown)


AlexanderGeere (approved):

cityremade:
is this `this` needed? `mapview = this) {` ?

simon-leech (approved):
Tested on default mapview and also a custom view with custom mapviews and custom interactions and looks to be working as previously

PR #2835, opened 2026-05-26
Add cluster.label to layerFields set

dbauszus-glx:
The cluster.label field must be valid for requests to the WKT layer query template.

The cluster.label property value must be added to the layerFields set in the checkFieldsParam method to allow requests to pass.


dbauszus-glx:
This should be on a patch branch.

PR #2836, opened 2026-05-26
Layer cluster label field

### LINKED ISSUE #2834: `checkFieldsParam` - field used in clusters are erroring
If a layer has a `cluster` object that contains a `label` that is not present in the `infoj`. 
This field is missed in the `layerFields` creation method, when created by `objPropValueSet`. 

e.g. 
``` json
 "cluster": {
    "distance": 15,
    "label": "operator"
  }
```

If `operator` is not shown in the infoj, the layer will not load currently. 
This is incorrect as this is valid configuration. The method must check in the `cluster.label` value too.

dbauszus-glx:
The cluster.label field must be valid for requests to the WKT layer query template.

The cluster.label property value must be added to the layerFields set in the checkFieldsParam method to allow requests to pass.

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -359,11 +359,6 @@ async function checkFieldsParam(req, res) {
 
   objPropValueSet(req.params.layer, 'field', layerFields);
 
-  // Technical debt: The cluster label should be added as a field in the layer template to prevent direct reference here.
-  if (req.params.layer.cluster?.label) {
-    layerFields.add(req.params.layer.cluster.label);
-  }
-
   req.params.fieldsMap = new Map();
 
   for (const field of req.params.fields.split(',')) {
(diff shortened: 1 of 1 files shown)


cityremade (approved):

simon-leech (approved):
This works for me! 
A thought though - What about if we nullish assign cluster.label to cluster.field inside styleParser? Then it would always be picked up without this additional if statement?

dbauszus-glx:
@simon-leech That wouldn't matter because the check for fields is a security feature and happens in the query module with the template from the workspace. The template cannot be modified in the cache.

simon-leech:
> @simon-leech That wouldn't matter because the check for fields is a security feature and happens in the query module with the template from the workspace. The template cannot be modified in the cache.

Ahh of course - thanks for explaining! 👍

PR #2837, opened 2026-05-26
Legendicon function return

dbauszus-glx:
This PR fixes an issue introduced in this patch commit. https://github.com/GEOLYTIX/xyz/commit/23e16a179fbc6fbdad4d773ad00d698c313fa85d

The svgSymbolUrl method must return an SVG url, not the method which will cause a crash.

### DIFF
--- a/lib/ui/elements/legendIcon.mjs
+++ b/lib/ui/elements/legendIcon.mjs
@@ -175,7 +175,7 @@ function svgSymbolUrl(icon) {
 
   if (!Object.hasOwn(mapp.utils.svgSymbols, icon.type)) return;
 
-  return mapp.utils.svgSymbols[icon.type](icon.type);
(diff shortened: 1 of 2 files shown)


simon-leech:
image

This is throwing thousands of warnings into the console 

Which is leading to white legend dots even though the configuration is unchanged
image

simon-leech (changes requested):
Wrote feedback as a comment but forgot to mark as Request changes

AlexanderGeere (commented):
@dbauszus-glx 
I changed to pass the icon not just the type, i was testing on `bugs_testing/filter/all_filters.json` and i was getting a lot of warnings in the console. This fixed those warnings.

simon-leech (approved):
This looks to be working for me now 🥇

AlexanderGeere (approved):
Saving the locale works!

All working as expected

PR #2839, opened 2026-05-27
fix(Templates/Props): Exclude / Include Props

### LINKED ISSUE #2838: Exclude / Include Props with Template and Templates Array Not Working Correctly
When a `layer` has a `template` object, and a `templates` array. 
If I use `exclude_props` to exclude properties from the template loading - it should only apply to the `template` itself. 
Subsequent loading `templates` from the `templates` array should not have that property excluded too.

**Example Configuration** 

``` json
"layers": {
"layer_a": {
    "template": {
      "key": "layer_a",
      "src": "filepath.json",
      "exclude_props": [
        "report_params",
        "templates"
      ]
    },
    "templates": [
      {
        "src": "another_filepath.json"
      }
]
}
```

simon-leech:
## Description
This PR provides a failing test that shows the issue highlighted in the attached PR.

### DIFF
--- a/mod/workspace/mergeTemplates.js
+++ b/mod/workspace/mergeTemplates.js
@@ -107,7 +107,7 @@ Templates defined in the obj.templates array will be merged into object.
 
 @param {Object} obj
 @param {Object} template The template maybe an object with a src property or a string.
-@param {array} roles An array of user roles from request params.
+@param {array} [roles] An array of user roles from request params.
 @param {Object} [context] Optional base roles for context.
 @property {string} [obj.template] Key of template for the object.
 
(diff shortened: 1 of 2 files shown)


dbauszus-glx:
The templateProperties logic has been merged with the prepareTemplates method. The name was misleading and debugging harder with a single method being called from a single method. The method has been renamed filterTemplateProperties and does just that.

The Roles.objMerge(template, roles); has been unnested into the objTemplate method to be called prior to the filterTemplateProperties method. There is no need to nest the logic and provide the roles argument to the filterTemplateProperties method.

The assignment of template.exclude_props and template.include_props has been simplified with a nullish assignment.

The object.hasOwn check before the property delete is not required.

The template.exclude_props property itself will be deleted after its own iteration. This prevent the exclude_props being assigned to the obj in the merge and then removing properties from subsequent templates being merged into the object.

dbauszus-glx:
@simon-leech Is there a test for the include_props property?

dbauszus-glx (approved):

simon-leech:
@dbauszus-glx I will add an include test now while re-testing!

simon-leech:
@dbauszus-glx Just pushed include tests that is current passing 🥇

RobAndrewHurst (approved):

PR #2847, opened 2026-06-01
Correct template dbs connection

### LINKED ISSUE #2846: Template `dbs` property not being used
The line 
``` js
template.dbs =
    req.params.layer?.dbs || req.params.workspace.dbs || template.dbs;
```
prevents a template.dbs from being set on a template - if the template has a layer or workspace, it will never be used. 

This order is incorrect and was a symptom of attempting to fix this PR - #2805

simon-leech:
This PR fixes an issue with a query template having a template `dbs` property and it not being used. 

The order of `dbs` should be template, layer, workspace. 
Tests have been updated to fix this issue.

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -112,8 +112,7 @@ export default async function query(req, res) {
     return;
   }
   // Use template dbs, layer dbs if defined, or workspace dbs in the query.
-  template.dbs =
-    template.dbs || req.params.layer?.dbs || req.params.workspace.dbs;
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):

dbauszus-glx (approved):

dbauszus-glx:
Rather than template.dbs = template.dbs || ... we can just use a nullish assignment.

```
template.dbs ??= req.params.layer?.dbs || req.params.workspace.dbs;
```

PR #2849, opened 2026-06-02
Fix SSO-enabled API requests

### LINKED ISSUE #2848: SSO and API Keys - being redirected to login
On an application that has SSO enabled, calling any endpoint with a `&key=xx` or `&token=xx` parameter **should** run the query immediately. 
However, we are being redirected to login every time which is wrong - a `key` or `token` supersedes the need for authorisation via login.

Likely due to this 

Image

RobAndrewHurst:
## What's the craic?
Fixes SSO-enabled API requests with key or token query parameters being redirected to SAML login before they can execute.

### Issue?
`api.js` short-circuited any /saml URL before request params were validated or auth/routing logic could run. 
In private SSO-enabled apps, requests containing key or token should
supersede login auth and run immediately, but they were redirected to SAML login instead.

### Changes
- Move SAML route handling until after request params are validated.
- Skip the SAML short-circuit when a key or token param is present.
- Skip the private SAML login redirect for key or token requests and route them normally.
- Return 401 for failed key/token auth attempts instead of redirecting to login.

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -173,7 +173,7 @@ async function validateRequestAuth(req, res) {
 
   // The authentication method returns an error.
   if (user && user instanceof Error) {
-    if (req.headers.authorization || req.params?.token || req.params?.key) {
+    if (req.headers.authorization || hasKeyOrToken(req)) {
       // Request with failed authorization headers are not passed to login.
       return res.status(401).send(user.message);
     }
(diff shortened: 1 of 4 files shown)


AlexanderGeere (changes requested):
I am able to call the `workspace/locales` endpoint without logging in or using a token in the params on this PR.

When i try on patch I am prompted to login.

AlexanderGeere (approved):

RobAndrewHurst:
@AlexanderGeere This should all be fixed now :)

dbauszus-glx (approved):
We will need to test how this works deployed with the vercel router. I am happy for this to go in as it is since the route logic will be replaced anyways with the v5 major release.

PR #2853, opened 2026-06-05
Remove user.admin flag from token user

### LINKED ISSUE #2852: user.admin flag removal
The user.admin flag should only be set on a cookie when the user cookie is generated from an existing cookie, not from a token or [api] key.

The current logic to apply the user.admin flag from the payload means that the admin flag is still provided if removed from the user record in the acl.

dbauszus-glx:
A user login from a token should not have admin rights.

Admin rights should be removed if acl record is updated.

### DIFF
--- a/mod/user/cookie.js
+++ b/mod/user/cookie.js
@@ -99,7 +99,9 @@ export default async function cookie(req, res) {
       const user = rows[0];
 
       // Admin rights should not be added if not provided from a token.
-      user.admin = payload.admin;
+      if (payload.from_token) {
+        delete user.admin;
+      }
 
       // Assign title identifier to user object.
       user.title = xyzEnv.TITLE;


AlexanderGeere (approved):

RobAndrewHurst (approved):

mod/user/cookie.js
```diff
@@ -99,7 +99,9 @@ export default async function cookie(req, res) {
       const user = rows[0];
 
       // Admin rights should not be added if not provided from a token.
-      user.admin = payload.admin;
+      if (payload.from_token) {
```
simon-leech, line 102: Am I misunderstanding or shouldn't this happen before the const user = line? 
As otherwise here if from_token is true then the admin property doesn't exist? We use this for checking if admin view button should be accessible for example?

dbauszus-glx: The from_token does not mean cookie. The flag is set only when you provide a token as token parameter in the checkParamToken method in the auth module.

simon-leech (approved):

PR #2855, opened 2026-06-05
Enable versioning of userIndexedDB to create additional stores

### LINKED ISSUE #2804: Requirement to sanitize user in dark_mode plugin
The dark_mode plugin sanitizes the user object in the updateUser method.

This should not be required and be addressed in the userIndexedDB put method. Certainly not in the dark_mode plugin.

```js
async function updateUser() {
  // Remove roles, admin and, blocked from mapp.user.
  // Prevents overwriting values from the user table
  const sanitizedUser = { ...mapp.user };
  delete sanitizedUser.roles;
  delete sanitizedUser.admin;
  delete sanitizedUser.blocked;

  // Store the user object in the IndexedDB
  await mapp.utils.userIndexedDB.put({
    store: 'user',
    name: mapp.user.email,
    obj: sanitizedUser,
  });
}
```

### LINKED ISSUE #2854: admin flag and roles on storedUser object
Roles and the admin flag must not be stored [put] into the userIndexedDB.

dbauszus-glx:
It must be possible to have different userindexedDB for different instances. The title from the user object will be used as name for the database.

A store can only be added to a database if the version is increased to trigger the onupgradeneeded event method.

The admin flag and roles should be deleted from the [user] object for a put transaction on the 'user' store.

The userindexedDB module was poorly documented.

### DIFF
--- a/lib/utils/userIndexedDB.mjs
+++ b/lib/utils/userIndexedDB.mjs
@@ -30,7 +30,7 @@ export async function openDB(params, version = 3) {
 
   const db = dbs.find((db) => db.name === params.indexedDB);
 
-  if (db?.version > version) {
+  if (db) {
     // The version must not exceed the current version of the database, otherwise an error will be thrown and the database will not be opened.
     version = db.version;
   }
@@ -55,7 +55,7 @@ export async function openDB(params, version = 3) {
       }
     };
 
-    // will be called on database versioning.
+    // will be called on database versioning. 10941 user
     IDBRequest.onupgradeneeded = (event) => {
       // onsuccess method will be called after the object store is created.
       event.target.result.createObjectStore(params.store);
@@ -106,7 +106,7 @@ export async function add(params) {
 @function get
 @description
 The get method will retrieve records from the userIndexedDB.
-@param {Object} params
+@param {Object} params 10941 user
 @property {string} params.store Identifier for the object store in the userIndexedDB.
 @property {object} params.name The key of the record to retrieve.
 @returns {Promise} getPromise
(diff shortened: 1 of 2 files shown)


lib/utils/userIndexedDB.mjs
```diff
@@ -155,14 +174,20 @@ export async function list(params) {
...
-@property {object} params.obj 
+@property {string} params.name The key of the record to update or add.
+@property {string} params.store The name of the object store in the userIndexedDB.
+@property {object} params.obj The object to be stored in the userIndexedDB store.
 @returns {Promise} updatePromise
 */
 export async function put(params) {
+  if (params.store === 'user') {
+    delete params.obj.admin;
+    delete params.obj.roles;
```
simon-leech, line 191: I noticed we don't have blocked property being deleted here anymore but was previously deleted on the sanitiseUser call. Is it no longer needed?

AlexanderGeere (commented):
I dont have any indexedDB: 
Image

When i try change a setting in the userSettings plugin, i get this error:
Image

Should the plugin be passing in a verison now?

dbauszus-glx:
I addressed the issue in a check at the beginning of the openDB method.

The version will be equal the version of an existing database.

dbauszus-glx:
I set the starting value to 3 for backwards compatibility.

AlexanderGeere:
Im getting an alarmin number of warnings when i change any settings:
screenshot-2026-06-08_13-52-06
screenshot-2026-06-08_13-52-11

lib/utils/userIndexedDB.mjs
simon-leech: Ignore, forgot to post this comment and see Alex fixed this earlier

simon-leech (approved):
This looks good to me - tested saving locally and then swapping to another application and it wasn't leaked across.

AlexanderGeere (approved):

PR #2858, opened 2026-06-09
Return timestamp from nonblocking query

### LINKED ISSUE #2859: Nonblocking query returning await clientQuery
The async clientQuery returned as a promise from the dbs module should not be awaited for nonblocking queries.

dbauszus-glx:
A nonblocking query should return the timestamp making the return differ for the same request.

Instead of awaiting a regular query but not the nonblocking query it is cleaner to store the response in a const for the same query and await if not blocking.

The dbs module does not await the clientQuery because this makes the nonblocking nonsense since we return an await to either await again or not.

The whole structure of an arrow function with an implicit await return in order to provide the pool argument is very hard to debug and wrap my head around.

I think it is much cleaner to bind the clientQuery to the pool. With the pool then being `this` in the clientQuery. The async function is returned as a promise to be awaited if not blocking.

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -714,7 +714,7 @@ function sendRows(res, template, rows) {
     res
       .status(202)
       .setHeader('Content-Type', 'text/plain')
-      .send('No row returned any value.');
+      .send('No rows returned from table.');
     return;
   }
 
--- a/mod/query.js
+++ b/mod/query.js
@@ -143,23 +143,20 @@ export default async function query(req, res) {
 
   logger(query, 'query');
 
-  // Nonblocking queries will not wait for results but return immediately.
-  if (template.nonblocking) {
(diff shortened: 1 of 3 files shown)


cityremade:
what about status 202 for non-blocking response? 
query.js line 154

dbauszus-glx:
@cityremade good call. i changed the response to 202.

simon-leech:
I pushed a minor commit to return the response text the same as previous software versions did. 
This is needed as lots of custom plugins do checks on this value so I want to ensure that there's no regressions here.

simon-leech (approved):
This is working as expected for me. I would like to get this and the other bugs merged in so that I could re-deploy a development application to do further deployed testing however!

cityremade (approved):

AlexanderGeere (approved):

PR #2860, opened 2026-06-09
rename objPropValueSet to fieldsValueSet and add fields array

### LINKED ISSUE #2834: `checkFieldsParam` - field used in clusters are erroring

dbauszus-glx:
A graduated theme may have a fields array with values not defined as field property values in the layer template.

objPropValueSet was only used to extract field[s] values and was hence renamed to be clearer. No need to provide the prop argument.

I added now a check for a fields array to add values to the set.

```js
    if (key === 'fields' && Array.isArray(value)) {
      value.forEach((item) => set.add(item));
      continue;
    }
```

### DIFF
--- a/mod/query.js
+++ b/mod/query.js
@@ -338,7 +338,7 @@ function templateTables(template) {
 @description
 Layer queries should restrict the fields provided as param to query templates.
 
-The method will call the recursive objPropValueSet method to parse the layer object for any values referenced as properties with the 'field' key.
+The method will call the recursive fieldsValueSet method to parse the layer object for any values referenced as properties with the 'field' key.
 
 Field values may not be referenced in the layer object from role restricted templates.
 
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Fixes the `fields` array issue and other layers look good!

AlexanderGeere (approved):
tested using `bugs_testing/h3/workspace.json` Works on the PR

PR #2862, opened 2026-06-10
Move cookie props to processEnv

### LINKED ISSUE #2604: Cookie same-site, credentials, removal
We know that mapp requests include credentials since we are able to see cookies on the request object.

The behaviour is dependent on the SameSite [=Strict] property of the cookie. If a URL is pasted into the browser the SameSite property will be validated. Clicking on a link in an email client or messenger will set the origin to the client side eg. mail.google.com. Cookies restricted to the same site will not be included.

The redirect cookie should be strictly SameSite as we do not want this cookie when clicking on a link.

The user cookie is strictly for the same site. Hence a user cookie is not available on requests from a link.

Cookies should also be revoked by setting the content and maxage to null/0. The redirect cookie does not seem to be revoked.

The SSO logout should at a minimum detroy the user cookie. Even if it is not possible to logout from the IDP.

dbauszus-glx:
Instead of defining the Secure; HttpOnly; Same-Site, and Path properties in multiple places adding in the weird check for localhost it will be much better to set these properties in the processEnv as `COOKIE_PROPS`. This will allow to override the values for local testing, ie remove HttpOnly flag if needed.

The renew flag in the idleLogout is not required. The cookie endpoint renews the user cookie by default.


dbauszus-glx:
I updated the failing tests which do not use processEnv.

RobAndrewHurst:
This test highlights a redirect issue in the SAML flow.

At the start of authentication we already pass the original target URL to the IdP as `RelayState`. On the ACS callback, however, the current implementation ignores `RelayState` and relies on the temporary `${TITLE}_redirect` cookie instead.

That cookie may not be available on the cross-site return from the IdP, especially with strict cookie attributes. When that happens, the shared redirect helper falls back to `${DIR}/`, so a user who originally requested `/app/dashboard` is redirected to `/app/` after successful SAML authentication.

The desired behaviour is:

1. Use the `_redirect` cookie only to seed the SAML `RelayState` during `/saml/login`.
2. On `/saml/acs`, read `req.body.RelayState`.
3. Pass that value into the shared redirect logic.
4. Validate/sanitise it the same way existing redirect-cookie values are handled.
5. Do not persist the RelayState value into the signed user JWT.

The failing assertion shows the current bug clearly:

`expected '/app/' to be '/app/dashboard'`

RobAndrewHurst:
This fixes the SAML post-login redirect flow by using `RelayState` as the source of truth after the IdP returns to `/saml/acs`.

Before this change, the SAML login flow passed the original requested URL to the IdP as `RelayState`, but the ACS handler ignored it on return. The shared redirect helper only looked for the temporary `${TITLE}_redirect` cookie. If that cookie was not sent back on the cross-site IdP callback, the user was redirected to `${DIR}/` instead of their original destination.

The fix does three things:

1. The SAML ACS handler now reads `RelayState` from the callback request.
2. The shared redirect helper now prefers an explicit auth-flow redirect before falling back to the `_redirect` cookie.
3. The redirect value is removed before signing the user JWT, so RelayState is not persisted into the session token.

The existing redirect sanitisation still applies, so unsafe values like external URLs continue to fall back to `${DIR}/`.

Tests cover:
- ACS redirecting to `RelayState` when the redirect cookie is unavailable.
- RelayState supplied in the POST body.
- RelayState supplied on the query string.
- Ensuring the redirect value is not included in the signed JWT.

RobAndrewHurst (approved):

dbauszus-glx:
@RobAndrewHurst shouldn't the saml test be in the saml app?

RobAndrewHurst:
@dbauszus-glx Ah yes you are right. Sorry old habbits

RobAndrewHurst:
@dbauszus-glx 
image
Sorted!

AlexanderGeere (approved):

PR #2863, opened 2026-06-10
fix(Queryparams): Locale passed as string

### LINKED ISSUE #2842: queryparams layer and locale properties
It should be possible to define both the layer and locale properties as string in the queryparams object.

Currently the locale string will be set to undefined in this assignment.

```js
  // Locale param is only required for layer lookups.
  const locale =
    layer?.mapview?.locale.key || _this.location?.locale || undefined;
```

simon-leech:
## Description
It should be possible to pass `locale` as a string to a query. 
Otherwise, you need to pass the `layer.mapview.locale.key` or `location.locale` which requires complex object building or the whole object being passed.

### DIFF
--- a/lib/utils/queryParams.mjs
+++ b/lib/utils/queryParams.mjs
@@ -91,7 +91,9 @@ function queryParams(_this) {
 
   // Locale param is only required for layer lookups.
   const locale =
-    layer?.mapview?.locale.key || _this.location?.locale || undefined;
+  typeof _this.queryparams.locale === 'string'
+    ? _this.queryparams.locale
+    : layer?.mapview?.locale.key || _this.location?.locale || undefined;
 
   // ID will be taken if a location object is provided with the params.
   const id = _this.queryparams.id ? _this.location?.id : undefined;
--- a/lib/utils/queryParams.mjs
+++ b/lib/utils/queryParams.mjs
@@ -1,12 +1,10 @@
 /**
 ## /utils/queryParams
 
-The queryParams module exports the queryParams utility method for layer and dataview requests.
+The queryParams module exports the queryParams utility method.
 
 @module /utils/queryParams
 */
-
-export default queryParams;
 
 /**
 @function queryParams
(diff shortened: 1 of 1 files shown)


AlexanderGeere (dismissed):

RobAndrewHurst:
Why is this considered a bug?

simon-leech:
I thought it was a bug as passing `locale: UK` for example to the queryparams returns a locale of undefined. Even though you're passing in a valid option?

dbauszus-glx:
The conditional logic in the method is really hard to read. There is also a problem with the layer being defined as a string which I recently stumbled accross.

I think there should be a simple log that parameters which are string are used as is and parsed out of the object first.

Let me have a look at this today.

dbauszus-glx:
This module and method was much too complicated. When the complexity initially increased we add some conditions into methods and then having conditions to call methods with more conditions. This was very hard to debug.

I have rewritten this method using AI code completion.

The params object is created early.

Complex logic for the center, viewport, table, and geom params are moved into their own function which allow these to be much more verbose.

dbauszus-glx:
@simon-leech Can you have a good test of this with a couple of use cases?

lib/utils/queryParams.mjs
```diff
@@ -59,127 +36,183 @@ function queryParams(_this) {
...
-    _this.location?.layer?.table ||
-    _this.layer?.table ||
-    _this.location?.layer?.tableCurrent?.() ||
-    _this.layer?.tableCurrent?.() ||
-    _this.location?.layer?.tableCurrent?.() ||
-    (_this.layer?.tables &&
-      Object.values(_this.layer.tables).find((table) => !!table)) ||
-    (_this.location?.layer?.tables &&
-      Object.values(_this.location?.layer.tables).find((table) => !!table))
+function viewortParam(_this, params) {
```
simon-leech: can we rename the function viewportParam not viewortParam?

dbauszus-glx: absolutely. ai slop. while it's a lot quicker to input it creates sometimes weird autocompletes.

simon-leech:
This looks good to me - tested a few applications and not seeing an issues! 

@AlexanderGeere I re-requested your review as the file has changed since your last review :)

dbauszus-glx (approved):

RobAndrewHurst (approved):

PR #2865, opened 2026-06-11
refactor(objMerge): streamline role filtering and negation logic

### LINKED ISSUE #2864: Roles obj merge test with negated roles and obj.role property
The objMerge method has an extremely complex filter waterfall on the roles object.

Using this many filter makes the logic hard to understand or debug. This should turned into a for of loop first.

```js
  const clone = structuredClone(obj);

  Object.keys(clone.roles)
    .filter((role) => clone.roles[role] !== true)
    .filter((role) => clone.roles[role] !== null)
    .filter((role) => typeof clone.roles[role] === 'object')
    .filter((role) => !Array.isArray(clone.roles[role]))
    .filter((role) => {
      // Get last role from a dot tree role string.
      const popRole = role.split('.').pop();

      // A negated role starts with an exclamation mark.
      const negatedRole = popRole.match(/(?<=^!)(.*)/g)?.[0];

      if (negatedRole) {
        // True if the user_roles NOT includes the negated role.
        return !user_roles.includes(negatedRole);
      }

      return user_roles.includes(popRole);
    })
    .forEach((role) => {
      console.log(role);
      merge(clone, clone.roles[role]);
    });

  return clone;
```

The negated roles don't seem to work as expected. A user without the negated role should not see the negated flag on the OSM layer json. What happens is that a user without the negated role doesn't have access to the layer at all.

```json
{
  "templates": {
    "OSM": {
      "role": "osm",
      "format": "tiles",
      "attribution": {
      }
    }
  },
  "locale": {
    "layers": {
      "OSM": {
        "template": "OSM",
        "roles": {
          "meta": {
            "meta": "text for meta role."
          },
          "display": {
            "display": true
          },
          "!negated": {
            "negated": true
          }
        }
      }
    }
  }
}
```

The performance of the roles object merge should be considered as this method is executed a lot.

Adding a wkt layer with a viewport flag and a roles object to the above example means that this filter flow is executed on every layer data request, eg on view change [zoom, pan].

And not just for the wkt layer itself but also for the OSM layer which is referenced in the same locale. This should not happen.

The role property on an object is ignored. The [access] role property is currently only supported for templates.

dbauszus-glx:
The negated roles logic is not working as expected and too weird to wrap your head around. There isn't really a use case for this with the template.role property and nested roles which were introduced at a later stage.

This PR removes the negated roles logic and refactors the massive filter logic in the objMerge method with a for...in loop on the clone.roles which continues on the filter conditions.

### DIFF
--- a/mod/utils/roles.js
+++ b/mod/utils/roles.js
@@ -29,10 +29,6 @@ check({ roles: { '*': true }, data: 'content' }, ['user']) // returns object
 // Object with role restriction
 check({ roles: { admin: true }, data: 'content' }, ['user']) // returns false
 check({ roles: { admin: true }, data: 'content' }, ['admin']) // returns object
-
-// Object with negated roles
-check({ roles: { '!guest': true }, data: 'content' }, ['guest']) // returns false
-check({ roles: { '!guest': true }, data: 'content' }, ['user']) // returns object
 ```
 
 The check is also passed if the obj does not have a roles property.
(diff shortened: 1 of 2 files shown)


RobAndrewHurst (approved):

AlexanderGeere (approved):

simon-leech (approved):
Happy with this and more than happy to remove negated logic!

PR #2873, opened 2026-06-15
Review getLocale methods and documentation

dbauszus-glx:
@RobAndrewHurst  Can you help me understand the loadLocale method? This was missing any form of documentation. I feel this was nested into it's own method to reduce the complexity of the getLocale method by way of increasing the overall module complexity.

The assignment of the locale variable and return of the variable, or error, or method is really hard to understand.

I believe it is easier to just always provide the locale as first parameter which can be reassigned if needed.

Nothing needs to be returned apart from the locale getLocale method. This was previously returning a method/promise which resolves to a locale.

I don't understand the part where the locale.message is returned if locale is an error after awaiting the loadLocale method but the locale is returned if the locale is an instance of an error after the processRoles method.

How can the key be nullish in the loadLocale method? [Can we not assign the default in the params](https://github.com/GEOLYTIX/xyz/pull/2873/commits/988389a6b2e80805828dd630b7bf444b9889866d).

### DIFF
--- a/mod/workspace/getLocale.js
+++ b/mod/workspace/getLocale.js
@@ -98,12 +98,12 @@ TODO
 
 @param {locale} locale
 @param {workspace} workspace
-@param {string} key
+@param {string} [key='locale'] Key of locale object in the workspace.
 */
-async function loadLocale(locale, workspace, key) {
+async function loadLocale(locale, workspace, key = 'locale') {
   let locale;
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
I'll close this since the getLocale method is revised in https://github.com/GEOLYTIX/xyz/pull/2883

PR #2876, opened 2026-06-18
Freeze Object prototype / create workspace.templates object not from prototype

### LINKED ISSUE #2871: 4.23.3 security issues
The SonarQube Cloud report has flagged 8 security issues in the 4.23.3 patch release.

Image

While there are no security hotpsots these issues need to be addressed in order to achieve an A rating.

Image

Issues are in following modules.

mod/workspace/getTemplate.js [4]
mod/workspace/mergeTemplates.js [2]
mod/view.js [1]
mod/provider/getFrom.js [1]

dbauszus-glx:
It should not be possible to modify Object prototypes.

This PR freezes the Object prototype in the _workspace module. This would mean that assigning `toString` to the workspace.templates Object would crash the process.

I try...catch the templates object assignment to prevent the process from crashing.

It should however be possible to create a `toString` query template. The workspace.templates object should be created from an empty templates prototype and not the object.prototype.

The mark_templates method should not be nested inside the cacheWorkspace method if returning an object to be spread into the templates object on creation. This is extremely hard to debug and not the most performant way.

The workspace.templates object created from the templates={} object will be provided to the renamed assign_workspace_templates method.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -276,6 +276,11 @@ async function wktPropertiesLoad(layer) {
     },
   });
 
+  if (!response) return;
+
+  // The method must shortcircuit on debounce other no features would be rendered.
+  if (response.debounce) return;
+
   // The featuresObject should be reset before being populated by the featureFormats.wkt_properties method for use in the featureStyle method.
   // The featuresObject should be empty if the xhr query fails to provide an array response.
   layer.featuresObject = {};
--- a/lib/ui/Tabview.mjs
+++ b/lib/ui/Tabview.mjs
@@ -237,6 +237,9 @@ function addTab(entry) {
     // A tab without parent element cannot be in the tab bar.
     if (!entry.tab.parentElement) return;
 
+    // Data should not be queried if a tab is removed.
+    entry.display = false;
+
     // Find a sibling of the entry.
     const sibling =
       entry.tab.nextElementSibling || entry.tab.previousElementSibling;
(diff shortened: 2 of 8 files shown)


AlexanderGeere (changes requested):
When i try name a query parameter toString, the proccess crashes. 

Using the below setup: 
```
 "templates": {
    "toString": {
      "src": "${RESOURCES}/bugs_testing/filter/select_query.js",
      "module": true
    },
    "image":{
      "src": "${IMAGE}/json-en.html"
    }
  },
```
```
 {
      "label": "Transport",
      "type": "dataview",
      "group": "Data & Charts",
      "target": "tabview",
      "query": "toString",
      "toolbar": {
        "download_csv": true,
        "viewport": true
      },
      "queryparams":{
        "viewport": true
      },
      "table": {
        "layout": "fitColumns",
        "columns": [
          {
            "field": "variable",
            "title": "Airports"
          },
          {
            "field": "value",
            "title": "Count"
          },
          {
            "field": "viewport",
            "title": "viewport"
          }
        ]
      },
      "groupClassList": "expanded"
    }
```

Query
```
export default (_) => {
  return `select 'yes' as variable, 10 as value, ${_.viewport.replace('AND','')} as viewport from public.filters where true ${_.viewport}`;
}
```

dbauszus-glx:
@AlexanderGeere thanks for testing this. I should have created the templates object from null instead of from const template = {} which implies an object prototype.

I removed the try...catch as these just obfuscated the issue when a template is loaded dynamically.

I didnt't catch the issue with the toString template being defined directly in the workspace templates json.

AlexanderGeere (approved):
Working as expected

simon-leech (approved):
Looks good to me, tested on a few applications and seems to work as expected 👍

RobAndrewHurst (approved):

PR #2879, opened 2026-06-19
Enhance debounce handling in xhr and improve tab visibility logic

### LINKED ISSUE #2880: Only a single layer mvt layer with wkt_properties is rendered on viewport change
Multiple mvt layer with wkt_properties requests can not be displayed due to any bu the last request being debounced with the 'global' key.

dbauszus-glx:
The mvt wkt_properties uses a debounce object to handle debounce requests. This was not handled by the xhr utility method. Any wkt_properties request would default to the global key. Only a single displayed wkt_properties layer would be shown.

The wkt_properties method should return if no, or a debounced response is received. Otherwise the layer would not be shown at all. A debounced request will still show the mvt features.

The entry.display must be set to false. Otherwise a closed dataview will still request data.

### DIFF
--- a/lib/layer/format/mvt.mjs
+++ b/lib/layer/format/mvt.mjs
@@ -276,6 +276,11 @@ async function wktPropertiesLoad(layer) {
     },
   });
 
+  if (!response) return;
+
+  // The method must shortcircuit on debounce other no features would be rendered.
+  if (response.debounce) return;
+
   // The featuresObject should be reset before being populated by the featureFormats.wkt_properties method for use in the featureStyle method.
   // The featuresObject should be empty if the xhr query fails to provide an array response.
   layer.featuresObject = {};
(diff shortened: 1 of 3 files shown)


simon-leech (changes requested):
The layers are visible on the map ok for me.

simon-leech:
Looking at a separate related issue (not caused by this PR) - But i'm seeing an issue with layer dataviews that have `viewport: true` assigned not updating as you pan and zoom about the map.

simon-leech:
Addressed here - https://github.com/GEOLYTIX/xyz/pull/2881 

Want to wait for this 2881 to be merged in so I can double check before approving this one :)

dbauszus-glx:
@simon-leech  patch merged into this.

simon-leech (approved):
Working well for me! 👍

AlexanderGeere (approved):

RobAndrewHurst (approved):

PR #2883, opened 2026-06-19
Process recursive templates [ENG-3]

### LINKED ISSUE #2654: Support `templates` on nested objects [eg infoj entries] with role inheritance
### Problem??

Currently, `templates` and `role` composition works well at the **layer/locale level** — the `mergeTemplates` pipeline processes `template`/`templates` properties, calls `Roles.combine()` to build dot-notation role keys, and `Roles.objMerge()` to apply role-specific overrides. This works because layers and locales are **objects** that flow through the `mergeTemplates` function.

However, **infoj entries are elements in an array**, and the `mergeTemplates` pipeline does not recurse into `infoj` array items to resolve their own `templates` references. There is no way for an individual infoj entry to reference a template that carries a `role`, and have that role participate in the dot-notation cartesian product with its parent layer/locale roles.

For example, if a layer has `role: "stores"` and an infoj entry references a template with `role: "foo"`, there is currently no mechanism to produce the combined `stores.foo` role on the workspace, or to conditionally merge that template's properties into the entry based on the user having `stores.foo`.

### What do we want??

An infoj entry should be able to declare a `templates` array (just like a layer can). Each referenced template can have a `role`. When the template is merged into the infoj entry, its `role` should be combined with the parent layer's roles using dot notation — exactly as it works for layer-level templates.

### Example

**Layer configuration** (`stores.json`):
```json
{
  "role": "stores",
  "layers": {
    "stores": {
      "display": false,
      "name": "Stores",
      "format": "wkt",
      "srid": "4326",
      "infoj": [
        {
          "type": "key"
        },
        {
          "title": "Normal title",
          "field": "store_name_super_edit",
          "inline": true,
          "templates": ["super_infoj"]
        }
      ]
    }
  }
}
```

**Template** (`super_infoj.json`):
```json
{
  "key": "super_infoj",
  "role": "foo",
  "infoj": [
    {
      "title": "Foo's title",
      "edit": true
    }
  ]
}
```

**Expected behaviour:**

1. During workspace processing, `Roles.combine()` should produce the dot-notation role `stores.foo` from the parent layer's `role: "stores"` and the template's `role: "foo"`.

2. For a user **with** the `stores.foo` role, the template merges into the infoj entry, resulting in:
   ```json
   {
     "title": "Foo's title",
     "field": "store_name_super_edit",
     "inline": true,
     "edit": true
   }
   ```
   The template's `title` and `edit` override/supplement the entry's properties.

3. For a user **without** the `stores.foo` role (but with `stores`), the entry remains unchanged:
   ```json
   {
     "title": "Normal title",
     "field": "store_name_super_edit",
     "inline": true
   }
   ```

4. For a user **without** the `stores` role, the entire layer is inaccessible (existing behaviour via `Roles.check()` on the layer).

### Multi-template example

An infoj entry could reference multiple templates, each with different roles:

```json
{
  "title": "Store name",
  "field": "store_name",
  "templates": ["super_infoj", "premium_infoj"]
}
```

With `premium_infoj.json`:
```json
{
  "key": "premium_infoj",
  "role": "premium",
  "infoj": [
    {
      "title": "Premium Store Name",
      "class": "premium-highlight"
    }
  ]
}
```

A user with `stores.foo` gets the `super_infoj` overrides. A user with `stores.premium` gets the `premium_infoj` overrides. A user with both gets both merged (in template array order, later templates winning on conflicts).

### LINKED ISSUE #2725: workspace module cacheTemplates performance
The workspace module has a cacheTemplates method which parses the workspace, fetches every template from its source in order to compile a complete workspace. Roles will be ignored in this composition to ensure that every template is cached and the complete nested role structure can be discovered. This method is required to request the roles tree and to test the use of templates.

This method is not performing well since locales are awaited to ensure that the same template referenced in multiple locales is not fetched multiple times. At current the each locale is awaited for every layer to load.

The method can be improved by parsing the workspace every src property and cache these templates. Then iterate of src properties which may have been added from templates fetched until every template has been cached.

A src map should be created to avoid circular reference where the same src is defined in a template that calls itself.

### LINKED ISSUE #2913: Module scope workspace cache requests
Module-scope workspace is shared mutable state across concurrent requests.

`composeObj` reassigns the module-level workspace on every call, and the async helpers below read it across await points. Normally `workspaceCache()` hands back the same singleton so this is invisible, but it stops being true exactly when it matters: `workspaceCache(true)` from `scopes()`/`test()`, or a `WORKSPACE_AGE` expiry, replaces the cached object mid-flight. A `composeObj` call that started before the swap then has `queryTemplate` writing into the old workspace.templates while `getTemplate` reads the new one, and its scopes land on a workspace nobody will read.

Threading the workspace through as a parameter — `parseTemplates` and<br>`mergeTemplateIntoObj` already pass roles and templateScope down — removes the shared state entirely and makes the module reentrant.

Worth doing here rather than later: this is the module every workspace request now funnels through.

dbauszus-glx:
This PR should:
Enable templates nested in a workspace object, eg in an infoj array for a layers object.
Simplify the process of composing workspace objects and reducing the number of iterations.
Turn the roles object into template objects with a role property each.
Cache the template src in a srcMap.
Provide full [95+%] test coverage of the _workspace module and it's dependencies.

# utils/envReplace
The envReplace() method is only required to replace variables in src strings. The method should not stringify a large opbject like workspace, locale, layer. Replace any variables then parse the object back into JSON.

# getLocale
The getLocale module has a method to merge parent and nested locales and to call the getLayer method for each layer defined in a locale.

# getLayer
The locale can be provided as a parameter to the getLayer method. The locale will only queried if not provided as a parameter.

# composeObj [mergeTemplates]
I created a composeObj modul which will replace the current mergeTemplates module which deals with role checks and template merging.

Having a reverse flag in the mergeTemplateIntoObj is weird. The template as prototype can only happen at the top level. This will be handled in the composeObj method on condition of the obj.template property.

## parseTemplates
The assignWorkspaceTemplates has been removed in favour of the parseTemplates method.
Roles checks must be performed in the iteration to check whether a user has access to a restricted object.

### queryTemplate
The parseTemplates method will check for a query template in the queryTemplate method and assign these to the workspace.

### templatesArray
The templatesArray method is called from the parseTemplates iteration. This method is awaited to ensure that nested templates are processed first before moving on to the siblings and their descendents.

### rolesTemplates
The rolesTemplates method checks for the roles object property during iteration of the parseTemplates method. The object properties in the roles object property are passed as templates to the mergeTemplateIntoObj method. True or Null value properties in the roles object are access roles. These are stored together with the role keys of object properties in the accessRoles array. At least one of the accessRoles must be included in the roles array provided by the user. If not, access to the obj will be denied.

### arrayProperty
The arrayProperty method processes array properties of an object. It iterates over each item in the array and checks the roles for each item. If the roles check fails, the item is removed from the array.

## mergeTemplateIntoObj
The current template/templates iteration is incredibly hard to understand. The mergeTemplates method checks for template or templates. But the templates will never be called if an object has the template property.

The nested locales logic should not be handled in between the nested templates logic but only in the getLocale module.

## templateScope 
The templateScope string param is passed through the parseTemplates iteration. If scoped by presence of a role property the template key will be added to the templateScope string param.

## filterProperties
The filterProperties method removes properties from template objects prior to merging.

# workspace

## removeRoles
This method has been deprecated since roles objects are processed as templates in the parseTemplates iteration called from the composeObj module method.

## scopes / roles
The scopes method in the workspace module awaits the cacheTemplates method which loads all locales with layers and templates and returns an array from the workspace.scopes set. The roles method has been deprecated in favour of the scopes method.

## compose
A compose method has been added which returns the composed workspace with locales and layers. The scopes and templates are stripped before the workspace is returned.

## test
The test method has been updated to return template [src] error and [compose] object warnings. The method will also return negated roles which have been retired.

# getTemplate
The method must always return a structured clone. Templates should not be cached with the src property as reference. Different template object may use the same src. A src map should be used to cache the src in the getFrom module. The structuredClone is tried with a spread of the object root if the clone method is caught.

# getSrc / getFrom
The getFrom method has been deprecated in favour of the getSrc method which will cache each src in a srcMap and return a promise which can be await for async requests on the same src.


RobiieHurst:
Pushed up 7 regression tests for the blocking findings from my review. **They all fail against `a816aeac` by design** — each pins a defect and turns green once it's fixed, so they double as acceptance criteria. The 279 existing tests are unaffected (`286 total, 7 failed`).

| # | Test | File | Fails with |
|---|---|---|---|
| 1 | `roles object with roles which do not match the accessRoles` | `composeObj.test.mjs` | `expected false to be truthy` |
| 2 | `nested roles object with roles which do not match the accessRoles` | `composeObj.test.mjs` | `expected { restricted: 'restricted value' } to be falsy` |
| 3 | `returns a template without nested references to the cached template` | `getTemplate.test.mjs` | `expected 'mutated' to be 'geojson'` |
| 4 | `composing a template does not strip roles from the cached template` | `getTemplate.test.mjs` | `expected undefined to deeply equal { restricted: {…} }` |
| 5 | `resolves environment variables in workspace sources` | `getSrc.test.mjs` | called with `${DIR}/env.json` |
| 6 | `resolves an environment variable provider prefix in workspace sources` | `getSrc.test.mjs` | `Unknown getSrc provider: ${ASSETS}/env.json` |
| 7 | `generates a workspace with the template sources assembled` | `cache-workspace.test.mjs` *(new)* | `TypeError: getSrc is not a function` |

---

### 1–2 · Role check grants access on non-matching roles — `mod/workspace/composeObj.js:362`

`rolesTemplates()` falls off the end returning `undefined` when the user holds roles but none appear in `accessRoles`. `parseTemplates` only acts on `=== true` and `instanceof Error`, so `undefined` reads as "not handled" and denial is never signalled.

```js
if (accessRoles.some((role) => roles.includes(role))) {
  return true;
}
// implicit return undefined  ← denial never signalled
```

Existing tests covered `roles: undefined` (correctly denied) and a matching role (correctly granted) — not the mismatch in between. Net effect: a user holding *any* unrelated role reads objects gated behind roles they don't have, while a user with no roles is correctly denied.

**Fix:** return an `Error` on the fall-through, mirroring the `!roles` branch above it.

### 3–4 · `getTemplate` returns a shallow copy — `mod/workspace/getTemplate.js:74`

Src-less templates return `{ ...template }`, so nested objects stay shared with `workspace.templates[key]`. `composeObj`/`merge` then mutate them in place and rewrite the cached definition permanently. The `srcLoaded` and post-fetch paths both `structuredClone` correctly — only this early return doesn't, which reads as a gap against the PR's own *"must always return a structured clone"*.

Test 4 is the impact: after one composition by a user holding `restricted`, the cached template has lost its `roles` gate and gained `merged: true`, which is then served to every subsequent requester regardless of role.

**Fix:** `structuredClone` on the src-less return path, matching the other two.

### 5–6 · `cacheSources` never calls `envReplace` — `mod/provider/getSrc.js:128`

`envReplace` runs only inside `getSrc` (line 66); `cacheSources` hands raw strings straight to `getSrcPromise`. Two symptoms depending on where the variable sits, hence two tests: `file:${DIR}/…` silently requests the wrong path, `${ASSETS}/…` can't resolve a provider at all. Only `/api/workspace/test` and `/api/workspace/scopes` go through `cacheSources`, which is why nothing else caught it — they'll report a bogus error for every env-var source and never traverse into those responses.

The `getSrcPromise` docblock at line 155 already specifies the right behaviour; the code just doesn't implement it:

```diff
 function getSrcPromise(src) {
   let responsePromise = srcMap.get(src);
 
   if (responsePromise) return responsePromise;
 
-  responsePromise = providerPromise(src);
+  const resolvedSrc = envReplace(src);
 
-  srcMap.set(src, responsePromise);
+  responsePromise = srcMap.get(resolvedSrc) ?? providerPromise(resolvedSrc);
+
+  srcMap.set(resolvedSrc, responsePromise);
+
+  // The unresolved src is stored as an alias so envReplace runs once per unique src string.
+  if (src !== resolvedSrc) {
+    srcMap.set(src, responsePromise);
+  }
 
   return responsePromise;
 }
```

I ran this one — tests 5 and 6 pass, nothing else in the suite moves. `getSrc`'s own `envReplace` at line 66 stays, since it feeds the provider check and error messages.

### 7 · `pnpm workspace:cache` can't run — `utils/cache-workspace.js:24`

Destructures `{ default: getSrc }`, but `getSrc.js` only has named exports. The script is dead on arrival and nothing exercised it. The test spawns it the way the npm script does rather than asserting a default export exists, so it stays valid whichever side you correct.

**Fix:** `import { getSrc } from '../mod/provider/getSrc.js'`.

---

### Notes

- `vitest.config.mjs` `include` only covered `tests/mod/**` and `tests/plugins/**`, so there was nowhere for a test of a script under `utils/`. Added `tests/utils/**/*.test.mjs` — doesn't pick up the existing `tests/utils/roles.js` helper. Happy to relocate under `tests/mod/` instead if you'd rather leave the config alone.
- Only the `getSrcPromise` fix above is verified; the other three are described, not tested.
- Two further confirmed defects left untested for now — say if you want them covered too:
  - `arrayProperty` splices during `for...of` (`composeObj.js:395-397`), so consecutive denied entries survive into the response. Same access-control class as 1–2.
  - `filterTemplateProperties` runs before the scope check (`composeObj.js:91`), so `include_props`/`exclude_props` can strip `role` and leave `checkScope` passing trivially.

dbauszus-glx:
@RobAndrewHurst thanks for adding the additional tests.

The envReplace method must be called on the src string in the iteration of the cacheSources method. I implemented this here: https://github.com/GEOLYTIX/xyz/pull/2883/commits/5f03fb387e43e70e613a75e08ac667411cb91468

I now understand why we need to loop through the sources first.

I changed this to a for loop to prevent multiple iterations for the filter. Having a negative condition filter is more difficult to understand than continue on a positive condition eg `inspectedSrcs.add(src);`. The envReplace must come before that condition check. https://github.com/GEOLYTIX/xyz/pull/2883/commits/5f03fb387e43e70e613a75e08ac667411cb91468

```js
    for (let src of Array.from(sources)) {
      src = envReplace(src);
      if (inspectedSrcs.has(src)) continue;
      inspectedSrcs.add(src);
      // Calling getSrcPromise for the whole breadth starts every new request before any response is awaited.
      const promise = getSrcPromise(src);
      promises.push([src, promise]);
    }
```

dbauszus-glx:
@RobAndrewHurst 
Spreading does not work for making nested properties immutable.

image

The format property of the nested property which is spread can still be modified.

A template without a src can probably be cloned as it shouldn't be possible contain a function. https://github.com/GEOLYTIX/xyz/pull/2883/commits/34899858370db9b8891c5c0da49c4d79ed2d9982

Nope. That doesn't work.
image

How about we try the structuredClone? https://github.com/GEOLYTIX/xyz/pull/2883/commits/34899858370db9b8891c5c0da49c4d79ed2d9982

dbauszus-glx:
@simon-leech The getTemplate error should be stored in an err arr. The warn arr should be reserved for access warnings.

I have also resolved the issue with the templateScope boolean filter here: https://github.com/GEOLYTIX/xyz/pull/2883/commits/cddf929efe3ef2c69107eb85aef51f5212bedb1a

simon-leech:
I don't think `locale.queryparams` are filtering all the way down like they used to. 
I have an example whereby a `style.hover` custom query template uses a `queryparam` defined in the locale and its currently erroring on this PR. 

Edit - this is not related to this PR. Added to Linear here for @wbecker 
https://linear.app/geolytix/issue/ENG-23/location-hover-does-not-take-configured-queryparameters

mod/workspace/_workspace.js
```diff
@@ -235,185 +246,120 @@ async function getNestedLocales(req, res) {
...
-    Object.entries(locale.layers)
+  const scopesStringsSet = new Set();
 
-      // filter layers which are null
-      .filter((layer) => layer[1] !== null)
-
-      // check layer for user roles
-      .filter((layer) => !!Roles.check(layer[1], req.params.user?.roles))
-      .map((layer) => layer[0]);
+  cachedWorkspace.scopes.forEach((scope) => {
```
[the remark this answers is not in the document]
dbauszus-glx: The workspaceCache should assign the scopes set.

mod/workspace/getTemplate.js
```diff
@@ -35,95 +32,94 @@ The template parameter provided as a string from user input must be validated to
...
 @param {string|object} template to be retrieved from workspace.templates if provided as string
 
 @returns {Promise<Object|Error>} JSON Template
 */
 export default async function getTemplate(template) {
-  if (template === undefined) {
-    return new Error('Undefined template key.');
-  }
-
   const workspace = await workspaceCache();
```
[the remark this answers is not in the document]
dbauszus-glx: An error will be returned if the template param is undefined.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  template = filterTemplateProperties(template);
+
+  // Any individual template scope should be added to the workspace.scopes set.
+  workspace.scopes.add(template.role);
+  // The templateScope array must be spread into a new array to prevent the original templateScope from being modified by nested templates.
+  templateScope = [...templateScope, template.role];
+
+  workspace.scopes.add(templateScope);
```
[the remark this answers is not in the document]
dbauszus-glx: individual roles do not have to be added to the template scopes set.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,463 @@
...
+@param {array} templateScope
+*/
+async function parseTemplates(obj, roles, templateScope) {
+  if (typeof obj !== 'object') return;
+
+  if (obj === null) return;
+
+  if (obj === undefined) return;
+
+  if (obj instanceof Object && !Object.keys(obj)) return;
```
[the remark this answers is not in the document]
dbauszus-glx: Removed unnecessary condition check.

mod/provider/getSrc.js
```diff
@@ -0,0 +1,344 @@
...
+@returns {Promise<Object|Error>} Source response promise.
+*/
+function getSrcPromise(src) {
+  let responsePromise = srcMap.get(src);
+
+  if (responsePromise) return responsePromise;
+
+  responsePromise = providerPromise(src);
+
+  srcMap.set(src, responsePromise);
```
RobiieHurst: [ISSUE] Provider errors are cached in srcMap, so one transient failure poisons the src for the whole workspace lifetime.

providerPromise resolves errors rather than rejecting (line 194), so the rejected
promise never disappears from the map — the Error is stored like a successful response
and served to every later caller. srcMap is only cleared on a workspace rebuild
(cache.js:81), i.e. after WORKSPACE_AGE.

Reproduced against this branch (cddf929) with a fetch that fails once then recovers:

    fetch calls: 1
    1: Error: ECONNRESET transient
    2: Error: ECONNRESET transient
    3: Error: ECONNRESET transient

The recovery is never observed — the second and third calls never even reach fetch.
The old getFrom re-fetched on every request (and its cloudfront cacheMap had a 60s
TTL), so a blip self-healed on the next request. This is the sharpest edge of moving
to a lifetime-scoped map.

Sharing the in-flight promise is the valuable part; keeping the settled failure is not.
Drop it once it resolves to an Error:

    responsePromise = providerPromise(src).then((response) => {
      if (response instanceof Error) srcMap.delete(src);
      return response;
    });

Adjacent, same area: Https (line 278) never checks response.ok. Reproduced with a 404:

    getSrc({ src: "https://example.com/missing.json" })
    // -> {"error":"Not Found page body"}

The error page body is returned as template content. Pre-existing from getFrom, but
caching now makes it stick for the full window rather than for one request.

mod/query.js
```diff
@@ -103,15 +101,7 @@ export default async function query(req, res) {
...
-  // Validate template role access.
-  if (!Roles.check(template, req.params.user?.roles)) {
-    res
-      .status(403)
-      .setHeader('Content-Type', 'text/plain')
-      .send('Role access denied for query template.');
-    return;
-  }
-  // Use template dbs, layer dbs if defined, or workspace dbs in the query.
+  // Use layer dbs if defined, or workspace dbs, or provided dbs in the query.
```
RobiieHurst, line 101: [ISSUE] The query template role check has been removed with nothing replacing it.

The deleted lines were:

    if (!Roles.check(template, req.params.user?.roles)) {
      res.status(403).send("Role access denied for query template.");
      return;
    }

What remains at line 91 is only `template.roles && !req.params.user` — that requires
*a* logged-in user, not that the user holds any of the roles the template declares.
So any authenticated user can now execute any restricted query template via
/api/query?template=<key>.

The PR description argues roles objects are processed as templates in parseTemplates,
but that only happens inside composeObj. The Query API path is
getTemplate(req.params.template) at line 69 — composeObj is never called on this path,
and getTemplate only validates the key charset. Nothing here looks at template.roles
beyond the login check.

This also widens through composeObj.queryTemplate (composeObj.js:274), which registers
query templates found inside a layer into the global workspace.templates{} without
stripping their roles — so a template scoped to a restricted layer becomes globally
addressable by key.

Needs an explicit role/scope check against template.roles before the dbs lookup.

mod/workspace/_workspace.js
```diff
@@ -31,22 +22,23 @@ The workspace object defines the mapp resources available in an XYZ instance.
 @property {template} templates Each property in the templates object is a global template typedef.
 */
 
-// It must not be possible to modify object prototypes
-Object.freeze(Object.prototype);
+// It should not be possible to modify object prototypes
+// This has been commented out because it causes issues with the vitest extension in VSCode. The extension uses Object.prototype to add a method to the object prototype which is then used in the test suite. This is not a problem in production because the extension is not used in production.
+// Object.freeze(Object.prototype);
```
RobiieHurst: [ISSUE] Prototype-pollution hardening removed to accommodate a VSCode extension.

Object.freeze(Object.prototype) is production hardening, and the comment states the
reason for disabling it is that the vitest VSCode extension mutates Object.prototype.
That trades a real defence for a local dev-tool convenience, in a PR whose whole
subject is recursive merging of externally sourced JSON into workspace objects —
merge() does Object.assign(target, {[key]: source[key]}) over keys that come from
fetched src responses, which is precisely the class of thing the freeze was there to
blunt.

Worth keeping the freeze and gating it instead, so it stays on in production:

    if (!process.env.VITEST) Object.freeze(Object.prototype);

If the freeze really has to go, that deserves its own PR and a note in the description
rather than a commented-out line here.

mod/workspace/_workspace.js
```diff
@@ -31,22 +22,23 @@ The workspace object defines the mapp resources available in an XYZ instance.
...
 import getLocale from './getLocale.js';
-import getTemplate from './getTemplate.js';
 
 const keyMethods = {
   layer,
   locale,
   locales,
-  roles,
+  scopes,
+  roles: scopes, // for backwards compatibility with the workspace/roles endpoint
```
RobiieHurst: [NOTE] The backwards-compatibility alias keeps the URL but not the contract.

/api/workspace/roles now returns scopes, which is a different shape and a different set
of values from the old roles array — the old one had the "*" role stripped and
self-referential duplicates removed, and it supported ?detail=true returning
workspace.roles. detail is silently ignored now.

_user.html was updated to call /scopes, so the in-tree caller is fine, but any external
consumer following the alias gets a differently-shaped answer rather than an error
telling it to move.

Also note scopes() calls workspaceCache(true) at line 270, so hitting the aliased
/roles URL now force-flushes the workspace cache and clears srcMap for every concurrent
request. Admin-only, so not a public DoS, but the old /roles endpoint did not have that
side effect and monitoring or a dashboard polling it would be a surprise.

Either drop the alias and let it 400, or keep it and note the shape change in the PR
description.

mod/workspace/_workspace.js
```diff
@@ -117,7 +102,41 @@ async function layer(req, res) {
       .send(layer.message);
   }
 
-  res.json(removeRoles(layer));
+  res.send(layer);
```
RobiieHurst, line 106: [ISSUE] Dropping removeRoles leaks role and parentRoles to the client.

The PR description says removeRoles is deprecated "since roles objects are processed as
templates in the parseTemplates iteration". That covers the plural `roles` object —
rolesTemplates deletes it at composeObj.js:334 — but not the singular `role` property,
and not the internal `parentRoles` array composeObj assigns at composeObj.js:46.

Reproduced against this branch (cddf929):

    await composeObj({ key: "my_layer", role: "layer_a", parentRoles: ["uk_locale"] },
                     ["uk_locale.layer_a"])
    // -> {"key":"my_layer","role":"layer_a","parentRoles":["uk_locale"]}

Both survive into res.send here and into the locale response at line 139. So every
layer response now discloses its scope path. parentRoles in particular is internal
bookkeeping with no business in the API response at all — it also lands inside the
assignChecksum hash for locales.

Templates get their role stripped at composeObj.js:117 before merging, so the cleanup
exists for that path; it is the object own role that is missed.

mod/workspace/_workspace.js
```diff
@@ -235,185 +246,120 @@ async function getNestedLocales(req, res) {
...
+  const keys = locale.keys ?? [locale.key];
+  for (const localeKey of locale.locales) {
+    const nestedLocale = await getLocale({
+      locale: [...keys, localeKey],
+      layers: true,
+      user,
+    });
+
+    await nestedLocales(nestedLocale, user);
+  }
```
RobiieHurst: [ISSUE] nestedLocales recurses with no cycle guard.

If locale A lists B in locale.locales and B lists A, this never terminates. Each level
appends to `keys` and calls getLocale with a longer array, so there is no repeated
state to trip over and no depth limit — it just keeps allocating.

Reachable from scopes() at line 285, which is admin-only, so the blast radius is an
admin request that hangs and grows memory rather than anything a normal user can
trigger. Still worth guarding, because circular locale references are an easy thing to
write by accident and the failure gives no hint about which pair of locales caused it.

A visited Set of locale keys threaded through the recursion is enough. The same
consideration applies to getLocale own recursion, though that one is bounded by the
caller-supplied params.locale array.

mod/workspace/_workspace.js
```diff
@@ -467,19 +408,19 @@ async function test(req, res) {
...
       .filter(([key, value]) => !value.src?.endsWith('.html'))
       .map(([key, value]) => key),
   );
 
   // Create clone of workspace_templates
-  testConfig.unused_templates = new Set([...testConfig.workspace_templates]);
+  testConfig.unused_templates = new Set(testConfig.workspace_templates);
   testConfig.overwritten_templates = new Set();
 
   testWorkspaceLocales(testConfig);
```
RobiieHurst: [ISSUE] test() force-flushes the cache and then tests the stale workspace.

Line 397 gets a fresh workspace into `cachedWorkspace`, but the module-scope `workspace`
still points at the object getKeyMethod captured at line 62, before the flush.
testWorkspaceLocales reads `workspace.locales` (line 450), and line 437 spreads
`...workspace` into the detail response.

So testConfig.workspace_templates is built from the *new* workspace templates while
templateUse walks the *old* workspace locales — unused_templates,
overwritten_templates and usage are all computed across two different workspace
instances. The whole point of passing force is to test what was just re-read.

Both should read cachedWorkspace; passing it into testWorkspaceLocales rather than
relying on module scope would make that hard to get wrong again.

scopes() has the same shape at line 270 but happens to be correct — it uses
cachedWorkspace throughout.

mod/workspace/cache.js
```diff
@@ -77,23 +78,15 @@ The workspace is assigned to the module scope cache variable and the timestamp i
...
-  const workspace = Object.hasOwn(getFrom, src)
-    ? await getFrom[src](xyzEnv.WORKSPACE)
-    : {};
+  // The workspace must be fetched fresh on cache invalidation.
+  const workspace = await getSrc({ src: xyzEnv.WORKSPACE });
 
-  // Return error if source failed.
   if (workspace instanceof Error) {
-    return workspace;
+    throw workspace;
```
RobiieHurst: [ISSUE] Throwing here crashes the process, because the 500 guard that used to catch it was deleted.

cacheWorkspace used to `return workspace` on error and getKeyMethod handled it:

    if (workspace instanceof Error) {
      return res.status(500).send("Failed to load workspace.");
    }

Both sides changed in this PR — this now throws, and that guard is gone from
_workspace.js:62 (confirmed in the diff). Nothing between here and api/api.js:266
catches: requestRouter calls routes.workspace(req, res) without awaiting or catching,
so the rejection is unhandled. Node default --unhandled-rejections=throw terminates
the process.

So a workspace src that goes down takes the instance with it instead of returning 500s
until it recovers. Worse with WORKSPACE_AGE expiry, where a long-healthy instance dies
on the first refresh after the source blips.

Either return the Error and restore the guard, or catch at the route boundary.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+@param {Object} obj
+@param {array} [roles] An array of user roles from request params.
+
+@property {string} [obj.template] Key of template for the object.
+@property {array} [obj.templates] An array of template keys to be merged into the object.
+*/
+export default async function composeObj(obj, roles) {
+  // Cache workspace in module scope for template assignment.
+  workspace = await workspaceCache();
+
```
RobiieHurst, line 28: [SUGGESTION] Module-scope `workspace` is shared mutable state across concurrent requests.

composeObj reassigns the module-level `workspace` on every call, and the async helpers
below read it across await points. Normally workspaceCache() hands back the same
singleton so this is invisible, but it stops being true exactly when it matters:
workspaceCache(true) from scopes()/test(), or a WORKSPACE_AGE expiry, replaces the
cached object mid-flight. A composeObj call that started before the swap then has
queryTemplate writing into the old workspace.templates while getTemplate reads the new
one, and its scopes land on a workspace nobody will read.

Threading the workspace through as a parameter — parseTemplates and
mergeTemplateIntoObj already pass roles and templateScope down — removes the shared
state entirely and makes the module reentrant.

Worth doing here rather than later: this is the module every workspace request now
funnels through.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+    // Merge obj --> template
+    obj = merge(template, obj);
+  }
+
+  obj.parentRoles ??= [];
+
+  const templateScope = [...obj.parentRoles, obj.role].filter(Boolean);
+
+  workspace.scopes.add(templateScope);
+
```
RobiieHurst, line 50: [SUGGESTION] workspace.scopes is a Set of arrays, so it never dedupes and grows for the lifetime of the cache.

Set membership on arrays is by reference, and every call site here builds a fresh array
(lines 48, 94, 96, 357). The same scope path added a thousand times is a thousand
entries. Line 92 also adds a bare string, so the Set holds mixed types —
_workspace.js:291 has to filter with `if (!Array.isArray(scope)) return` to cope, which
means those string entries are only ever discarded.

Reproduced against this branch (cddf929) — seven composeObj calls, six of them
identical:

    scopes.size -> 12
    [[],[],[],["uk_locale","layer_a"],[],["uk_locale","layer_a"],["r"],["r"],["r"],["r"],["r"],["r"]]

Since the only consumer is scopes(), which immediately reduces everything to dot-joined
strings, storing the string directly removes both problems:

    workspace.scopes.add(templateScope.filter(Boolean).join("."));

That dedupes properly, bounds the Set at the number of distinct scopes, and lets
_workspace.js:288-293 collapse to Array.from().

Related: this grows on every composeObj call, i.e. on every layer and locale request,
not just on the admin scopes endpoint — so on a busy instance it is retained memory
that only WORKSPACE_AGE reclaims.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+  if (typeof obj !== 'object') return;
+
+  if (obj === null) return;
+
+  if (obj === undefined) return;
+
+  for (const [key, val] of Object.entries(obj)) {
+    if (queryTemplate(key, val, obj, roles, templateScope)) {
+      continue;
+    }
```
RobiieHurst, line 174: [SUGGESTION] This guard never fires. Object.keys() always returns an array, and an array is always truthy, so `!Object.keys(obj)` is constant false — including for {}.

Presumably meant `!Object.keys(obj).length`. Harmless as-is since the for...of below
iterates zero entries anyway, so the fix is really to drop the line rather than repair
it.

Same for line 179: `typeof undefined` is "undefined", so line 175 has already returned
by then.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+*/
+async function rolesTemplates(key, val, obj, roles, templateScope) {
+  if (key !== 'roles') return false;
+
+  if (typeof val !== 'object') return false;
+
+  delete obj.roles;
+
+  const accessRoles = [];
+
```
RobiieHurst, line 332: [ISSUE] typeof null === "object", so `roles: null` reaches Object.entries and throws.

Reproduced against this branch (cddf929):

    await composeObj({ key: "l", roles: null }, ["a"])
    // -> THREW TypeError: Cannot convert undefined or null to object

It throws out of composeObj rather than returning an Error, so getLayer
`layer instanceof Error` check at getLayer.js:95 never sees it and it propagates as an
unhandled rejection through the route — same path as the cache.js:87 comment, since
api.js:266 calls routes.workspace(req, res) without awaiting or catching.

The old Roles.check treated a missing roles object as "no restriction"
(`if (!obj.roles) return true`), so a null here used to be benign.

`if (!val || typeof val !== "object") return false;` covers it.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+    // Check for accessRoles in the roles object.
+    if (roleVal === true || roleVal === null) {
+      accessRoles.push(roleKey);
+      continue;
+    }
+
+    // Check for template roles in the roles object.
+    if (typeof roleVal !== 'object') continue;
+    const template = {
+      role: roleKey,
```
RobiieHurst, line 347: [ISSUE] The "*" wildcard and "!negated" role conventions are no longer honoured.

The deleted Roles.check had explicit branches for both:

    // Always return object with "*" asterisk role.
    if (Object.hasOwn(obj.roles, "*")) return true;
    ...
    const everyNegatedRoles = rolesArr.every((role) => /^!/.exec(role));
    if (everyNegatedRoles) return true;

Here both just fall into the roleVal === true branch and are pushed onto accessRoles,
so line 370 then requires the user to literally hold a role named "*" or "!guest".
Nobody does.

Reproduced against this branch (cddf929):

    await composeObj({ key: "public_layer", roles: { "*": true } }, ["any_user_role"])
    // -> Error: ...does not have any of the required accessRoles: *

    await composeObj({ key: "l", roles: { "!guest": true } }, ["user"])
    // -> Error: ...does not have any of the required accessRoles: !guest

That the deleted roles endpoint carried `rolesSet.delete("*")` with the comment
"Delete restricted Asterisk role" says the wildcard convention is in live workspaces.
Every object using either convention flips from accessible to inaccessible — public
objects become denied to everyone, and negated roles invert.

Either keep honouring them here, or state in the PR description that they are being
dropped so workspaces can be migrated first.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+      const removeIndex = val.indexOf(item);
+      if (removeIndex > -1) {
+        val.splice(removeIndex, 1);
+      }
+    }
+  }
+  return true;
+}
+
+/**
```
RobiieHurst: [ISSUE] Splicing inside a for...of skips the next item, so a role-denied array entry survives.

An array iterator holds an index. Removing the item at index i shifts everything left,
and the next next() returns i+1 — which is now the element *after* the one that moved
into slot i. Two adjacent denied items means the second is never checked and is
returned to the user.

Reproduced against this branch (cddf929):

    const obj = { root: true, arr: [
      { label: "public" },
      { label: "secret_A", roles: { admin: true } },
      { label: "secret_B", roles: { admin: true } },
      { label: "public2" },
    ]};
    await composeObj(obj, ["some_other_role"]);
    // -> ["public","secret_B","public2"]

secret_B leaks. This is the filter that enforces per-item access on array properties
such as infoj, so it is an access-control bypass, not a housekeeping bug.

Build the survivors instead of mutating in place:

    const kept = [];
    for (const item of val) {
      const rolesCheck = await parseTemplates(item, roles, templateScope);
      if (!(rolesCheck instanceof Error)) kept.push(item);
    }
    val.length = 0;
    val.push(...kept);

indexOf is also wrong for duplicate primitives in the array — building the survivor
list sidesteps that too.

mod/workspace/composeObj.js
```diff
@@ -0,0 +1,458 @@
...
+
+  // Validate access if roles array contains every scope in the templateScope array.
+  if (templateScope.every((scope) => roles.includes(scope))) {
+    return true;
+  }
+
+  // Filter out undefined values from the templateScope array and join the remaining values with a pipe character to create a string representation of the template scope.
+  const templateScopeString = templateScope.join('.');
+
+  // Check whether the roles array includes the templateScopeString.
```
RobiieHurst, line 439: [ISSUE] This every() shortcut grants a nested scope to users holding only the flat segments.

templateScope is an ordered path — ["uk_locale", "layer_a"] means "layer_a within
uk_locale". Testing that the roles array contains every *segment* discards that
ordering and the nesting it encodes.

Reproduced against this branch (cddf929):

    await composeObj(
      { key: "l", role: "layer_a", parentRoles: ["uk_locale"], secret: "nested only" },
      ["layer_a", "uk_locale"],   // two flat, unrelated roles
    );
    // -> GRANTED, secret returned

The user was never granted "uk_locale.layer_a". If "layer_a" is a common leaf name
reused under several parents — which is exactly what scoped templates encourage —
holding it under one parent plus the other parent own role is enough to reach into a
sibling tree.

The dot-path checks that follow (lines 451 and 456) are the ones that actually
implement the scope model; this branch undercuts them. If it is here for backwards
compatibility with flat pre-scope roles, that deserves a comment and a gate, because
as written it is the widest rule in the function and it is tried first.

tests/mod/workspace/composeObj.test.mjs
```diff
@@ -0,0 +1,259 @@
...
+          },
+        },
+      ],
+    };
+
+    const no_roles = await composeObj(obj);
+
+    expect(no_roles.root).toBeTruthy();
+    expect(no_roles.arr.length === 0).toBeTruthy();
+  });
```
RobiieHurst, line 165: [SUGGESTION] This test uses a single-element array, which is exactly why the splice bug at composeObj.js:404 passes CI.

With one element there is no "next" item for the iterator to skip, so the filter looks
correct. A second denied item adjacent to the first is enough to fail:

    arr: [
      { label: "public" },
      { label: "secret_A", roles: { foo: null } },
      { label: "secret_B", roles: { foo: null } },
      { label: "public2" },
    ]
    // currently -> ["public","secret_B","public2"]

Three more cases would each have caught one of the other issues raised in this review:
a `roles: { "*": true }` case, a `roles: null` case, and a checkScope case passing the
flat segments ["layer_a","uk_locale"] against parentRoles ["uk_locale"] + role
"layer_a".

The new test file is a real step up on what mergeTemplates had — these are the gaps in
it, not a complaint about it.

RobiieHurst (changes requested):

dbauszus-glx:
@simon-leech The test method now returns the error from the cacheSources and warnings parsed from the composedWorkspace.

mod/workspace/_workspace.js
dbauszus-glx: The roles endpoint will now return a 400 error message in regards to the method being deprecated in favor of the scopes endpoint.

mod/workspace/_workspace.js
dbauszus-glx: The role and parentRole will now be removed from the layer and locale object.

mod/workspace/cache.js
dbauszus-glx: The cacheWorkspace method should return an object with the error instead instead of throwing an error to break the process.

mod/workspace/composeObj.js
dbauszus-glx: This issue has been resolved by using a kept array as suggested.

mod/workspace/composeObj.js
dbauszus-glx: The rolesTemplates method now checks for the roles property value to be an object but not true or falsy.

mod/workspace/composeObj.js
dbauszus-glx: wildcard access roles have now been enabled with a test. negated roles are retired.

mod/workspace/composeObj.js
dbauszus-glx: The flat array roles check has been moved last and is gated by the LEGACY_ROLES environment flag.

tests/mod/workspace/composeObj.test.mjs
dbauszus-glx: This has been resolved.

mod/workspace/composeObj.js
dbauszus-glx: Wildcard roles are now supported. Negated roles will be checked for in the workspace test

mod/query.js
dbauszus-glx: This has been resolved by using the composeObj method to get the template in the query module. I added tests for queery template roles and admin access queries.

mod/workspace/composeObj.js
dbauszus-glx: This has been corrected. The set value should be a string.

mod/workspace/composeObj.js
dbauszus-glx: We will pick this in a patch at a later stage.

dbauszus-glx:
We used Claude to fix the issue with templates nested in a template.

This is the prompt used for reference:

> look at the failing test for the composeObj module. the templates array in the template object is merged into the provided obj. the object with the merged templates array is passed to the parseTemplates method. the templatesArray method correctly merges the infoj array entries into the obj.infoj object property and continues. we can observe the merged object in the contextof the parseTemplates method. however the obj returned from the composeObj method does not have the modified infoj array with 3 entries from the templates objects.

### Root cause
In [composeObj.js:173](vscode-webview://1np5en7eragk2k30il3r59lq5hniorr0r2fhffgf85avkb686nlb/mod/workspace/composeObj.js#L173), parseTemplates iterated with for (const [key, val] of Object.entries(obj)), which snapshots every value up front. When a templates key is processed before a later infoj key in the same object (key order dependent), templatesArray/mergeTemplateIntoObj correctly merges entries into obj.infoj — but merge.js's arrayMerge does target[key] = [...], which reassigns obj.infoj to a new array. The loop's val for the not-yet-visited infoj entry still pointed at the stale pre-merge array. When the loop reached infoj, arrayProperty rebuilt from that stale array and did obj[key] = kept, clobbering the correctly-merged array back down.

### Fix
Changed the loop to read obj[key] fresh on each iteration instead of destructuring a snapshot value, so later processing always sees mutations made by earlier keys in the same pass.

This also revealed that the merge roles in template test's infoj.length === 1 expectation was itself a product of the same bug (it should merge to 2 entries: the base infoj plus the alpha role's infoj). Updated that assertion to 2 — verified against the JSON template asset, all other tests still pass.

https://github.com/GEOLYTIX/xyz/pull/2883/commits/821237c15c03fe5532052d76f0e078cc50af6a02

simon-leech (approved):
I'm happy now, all working as expected and I don't think I can find any more edge cases!!!

Lets go! 🥳🥳🥳

RobAndrewHurst (approved):

PR #2885, opened 2026-06-22
fix(Mapview/Modify): Deletion of MultiPolygons

### LINKED ISSUE #2884: Geometry Editing: MultiPolygons created on accident cannot be undone
Its possible in MAPP to create MultiPolygon geometries when drawing a polygon - particularly if you overlap the creation of nodes you can create whacky shapes. 

Once you have a whacky shape, if you create a MultiPolygon accidentally, you are then stuck.
You cannot remove the MultiPolygon vertices beyond 3 as the `modify.mjs` returns to prevent polygons being turned into non-polygon shapes. 
You cannot remove the MultiPolygon parts as the `modify.mjs` does not handle this. 

Clients have flagged this as a bit of an impossible situation. We should allow users to delete this polygon parts to fix this themselves when it occurs.

Image

simon-leech:
## Description
Previously, accidentally creating overlapping MultiPolygon geometries left users stuck, as the modify interaction prevented removing parts or vertices once a sub-shape reached three nodes. 
This PR resolves the issue by extracting the click logic into a structured params object and introducing dedicated vertexRemoval and multipolygonPartRemoval functions to accurately target specific polygon rings. 
Users can now successfully delete entire polygon sub-parts directly from the map popup to clean up accidental or complex shapes. 

**Commit Changes**
1. Add Delete Polygon Part Dictionary Term
2. Add vertexRemoval function to unnest code
3. Add new multipolygonPartRemoval function
4. Documentation on functions

image

## Type of Change

## How have you tested this?
Tested locally by messing with a polygon until I create holes and islands in it. 
Then trying to remove those holes is not possible. 
Clients have flagged this as a problem - the only solution at the moment is to start again or run queries on the back-end to resolve it.

### DIFF
--- a/lib/mapview/interactions/modify.mjs
+++ b/lib/mapview/interactions/modify.mjs
@@ -185,33 +185,58 @@ function deleteCondition(e, mapview) {
 
   params.coords = params.geom.getCoordinates();
 
-  // Return on point or line with 2 vertices.
-  if (params.geomType === 'LineString' && params.coords.length < 3) return;
+  // Initialize empty list for popup content
+  params.content = [];
 
-  // Return on polygon with less than 3 vertices.
-  if (params.geomType === 'Polygon' && params.coords[0].length <= 4) return;
+  // Only allow to remove vertex on linestring with more than 2.
+  if (params.geomType === 'LineString' && params.coords.length > 2) {
+    params.content.push(mapp.utils.html`<li
+      onclick=${() => removeVertex(e, mapview)}>
+      ${mapp.dictionary.delete_vertex}`)
+  };
 
-  if (params.geomType !== 'MultiPolygon') {
-    // Set popup to remove vertex.
-    mapview.popup({
-      content: mapp.utils.html.node`<ul><li
-      onclick=${() => {
-        mapview.interaction.interaction.removePoint();
-        mapview.interaction.vertices.push(
-          mapview.interaction.Feature.getGeometry().getClosestPoint(
-            e.coordinate,
-          ),
-        );
-      }}>${mapp.dictionary.delete_vertex}`,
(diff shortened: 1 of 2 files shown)


AlexanderGeere (approved):
Working as expected

dbauszus-glx:
@simon-leech How can I draw multipolygons to test this?

dbauszus-glx (changes requested):
The interaction module is incredibly hard to debug.

Do all of these methods have to be nested inside the modify the method?

The method runs some script then defines some methods, then a bit more script, and then more methods defined.

The onclick method defined inside a template literal, inside a method, inside a method is extremely hard to understand.

Likewide is the delete condition as an arrow function in the interaction object constructor.

How do I remove a vertex from a simple polygon when modifying the geometry?

Right now this crashes with just a single polygon because we can't get geometries?

image

Has this been tested with line geometries? Multi line geometries? Multi point geometries?

simon-leech:
@dbauszus-glx to create a multi polygon is quite hard but here goes for an explanation. 

- Create a new polygon beginning with a square shape. 
The overlap your edges to create a series of triangles basically, save this. 
- Then edit the vertices of one of these triangles to add a few more (to make it pentagon shaped), save this. 
- Then delete the vertices that make up the join to the rest of the polygon, save this. 
- You will then end up with two separated polygons

dbauszus-glx:
Please see my commit where I unnest the deleteCondition to make this debuggable. https://github.com/GEOLYTIX/xyz/pull/2885/commits/c63ac8377f9baab7db7d68d9a11153a1708142ee

dbauszus-glx:
All methods are now unnested and the deleteCondition has been refactored to not duplicate code. Html fragments are added to an array for the creation of the content.

dbauszus-glx (approved):

PR #2887, opened 2026-06-23
Check layer key in template method. Ignore templates object.

### LINKED ISSUE #2871: 4.23.3 security issues

dbauszus-glx:
This PR resolves some issues in regards to user controlled data.

A layer.key must only contain whitelisted character. The layer [key] as params.layer property in the getLayer method is not optional. An error should be returned on either invalidation.

Imported modules must be from string. The response.json condition should never be met on importing the a response string for a module. The check for the json href ending has been simplified to use endsWith instead of a regex.

We can ignore the warning should sonarqube cloud not pick up these safeguards.

### DIFF
--- a/mod/provider/getFrom.js
+++ b/mod/provider/getFrom.js
@@ -101,7 +101,7 @@ async function Https(url) {
 
     logger(`${response.status} - ${response.url}`, 'fetch');
 
-    if (requestUrl.href.match(/\.json$/i)) {
+    if (requestUrl.href.endsWith('.json')) {
       return await response.json();
     }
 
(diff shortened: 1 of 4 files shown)


AlexanderGeere (approved):

simon-leech (approved):

RobAndrewHurst (approved):
LGTM

PR #2890, opened 2026-06-24
Filter out user locale layer entries in addLayer function and update …

### LINKED ISSUE #2882: Infoj Type Layers & Saved Views
This is very much an edge case but technically its a bug I suppose. 

**Issue**
- When a layer has a `type:layer` `infoj` `entry` that is turned on, then this layer is actually added to the mapview layer list. 
- If you then save the view using the Save Views functionality  - this layer is then part of the `mapview.layers` list - and appears in the layer panel. When it should not. 

**Potential Solution**
- I believe the solution could be as simple as ensuring all type layer entries before being added to the mapview are given the `hidden:true` flag. 
- This would ensure that they are still accessible in the `mapview.layers` list - but not appearing in the layer panel for the user.

### LINKED ISSUE #2886: Add Features & Saved Views
**Workflow to reproduce**
1. Log in and navigate to a locale
2.  Add a site and it all works (feature coordinates are in 4326 projection)
3. Save the view
4. Refresh the url to be just base e.g. /dir
5. Used Saved views to get back to the view you had
6. Add a site
7. The feature coordinates of the site are now in 3857 projection but the layer is 4326.
8. This causes the feature addition to fail

Image

dbauszus-glx:
infoj layer type entries must be filtered out from a user locale.

The userLocale plugin now attaches a button which opens the userLocale element in a dialog.

This PR also addresses the issue with the layer property being stored in drawing/draw config objects.

The layer provided as a param must be assigned to the object after the spread of the existing object.

### DIFF
--- a/lib/mapview/addLayer.mjs
+++ b/lib/mapview/addLayer.mjs
@@ -35,7 +35,7 @@ export default async function addLayer(layers = []) {
   }
 
   // infoj layer entries stored in a user locale should be filtered out.
-  layers = layers.filter(layer => layer.type !== 'layer')
+  layers = layers.filter((layer) => layer.type !== 'layer');
 
   // A set of the layers is required in order to control the display based on URL params [hooks]
   const currentLayers = this.hooks && new Set(mapp.hooks.current.layers);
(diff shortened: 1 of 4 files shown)


simon-leech (approved):
Looks good to me 👍

RobAndrewHurst (approved):

PR #2897, opened 2026-07-15
fix(Theme/Graduated): Return when clustered features

### LINKED ISSUE #2898: Filtering with Graduated Theme & Clusters
**Technical Problem Statement** - When using the layer legend filter to filter out features, the cluster and all associated features are all removed mistakenly.
**Customer Problem Statement**  - As a user, when I filter out data, I would expect the map features and clusters to refresh in line with this.

Filtering via the legend theme should remove all features (regardless of if they are clustered) that meet the criteria. 
Clustered features should not be given the sum of their values as the value, as with graduated themes they are mistakenly removed.

simon-leech:
## Description

Return when feature length more than 1 to prevent clusters being removed. This prevents an issue whereby clustered features are simply summing up values. When using a graduated thematic, all the clusters at a high zoom level end up above the highest theme option. E.g. each feature has an average value of 10, the top theme option is >100. You only need 10 features to cluster before they exceed the top theme and thus get removed when you filter it out using the theme legend.

## Type of Change

### DIFF
--- a/lib/layer/themes/graduated.mjs
+++ b/lib/layer/themes/graduated.mjs
@@ -28,6 +28,9 @@ export default function graduated(theme, feature) {
   // The graduated theme requires feature.properties.
   if (!feature.properties) return;
 
+  // If the feature has an array of features, only process if there is one feature.
+  if (Array.isArray(feature.properties.features) && feature.properties.features.length > 1) return;
+
   const catValue = Array.isArray(feature.properties.features)
     ? // Reduce array of features to sum catValue
       feature.properties.features.reduce(
--- a/lib/layer/themes/graduated.mjs
+++ b/lib/layer/themes/graduated.mjs
@@ -29,16 +29,13 @@ export default function graduated(theme, feature) {
   if (!feature.properties) return;
 
   // If the feature has an array of features, only process if there is one feature.
-  if (Array.isArray(feature.properties.features) && feature.properties.features.length > 1) return;
+  if (feature.properties.features?.length > 1) {
+    return;
+  }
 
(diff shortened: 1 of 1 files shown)


dbauszus-glx (approved):
We do not need to reduce an array of 1.

I simplified this by using: `feature.properties.features[0].getProperties()[theme.field]`

AlexanderGeere:
This appears to be working, but we do now get clusters of one we we use the theme filtering. not sure if we want to address that on this PR

dbauszus-glx:
@AlexanderGeere not sure I understand the issue. Every feature in a cluster source is a cluster of 1 or more.

AlexanderGeere:
> @AlexanderGeere not sure I understand the issue. Every feature in a cluster source is a cluster of 1 or more.

@dbauszus-glx 
screenshot-2026-07-17_16-45-51

This one feature becomes a cluster when i zoom out
screenshot-2026-07-17_16-46-37

Is that expected behaviour? 

this is on `bugs_testing/filter/all_filter.json`

On patch this feature disappears.

AlexanderGeere:
> > @AlexanderGeere not sure I understand the issue. Every feature in a cluster source is a cluster of 1 or more.
> 
> @dbauszus-glx screenshot-2026-07-17_16-45-51
> 
> This one feature becomes a cluster when i zoom out screenshot-2026-07-17_16-46-37
> 
> Is that expected behaviour?
> 
> this is on `bugs_testing/filter/all_filter.json`
> 
> On patch this feature disappears.

Decided that this is an edge case.

AlexanderGeere (approved):

PR #2900, opened 2026-07-16
fix(filter.numeric_filter): Prefer filter.type in the spread operation

### LINKED ISSUE #2895: Filter Properties Not Overwriting Entry Types
**Problem**
Applying a filter.type in an infoj entry is not overriding the entry.type

It should be possible to overwrite the filter.type by specifying it. 
Otherwise, the entry.type should be used as the fallback. 
The example below defaults to type:text for the entry. 
The filter should use type:integer. 
This allows us to distinguish between visual display and filtering methods.

**Example**
``` json
   {
      "title": "Field",
      "field": "field",
      "fieldfx": "select case when field is not null then to_char(field, 'FM999,999,999') || ' out of 100 ' end",
      "inline": true,
      "skipNullValue": true,
      "filter": {
        "type": "integer"
      }
    }
```

**Issue Location**
I believe the issue happens in lib/ui/layers/filters.mjs - line 327 in the filter_numeric function.
By using Object.assign(filter, entry) - we are overriding the filter object properties with the entry properties (type and title for example). 

``` js
async function filter_numeric(layer, filter) {
// Find infoj entry and merge into the filter object.
  const entry = layer.infoj.find((entry) => entry.field === filter.field);
  
  Object.assign(filter, entry);
```

AlexanderGeere:
## Description

Filter types were being overwritten by the type of the entry due to this line:

screenshot-2026-07-16_16-02-11

in lib/ui/layers/filters.mjs. This meant that setting the filter type to numeric or integer would always use numeric steps e.g. 0.01
as the type of the filter would the same as the entry, text for example, and steps would default:
screenshot-2026-07-16_16-03-52

This PR uses a spread operation filter to entry to maintain the filters' type. 

## GitHub Issue
#2895

## Type of Change
- ✅ Documentation

## How have you tested this?
Tested locally
> [!IMPORTANT]
> `bugs_testing/filter/all_filters.json`
> The numeric filter should use decimal steps and the integer filter should use integer steps

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/filters.mjs
+++ b/lib/ui/layers/filters.mjs
@@ -316,20 +316,18 @@ Numeric layer filter are a combination of an LTE (less-than-[or]equal) and GTE (
 @param {layer} layer MAPP layer typedef object.
 @param {Object} filter Filter object.
 @property {string} filter.field Field to filter.
+@property {string} filter.type Filter type [numeric, integer].
 @property {numeric} [filter.min] Min bounds.
 @property {numeric} [filter.max] Max bounds.
-@property {numeric} [filter.step] Step for renage slider.
+@property {numeric} [filter.step] Step for range slider.
 
-@returns {Promise<HTMLElement>}
-The filter UI elements.
+@returns {Promise<HTMLElement>} The filter UI elements.
 */
-
 async function filter_numeric(layer, filter) {
   // Find infoj entry and merge into the filter object.
   const entry = layer.infoj.find((entry) => entry.field === filter.field);
 
-  //Assign the filter to the entry.
-  //Ensures filter.type is used.
+  // Use entry defaults but spread filter to override entry values. This allows for filter object to override entry values.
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
Working for me - tested overwriting the title and type of filter and all good! 👍 

``` json
    {
      "title": "UK Rank",
      "field": "rp_rank",
      "fieldfx": "select case when rp_rank is not null then to_char(rp_rank, 'FM999,999,999') || ' of ' ||  1000 else null end",
      "inline": true,
      "skipNullValue": true,
      "filter": {
        "title": "HELLO",
        "type": "integer"
      }
    },
```

dbauszus-glx (approved):
I quickly revised the documentation. I highly recommend using inline copilot for this kind of things to make sure that missing properties are documented.

Image

PR #2902, opened 2026-07-20
fix(csv_upload): Assign 'null' where data is not supplied in the field 

### LINKED ISSUE #2899: CSV Import Mismatch
**Technical Problem Statement** - When importing a CSV, if not all rows have the same number of values in the columns, data is inserted to the wrong row.
**Customer Problem Statement**  - As a user, when I import a CSV - I expect my data to be assigned to the correct record once imported.

Uploading a CSV with rows of different lengths should infill missing values to preserve data structure. 
Uploading a CSV with all rows of the same length should preserve data structure.
Example Data
In this example, the type and name are missing from row 2. The name is missing from row 3. The type is missing from row 4.
id,type,lat,long,name
1,cafe,51.5074,-0.1278,Central Cafe
2,,51.5154,-0.0928,
3,park,51.5312,-0.1569,
4,,51.5033,-0.1195,London Eye

**Fix in the framework
Issue is in : /utils/csvUpload**
chunkRows is skipping missing columns
In your chunkRows function, you are iterating over the parsed fields of the row using for (const indx in fields).
splitRowIntoFields drops trailing empty columns
At the very end of your splitRowIntoFields function, you have this check:
If a row is missing values at the end (for example, it only has 1 or 3 fields instead of the expected 4), this loop finishes early. The columns that didn't get a value parsed do not get anything pushed to their arrays for that row.

**Current Workaround**
The current workaround ‘solution’ - is to ensure that every value in the CSV has a value, we are typically using a ‘-’ to represent null data. Then in the import function on the db, converting this back to actual null.

AlexanderGeere:
## Description
Fields in a csv are sometimes null, in this pr the missing data is replaced with 'null' to maintain the row length upon insert. 
```csv
A,,C,D,
A,B,C,D,E
,,C,D,
```
On the current patch these rows would get assigned to wrong columns as the empty fields would not be parsed at all 

On this PR the above would become
```csv
A,null,C,D,null
A,B,C,D,E
null,null,C,D,null
```
In the case of `varchar` and `text` the 'null' is set to null.
The nulls are handled by the casting in the query i.e null:numeric, null::text, etc. 

## GitHub Issue
#2899

## Type of Change
- ✅ Documentation

## How have you tested this?
Tested Locally
> [!IMPORTANT]
> `bugs_testing/dataviews/csvupload.json` with `places.csv`

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/utils/csvUpload.mjs
+++ b/lib/utils/csvUpload.mjs
@@ -127,8 +127,9 @@ export default async function csvUpload(file, params = {}) {
       // Split text file into rows on carriage return / new line.
       const rows = e.target.result.trim().split(/\r?\n/);
 
+      let matrix;
       try {
-        const matrix = rows.map((row) => splitRowIntoFields(row));
+        matrix = rows.map((row) => splitRowIntoFields(row));
       } catch (error) {
         resolve([error]);
         return;
--- a/lib/utils/csvUpload.mjs
+++ b/lib/utils/csvUpload.mjs
@@ -73,7 +73,7 @@ This function replaces single quotes in a string with double single quotes for S
 */
 function stringQuotes(val) {
   if (val === 'null') return null;
-  val = val.replaceAll("'", "''")
+  val = val.replaceAll("'", "''");
   return val;
 }
 
@@ -182,7 +182,6 @@ function chunkRows(matrix, params) {
   let chunk = {};
 
   matrix.forEach((fields) => {
-
     for (const indx in fields) {
       const field = Object.keys(params.fields)[indx];
       const type = params.fields[field];
@@ -260,7 +259,7 @@ function splitRowIntoFields(row) {
     expectedFieldCount = fields.length;
   } else if (fields.length !== expectedFieldCount) {
     throw new Error(
-      `Inconsistent number of fields. Expected ${expectedFieldCount}, but got ${fields.length}.`
+      `Inconsistent number of fields. Expected ${expectedFieldCount}, but got ${fields.length}.`,
     );
   }
   return fields;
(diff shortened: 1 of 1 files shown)


simon-leech (dismissed):
This is working for me - tested on a CSV where I used to have to workaround infill every value but without doing that, the data is all aligned now 👍

dbauszus-glx:
Does this have to be in a template literal? `val.replace(/'/g, "''")`

The null check should probably be first.

I'd use a function which allows you to document what is happening without code duplication.

I also think you want to use a replaceAll.

```
function valReplace(val) {
    if (val === 'null') return null;
    val = val.replaceAll("'","''")
    return val;
}
```

dbauszus-glx (changes requested):
I don't think this is right. You map the whole of the csv to get the max fields? This could potentially be millions of rows and break the process.

I think you only want to look at the header or maybe the first 10 lines or so, not all unlimited.

```
      // Check the number of columns in the csv across all rows and return the max or min count.
      // assign the one thats different
      const columnCounts = rows.map((row) => splitRowIntoFields(row).length);
      const columnCount =
        Math.max(...columnCounts) != Object.keys(params.fields).length
          ? Math.max(...columnCounts)
          : Math.min(...columnCounts);
```

dbauszus-glx:
I created documented stringQuotes method to process the varchar and text type fields.

dbauszus-glx:
splitRowIntoFields should only ever be run once. Not for all rows to get the columnCounts and then again for all rows in the chunkRows method.

dbauszus-glx (approved):
Please review my changes to the method.

The rows should be processed into a matrix to be processed for chunks.

A valid comma maybe inside a string value.

Throw an error if the length of a row is not the expected length. This stops the processing.

The error is caught and returned to be shown in the dialog.

simon-leech:
This is working well for me - happy with the changes 👍

AlexanderGeere:
I was getting this error:
screenshot-2026-07-22_09-10-32

In [4edf0e2](https://github.com/GEOLYTIX/xyz/pull/2902/commits/4edf0e2067e5baf809f9b828d1ba40016f00db33) I just defined`let matrix` outside of the scope of the try catch.

Working as expected for me!

PR #2905, opened 2026-07-28
Vitest Clean up & Scaffolding [ENG-9]

### LINKED ISSUE #2896: duplication of code in test scripts
Many, perhaps all, test script files share the same scaffold.

```js
import { afterAll, beforeAll, describe, expect, it } from 'vitest';
import checkWorkspaceCache from '../../../mod/workspace/cache.js';

//Assigning console.error to a property to restore original function with.
const originalConsole = console.error;

//erros from test so we can assert on them and not get polute the console.
const mockErrors = [];

beforeAll(() => {
  //Changing the console.error function to push to our local collection of messages.
  console.error = (message) => {
    mockErrors.push(message);
  };
});

afterAll(() => {
  console.error = originalConsole;
});
```

It might be beneficially to define this once in a test scaffold script to be imported by each of the test scripts.

RobAndrewHurst:
Every Vitest test module that needs to silence or capture console output repeats the same ~15 lines of boilerplate: stash the original method, declare a local message array, swap the method in a `beforeAll`, restore it in an `afterAll`. The scaffold was duplicated across five test modules with three slightly different spellings of the same idea (`originalConsole` / `originalConsoleError`, `mockErrors` / `mockWarns` / `mockWarnings`), and in two cases it was tangled into a `beforeAll` that was also doing unrelated workspace-cache setup.

This PR defines the scaffold once in a new `tests/scaffold.mjs` module and imports it where it's needed.

The module exports `mockConsole(method = 'error')`, which registers its own `beforeAll` / `afterAll` hooks and returns the array of collected messages:

```js
export function mockConsole(method = 'error') {
  const messages = [];

  //Assigning the console method to a property to restore the original function with.
  let originalConsole;

  beforeAll(() => {
    originalConsole = console[method];

    //Changing the console method to push to the local collection of messages.
    console[method] = (message) => {
      messages.push(message);
    };
  });

  afterAll(() => {
    console[method] = originalConsole;
  });

  return messages;
}
```

The single signature covers both usages found in the suite — capture-and-assert (`const mockWarnings = mockConsole('warn')`) and suppress-only, where the return value is simply ignored.

Test modules refactored onto the helper:

| File                                                      | Replaced with                              |
| --------------------------------------------------------- | ------------------------------------------ |
| `tests/mod/workspace/getTemplate.test.mjs`                | `mockConsole('error')`                     |
| `tests/mod/utils/sqlFilter.test.mjs`                      | `mockConsole('warn')`                      |
| `tests/mod/workspace/templates/sql_table_insert.test.mjs` | `const mockWarnings = mockConsole('warn')` |
| `tests/mod/view.test.mjs`                                 | `mockConsole('error')`                     |
| `tests/mod/query.test.mjs`                                | `mockConsole('error')`                     |

For `view` and `query` the console handling was lifted out of the `describe`'s `beforeAll` / `afterAll`, leaving those hooks responsible only for setting `xyzEnv` and loading the workspace cache. Both files no longer import `afterAll` from vitest at all.

`TESTING.md` gains a `#### Console mocks` subsection under Mocks documenting the helper, a note in Configuration describing what `tests/scaffold.mjs` is for, and a Best Practices bullet pointing future test modules at the scaffold rather than at copy-paste.

Two related duplications were left out of scope as they are not what the issue describes — happy to fold either in if wanted:

- the `globalThis.xyzEnv = {...}` + `await checkWorkspaceCache(...)` pattern repeated across ~6 modules
- the `vi.mock('../../mod/utils/logger.js', ...)` suppression stub duplicated in 3 modules

## GitHub Issue

Closes https://github.com/GEOLYTIX/xyz/issues/2896

## Type of Change

- ✅ Documentation
- ✅ Testing

## How have you tested this?

This is a refactor of test scaffolding with no changes to any module under `mod/`, so the existing suite is the verification — the same assertions must continue to pass against the shared helper.

```bash
pnpm test
```

```
Test Files  43 passed (43)
     Tests  277 passed (277)
```

Test output remains clean, confirming the suppression still takes effect: `mockConsole` swaps the method in a `beforeAll`, exactly as the inline scaffold did, so the hook ordering relative to workspace-cache loading is unchanged.

The one assertion that reads back a captured message — `expect(mockWarnings[0])` in `sql_table_insert.test.mjs` — passes unchanged. `mockConsole` deliberately collects only the first argument of each call, matching the previous behaviour so array indices in existing assertions keep their meaning.

Formatting and lint:

```bash
pnpm exec biome check tests/ TESTING.md
```

```
Checked 197 files in 27ms. No fixes applied.
```

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -103,6 +103,7 @@ Key points:
 
 - Only `tests/mod/**` and `tests/plugins/**` are included. Browser tests (`tests/lib/**`, `tests/browser/**`) are excluded — they use Codi and run in the browser.
 - `tests/setup.mjs` ensures `globalThis.xyzEnv` exists before any test module loads.
+- `tests/scaffold.mjs` provides shared setup helpers which test modules import explicitly, eg. `mockConsole`.
 - Tests run in parallel across files with a 10-second timeout per test.
 
 ### CI Pipeline
@@ -290,6 +291,38 @@ describe('HTTP Mock', () => {
 });
 ```
 
+#### Console mocks
+
+XYZ modules log warnings and errors to the console. Tests which trigger these paths would otherwise pollute the test output, and tests which assert on the logged message need to capture it.
+
+The `mockConsole` helper in `tests/scaffold.mjs` replaces a console method for the duration of the test file and restores the original in an `afterAll` hook. Messages are collected in the returned array so they can be asserted on.
+
+```javascript
+import { describe, it, expect } from 'vitest';
+import { mockConsole } from '../../scaffold.mjs';
+
+const mockWarnings = mockConsole('warn');
+
+describe('sql_table_insert', () => {
+  it('warns about a potential SQL injection', () => {
+    sql_table_insert(req);
+
+    expect(mockWarnings[0]).toEqual(
+      'Potential SQL Injection in sql_table_insert request body.',
+    );
+  });
+});
+```
+
+The method defaults to `error`. The return value can be ignored when the intent is only to suppress output.
+
+```javascript
+// Suppress console.error from getTemplate for missing template tests.
+mockConsole('error');
+```
+
+`mockConsole` registers its own `beforeAll` / `afterAll` hooks, so it must be called at the top level of a test module or inside a `describe` body -- not inside an `it` test.
+
 ### Available Assertions
 
 Vitest provides a rich set of assertions via the `expect()` API:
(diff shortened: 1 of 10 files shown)


dbauszus-glx (approved):
I commented out the object freeze in order to have tests pass in the vitest vscode extension.

AlexanderGeere:
@dbauszus-glx 
We need the `Object.freeze` to prevent crashing the process though right? we cant leave it out
See #2876

dbauszus-glx:
@AlexanderGeere We create the templates object from null to protect modification of the prototype.

```
workspace.templates = Object.create(null);
```

But you are right. We need to add a test for creating a template with the key 'toString' this shouldn't crash when add to the templates object created from null.

AlexanderGeere:
@dbauszus-glx I pushed a test, we can probably just remove that `object.freeze` as it isnt necessary

AlexanderGeere (approved):

PR #2908, opened 2026-07-29
Cloudinary status [ENG-34]

dbauszus-glx:
A fetch [post] request to the cloudinary api with a valid signature will not err but return with a status that is not 200. Ie 404 if the file size exceeds a certain limit depending on the cloudinary account.

The process should not update location and get stuck but end and enable the location view.

Fixes ENG-34


AlexanderGeere:
@dbauszus-glx 
I added the same re-enable to the signedUrl function, i was getting a 500 becuase i had not specified the URL in my env but the location was not being re-enabled.

AlexanderGeere (approved):

lib/ui/locations/entries/cloudinary.mjs
```diff
@@ -270,12 +270,13 @@ async function imgOnload(entry, img, file) {
     method: 'post',
   });
 
-  if (!response || response.error) {
+  if (!response || response.error || response.status !== 200) {
```
RobiieHurst: [ISSUE] errorDetail is always empty, so this alert never says what went wrong.

`fetch()` resolves a `Response`, which has no `.error` property, so
`response?.error?.message` is permanently `undefined` and `errorDetail`
is always "". `!response` is dead too — fetch never resolves falsy.
`response.status !== 200` is the only live condition here, and the user
just sees "Cloudinary Image upload failed!" with no reason.

The 404-on-oversized-file case that motivated this PR puts the reason in
the body: {"error":{"message":"File size too large..."}}. So:

    if (!response.ok) {
      const body = await response.json().catch(() => null);
      const errorDetail = body?.error?.message
        ? `Error: ${body.error.message}`
        : `Status: ${response.status}`;
      ...
    }

I think it might be better to use `!response.ok` over `status !== 200` — it accepts the whole
2xx range instead of assuming Cloudinary only ever answers 200.

RobiieHurst (changes requested):

lib/ui/locations/entries/cloudinary.mjs
dbauszus-glx: This has been resolved in the last committ.

simon-leech (approved):

RobAndrewHurst (approved):

PR #2910, opened 2026-07-30
fix(filter): Use includes/includeAll correctly [ENG-5]

### LINKED ISSUE #2909: Layers that have multi-filters ignore the include array
Steps to reproduce:
Example application - xyz_resources/bugs_testing/filter/retail_places.json. 
Contains 2 layers one with a filter.exclude and one with a filter.include array.

Turn on the Filter Include Layer 

Click the filtering button 

You will see a long list of filter options. Even though the configuration provided is

```json
"filter": {
         "include": [
           "prim_shop_pop"
         ]
       },
```

Context:

Using filter.include array on a layer that has the multi_filter object cannot work. 

In the multi_filter(layer){} function inside lib/ui/layers/panels/filter.mjs we set includeAll to true if the multi_filter object is provided. 

On line 238 of the same file we look for if includeAll or filter.include.includes() . However, as includeAll is hardcoded to try, you cannot ever use include array to limit the filter creation to just a subset of fields. 

It is working for exclude array fine. We largely use exclude in configuration, include is much more rarely used. 

Also at the minute we can workaround this by just using exclude.

AlexanderGeere:
## Description
The include option was previously not working as filters defined on the entry would be included regardless.
Using the multi_filter object in config would all set includeAll to true, this is now set to false if include is used. 

## GitHub Issue
#2909 

## Type of Change

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

## Code Quality Checklist

### DIFF
--- a/lib/ui/layers/panels/filter.mjs
+++ b/lib/ui/layers/panels/filter.mjs
@@ -241,30 +241,31 @@ function listFilter(layer) {
     ) {
       entry.type ??= 'text';
 
-      if (!Object.hasOwn(filterByType, entry.type)) continue;
-
       if (!Object.keys(entry.filter || {}).length)
         entry.filter = filterByType[entry.type];
+
+      // The filter is defined as a string e.g. "like"
+      if (typeof entry.filter === 'string') {
+        entry.filter = {
+          type: entry.filter,
+        };
+      }
+
+      if (
+        !entry.filter?.type ||
+        !Object.hasOwn(mapp.ui.layers.filters, entry.filter.type)
+      ) {
+        entry.filter ??= { type: entry.type };
+        console.warn(`${entry.filter.type} is not a valid filter type`);
+        continue;
+      }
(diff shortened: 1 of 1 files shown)


simon-leech (approved):
This is working for me. 
I tested with an `include` - only see those ones. 
I tested with neither `include` or `exclude` - see everything in the `infoj`.
I tested with an `exclude` - that is the only one that is excluded.

dbauszus-glx (approved):

wbecker (approved):
Just testing linear link

wbecker:
test comment

wbecker (changes requested):
It's just not good enough

wbecker:
sorry was just testing what happens if I reject something! Yes - it's good - I'll reverse that 🙂

wbecker (approved):

PR #2912, opened 2026-08-03
Deployment Documentation [ENG-8]

RobAndrewHurst:
## Make Varlock optional, including on Vercel

Previously a Vercel deployment **had** to ship a frozen `.varlock.blob` — without one the app threw on boot. This makes Varlock opt-in instead, so a project can just use Vercel project environment variables.

### What changed

`apps/xyz/mod/utils/processEnv.js` had two hard requirements:

1. It threw `Missing .varlock.blob` when `VERCEL` was set and no blob was present.
2. Less obvious — `internal.initVarlockEnv()` throws by itself when nothing has been hydrated. It now takes `{ allowFail: true }`, the same option Varlock's own auto-load path uses. Without this, removing (1) would just have moved the crash a few lines down.

With no blob, the app falls through to the environment variables the process was launched with, then applies its normal defaults.

### Behaviour

| Scenario | Before | After |
|---|---|---|
| `VERCEL=1`, no blob | ❌ throws | ✅ uses Vercel project env |
| plain `node`, no `varlock run` | ❌ throws | ✅ uses shell env |
| blob present (plain or encrypted + key) | ✅ hydrates | ✅ unchanged — blob still wins over `process.env` |
| encrypted blob, **no** `_VARLOCK_ENV_KEY` | ❌ throws | ❌ still throws |

That last row is deliberate: an undecryptable blob is a misconfiguration, not an opt-out.

Note this is unconditional, not Vercel-only — so a bare `node apps/xyz/server.js` now works too. The trade-off is losing Varlock's "try rerunning via `varlock run`" hint locally. Happy to scope it to `if (process.env.VERCEL)` if we'd rather keep that.

### Docs 📚️

New `DEPLOYMENT.md` covering both options — Vercel project env vars (simpler, Git-integration deploys work normally) vs. a frozen blob (keeps schema validation, `gsm()` refs, redaction). Also fixed `varlock/README.md`, which still said freezing was required.


DEPLOYMENT.md
```diff
@@ -0,0 +1,364 @@
...
+
+- Credentials for whatever backs your `gsm()` references — a service account JSON in `GOOGLE_CREDENTIALS`, or GCP Workload Identity Federation.
+- The freeze step before the deploy step:
+
+```yaml
+- run: pnpm install --ignore-scripts
+- run: pnpm freeze-env --env=production
+  env:
+    _VARLOCK_ENV_KEY: ${{ secrets.VARLOCK_ENV_KEY }}
+    GOOGLE_CREDENTIALS: ${{ secrets.GOOGLE_CREDENTIALS }}
```
wbecker, line 252: Is this bit enough to explain to people how to integrate with Google Secrets Manager? (It may be - I haven't done it before so not 100% sure - just double checking!)

wbecker (commented):
Looks like it covers most things well! Just double checking on whether we talk enough about integration…

simon-leech (approved):
Read over the docs and all makes sense to me! Happy to use this as the starting point for v5 and test a deployment once it's ready! 


Just to check it's correct that the api.test.mjs is being deleted in this PR?

dbauszus-glx:
@simon-leech yeah because the api module is no longer a thing with the express routing.

AlexanderGeere (approved):

dbauszus-glx (approved):
I reset the workspacePromise to null on err.

I updated the documents in regards to installing the cli, deploying with env in vercel.json and uploading the env to Vercel project.

PR #2914, opened 2026-08-04
Pass layer.queryparams to hover/infotip query [ENG-23]

cityremade:
Includes layer.queryparams in the infotip request to make them accessible inside a hover/infotip template.

### DIFF
--- a/lib/layer/featureHover.mjs
+++ b/lib/layer/featureHover.mjs
@@ -12,6 +12,14 @@ The featureHover method will send an xhr request for the provided feature argume
 
 @param {object} feature
 @param {layer} layer A decorated mapp layer object.
+@property {object} feature.properties The feature properties object.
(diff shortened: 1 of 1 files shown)


dbauszus-glx:
is this an issue in the current minor? or should this be a patch pr?

simon-leech:
I'm comfortable with this as part of the v4.24.0 minor - its not really a bug as its never been a thing, not like it just stopped working! 👍

dbauszus-glx:
Instead of creating the object and then spreading the object into a new object with another object inside the parameter for the paramString method it is much easier to just spread the layer.queryParams into the queryParams and provide this object as param for the paramString method.

dbauszus-glx (approved):
I added missing property descriptor with copilot.

simon-leech (approved):
This works as expected for me - `layer.queryparams` are passed into the infotip request 👍

PR #2915, opened 2026-08-05
v4.24 release notes [ENG-43]

dbauszus-glx:
A Claude skill was added to the repository to compile the release notes which are stored in the release-notes directory.

### DIFF
--- a/.claude/skills/release-notes/SKILL.md
+++ b/.claude/skills/release-notes/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: release-notes
+description: Use when drafting, generating, revising, or reviewing GitHub release notes, changelogs, or a "What's Changed" section for an XYZ release.
+---
+
+# Release Notes
+
+Create concise, user-focused release notes from the supplied pull requests, commits, or version range. Match the XYZ release format below exactly unless the user explicitly requests another format.
+
+## Gather The Changes
+
+1. Identify the repository, previous version, new version, and included pull requests from the user's request or repository context.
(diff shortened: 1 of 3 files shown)


RobAndrewHurst (approved):

.claude/skills/release-notes/SKILL.md
```diff
@@ -0,0 +1,73 @@
+---
```
wbecker, line 1: Do we want this released in the public repo?

.gitignore
```diff
@@ -35,3 +35,5 @@ docs/*
 !jsdoc*.json
 !codi.json
 !biome.json
+
+.claude/settings.local.json
```
wbecker, line 39: People get funny about AI being used in OSS repos - do we want to add this stuff in?

RobAndrewHurst: It's to do something that should be automated

.claude/skills/release-notes/SKILL.md
dbauszus-glx: I don't see the hurt in adding a SKILL which has valuable information.

.gitignore
wbecker: Just flagging it - if we are happy with it, that's fine!

dbauszus-glx: I understand the concern but it is not reasonable. It is worse to use skills but hiding the use or not disclosing which skills have been used. The skill provides context how we compile release notes.

cityremade (approved):

AlexanderGeere (approved):
I changed the wording around the test coverage, it previously said '%97+'.

Looks good otherwise

PR #2920, opened 2026-08-10
Reusable workspace scope API

### LINKED ISSUE #2918: Workspace scope logic is not reachable outside the workspace API
## Problem

`workspace/composeObj` records a templateScope for every composed object and gates access with `checkScope`. Hosts which resolve roles outside the workspace API cannot reach either of these.

Two things are currently private:

1. **`checkScope`** (`mod/workspace/composeObj.js`) is module scope. A host that resolves a user's roles from an external source has no way to evaluate a scope string with the same semantics composition applies, so it has to reimplement the match. Any reimplementation drifts from `checkScope` — in particular the rule that a scope is granted by a descendant role but not by an ancestor one, which is easy to get backwards.

2. **The scope enumeration** is inlined in the `scopes()` request handler (`mod/workspace/_workspace.js`). A templateScope is only recorded once the object defining it has been composed, so the complete set of scopes in a workspace is only obtainable by composing every locale, nested locale, and layer with role checks bypassed. That routine exists, but only behind `GET /api/workspace/scopes`. There is no way to call it in process.

`scopesArrayToTree` also writes directly to the response, so the tree conversion cannot be reused either.

## Why it matters

Anything that needs the scope universe — an admin UI, an external authorization layer, a test — either goes over HTTP to its own instance or reimplements composition. We have hit the second case downstream and it duplicated a large amount of workspace traversal that then broke when `mod/provider/getFrom.js` was replaced.

## Proposed change

- Export `checkScope` from `workspace/composeObj`.
- Extract the enumeration to a new `workspace/getScopes` module, exporting `getScopes()` plus `scopesArray()` and `scopesTree()` as pure functions.
- `scopes()` composes those instead of inlining them.

No behaviour change. `getScopes` performs the same forced cache rebuild, `cacheSources` call, and role bypassed locale composition, and still discards the composed locales.

## Out of scope

`_workspace.js` carries a `// TODO check why the scopes array is different from composedWorkspace`. The difference is that `scopes()` passes a throwaway `{ locales: {} }` to `nestedLocales` while `composeWorkspace()` writes the composed locales back into the cached workspace. Resolving that changes whether shared cache state is mutated, so it is deliberately left alone here and worth its own issue.

RobAndrewHurst:
Closes GEOLYTIX/xyz#2918

Makes the workspace scope logic reachable outside the workspace API. No behaviour change.

Re-creation of GEOLYTIX/xyz#2919, which GitHub auto-marked as merged when the branch was accidentally pushed to major (since undone).

### DIFF
--- a/apps/xyz/mod/workspace/_workspace.js
+++ b/apps/xyz/mod/workspace/_workspace.js
@@ -32,7 +32,7 @@ import logger from '../utils/logger.js';
 import workspaceCache from './cache.js';
 import getLayer from './getLayer.js';
 import getLocale from './getLocale.js';
-import getScopes, { scopesArray, scopesTree } from './getScopes.js';
+import { getScopes, scopesArray, scopesTree } from './scopes.js';
 
 const keyMethods = {
   layer,
--- a/apps/xyz/mod/workspace/_workspace.js
+++ b/apps/xyz/mod/workspace/_workspace.js
@@ -32,6 +32,7 @@ import logger from '../utils/logger.js';
 import workspaceCache from './cache.js';
 import getLayer from './getLayer.js';
 import getLocale from './getLocale.js';
+import getScopes, { scopesArray, scopesTree } from './getScopes.js';
 
 const keyMethods = {
   layer,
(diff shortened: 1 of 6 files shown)


dbauszus-glx:
I moved the checkScopes method to the getScopes module as it makes more sense for the composeObject to import this method and keep the scopes methods together.

Would it make sense to rename the getScopes method to just scopes and not define getScopes as default since this method would usually not be used on it's own but in combination with other methods from the same module scopes module.

dbauszus-glx:
I updated the checkScopes method to use some instead of findIndex in regards to the new SonarCloude issue flagged earlier.

dbauszus-glx (approved):
I changed the getScopes method to a named export rather than the default export and the name of the getScopes module to scopes to prevent ambiguity.

PR #2924, opened 2026-08-10
Only show legend from showcallback method [ENG-66] [ENG-92]

dbauszus-glx:
The theme layerStyle module method creates an empty div node for the legend and assigns the legends method to the layer.showCallbacks array.

The legend method eg categorized will replace the children in this legend node with the legend when called from the showCallbacks methods on show.

### DIFF
--- a/lib/ui/elements/layerStyle.mjs
+++ b/lib/ui/elements/layerStyle.mjs
@@ -349,7 +349,7 @@ function theme(layer) {
     });
 
     // The legend methods replace the children of the layer.style.legend node with the legend content.
-    layer.style.legend = mapp.utils.html.node`<div>`;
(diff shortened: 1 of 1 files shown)


simon-leech (changes requested):
This is definitely faster on initialization now! 

However, the legend is no longer inline in a grid like it is off this PR. 

OFF THE PR
Image

ON THIS PR 
Image

dbauszus-glx:
@simon-leech I forgot to add the legend class. Can you retest now.

simon-leech (approved):
This looks good to me now! 

Seems to make a real speed improvement too 🥇

cityremade (approved):

RobAndrewHurst (approved):

PR #2925, opened 2026-08-11
Add authorization provider seam to workspace composition

### LINKED ISSUE #2922: Support an injectable asynchronous authorization provider in workspace composition
## Motivation

Hosts composing the XYZ core increasingly need to authorize workspace access against an external policy store — in our case OpenFGA (relationship-based access control), but the same applies to any policy engine (OPA, Cerbos, SpiceDB). These systems answer access questions **asynchronously** and the answer cannot be expressed as a static `user.roles` array signed into a token: access is derived from tuples (user -> role -> policy -> scope) that change independently of the session.

The core currently evaluates access synchronously inside workspace composition via `checkScope(templateScope, roles)`, hardwired to the `user.roles` array.
There is no seam to substitute this decision, which forces downstream hosts to rebuild large parts of the workspace API outside the core (detailed below).

## Current behaviour

Composition is already fully async, but the access decision is not:

- `mod/workspace/composeObj.js` builds the scope chain `templateScope = [...obj.parentRoles, obj.role].filter(Boolean)`, records it in `workspace.scopes`, then gates with the synchronous `checkScope(templateScope, roles)`. A failed check returns `Error("User does not have access to object with template scope: …")`.

- `mergeTemplateIntoObj` repeats the same pattern per nested template (silently skipping the merge on deny), and `rolesTemplates` for role-keyed template branches.

- `checkScope` semantics: empty `templateScope` → allow without consultation; `roles === true` -> bypass (used internally by admin flows such as `cacheSources`, `scopes`, `nestedLocales`); otherwise membership / prefix matching against the roles array (plus `LEGACY_ROLES`).

- `getLocale` / `getLayer` pass `params.user?.roles` down into `composeObj`; the `/api/workspace` endpoints in `mod/workspace/_workspace.js` return the composed objects, delete `role` / `parentRoles`, and assign a checksum.

- `mod/query.js` and `mod/view.js` resolve templates/layers through the same role-gated composition.

## What an external-authorization host must currently do

Because the decision point cannot be substituted, our fork carries a parallel
runtime that:

1. **Elevates the user to `roles: true`** before delegating to core handlers, so composition returns the complete object regardless of roles config.
2. **Re-implements the `/api/workspace/:key` endpoints** for `locale`, `locales`, and `layer` — including its own composition cache, its own recursive path-based filtering, its own `roles` stripping, and its own checksum assignment — calling `getLocale`/`getLayer` as library functions.
3. **Intercepts `res.json`/`res.send`** on any remaining `/api/workspace` response to re-filter core output through the external policy check.
4. **Adds preflight middleware** on `/api/query` and `/view` that resolves the template and layer, checks the external store, and only then hands the elevated user to the core handler.

The costs are real: every request composes twice (once in core, once in the shim), the shim is coupled to core response shapes and checksum behaviour, and an elevated (`roles: true`) user flows through core handlers — any core route the interception layer misses fails **open**. Each core release requires re-verifying all of this by hand.

## Proposal

Introduce an authorization provider seam at the existing decision point. The
provider answers the exact question `checkScope` answers today, but async:

```js
/**
@typedef {Object} AuthorizationProvider
@property {(context: {
  scope: string[],       // the templateScope chain, e.g. ['retail', 'analyst']
  scopeKey: string,      // scope.join('.') — matches workspace.scopes entries
  obj: Object,           // the object being composed (locale/layer/template)
  user: Object|undefined // params.user
}) => Promise<boolean>} checkScope
*/
```

Registration could be either:

- **Router-level**: `createRouter(middleware, { authorization })`, threaded through to composition; or
- **Module-level**: a `mod/workspace/authorization.js` with `setAuthorizationProvider(provider)` and a default provider that implements the current roles-array semantics.

Given how deeply recursive composition is (and the module-scoped workspace cache already used by `composeObj`), module-level registration is the smaller change: 
`composeObj`, `mergeTemplateIntoObj`, and `rolesTemplates` replace their`checkScope(templateScope, roles)` calls with `await provider.checkScope(context)`.
All call sites are already inside async functions, so no signature changes propagate upward.

The `scopeKey` string is deliberately the same value composition already records in `workspace.scopes`. That gives external stores a clean contract: provision policy objects from `GET /api/workspace/scopes`, and answer checks against the same identifiers — no separate mapping layer.

## Semantics to agree on

1. **Empty scope chains.** `checkScope` currently short-circuits to allow when `templateScope` is empty, so unrestricted objects never reach the decision point. An external provider needs to be consulted for every composed node to support deny-by-default models. Suggestion: always call the provider; the default (roles) provider preserves the current early-allow, so behaviour is unchanged unless a host opts in.
2. **The `roles: true` bypass.** Internal flows (`cacheSources`, the `scopes` endpoint, `nestedLocales`) compose with `user: { roles: true }`. This should become an explicit internal/system bypass that skips the provider, rather than a magic value every provider must special-case — and it should no longer be reachable from request params.
3. **Deny shape.** Provider denial should map to today's outcomes: an `Error` for the root object, a silent skip for merged templates and array items — so clients see no behavioural difference.
4. **Failure mode.** A provider that throws must fail **closed** (composition error), never fall through to allow.
5. **Performance.** The provider can be called many times per composition. Providers are expected to cache/batch; the core could additionally dedupe checks by `scopeKey` + user within a single composition pass.
6. **Backwards compatibility.** With no provider registered, the default provider is the current `checkScope` logic verbatim (including `LEGACY_ROLES`). No workspace, token, or client changes required.

## Out of scope (follow-ups this enables)

- `mod/query.js` / `mod/view.js` consulting the provider for template, layer, `dbs`, and infoj field-template access before execution — which would let hosts drop preflight middleware entirely.
- Omitting `roles` configuration objects from client responses when an external provider is active.

## Acceptance criteria

- [x] An async authorization provider can be registered by a composing host.
- [x] `composeObj`, `mergeTemplateIntoObj`, and `rolesTemplates` route every
      scope decision through the provider, including empty scope chains.
- [x] The provider receives the scope chain, the joined `scopeKey` (matching
      `workspace.scopes` entries), the object under composition, and the user.
- [x] Internal admin composition uses an explicit system bypass, not
      `roles: true` via params.
- [x] Provider errors fail closed.
- [x] Default behaviour without a registered provider is byte-for-byte
      identical to current role checks.

RobAndrewHurst:
## What??

New module `mod/workspace/authorization.js`. Workspace composition route template-scope decisions through registered async provider. Off by default: `AUTHORIZATION_PROVIDER` xyzEnv flag gate it, same pattern as `LEGACY_ROLES`. Flag unset — `checkScope(scope, roles)`
semantics apply, provider never consulted, behaviour byte-identical.

## Why??

External policy stores (OpenFGA, OPA) answer async, keyed on subject — not roles array
signed into token. Core check sync inside composition. No seam = host must elevate users
to `roles: true`, re-implement workspace endpoints, intercept responses, preflight query/view.
Every request will then compose twice. Any missed route fail open.

Seam = decision made where data assembled. Missed integration path fail closed. Deny happens
before template merge — restricted content never enters object, no un-merge problem, per-user
checksums correct for free, one composition per request.

## How??

- `setAuthorizationProvider(provider)` register. No-arg call clear registration.
- `authorizeScope(context)` resolve one decision. Call sites await unconditionally — method
  condition itself on flag.
- Context: `{ scope, scopeKey, obj, roles }`. `scopeKey` = `scope.join('.')`, matches
  `workspace.scopes` entries — host provision policy store from `GET /api/workspace/scopes`,
  no mapping layer.
- Provider consulted for every composed node, empty scope chains included — deny-by-default
  models possible.

| Flag  | Provider   | Decision                                              |
| ----- | ---------- | ----------------------------------------------------- |
| unset | any        | `checkScope(scope, roles)` — provider never consulted |
| set   | registered | provider decide, must resolve `true`                  |
| set   | missing    | denied, fail closed                                   |
| set   | throws     | denied, fail closed                                   |

Deny shapes unchanged:

| Decision point         | On deny                          |
| ---------------------- | -------------------------------- |
| `composeObj` root      | `Error`, same message as before  |
| `mergeTemplateIntoObj` | merge silently skipped           |

Scope chains stay chained: `analyst` template inside `retail` object ask provider about
`retail.analyst`.

## Tests

`tests/mod/workspace/authorization.test.mjs`, 10 tests:

- flag unset: roles semantics apply, registered provider ignored (never called)
- provider allow: scoped object admitted, role-gated template merged, no user roles needed
- provider deny: root return Error, template merge skipped — despite matching user roles
- flag on without provider, provider throw: both fail closed
- merged templates consulted with chained scopeKey
- flag removed: roles semantics restored

Full suite 317/317. Flag-unset behaviour untouched.

## Files

| File                                         | Change                                                   |
| -------------------------------------------- | -------------------------------------------------------- |
| `mod/workspace/authorization.js`             | new                                                      |
| `mod/workspace/composeObj.js`                | both decisions await `authorizeScope`; `checkScope` import removed |
| `mod/utils/processEnv.js`                    | `AUTHORIZATION_PROVIDER` in xyzEnv typedef               |
| `tests/mod/workspace/authorization.test.mjs` | new, 10 tests                                            |

### DIFF
--- a/apps/xyz/mod/utils/processEnv.js
+++ b/apps/xyz/mod/utils/processEnv.js
@@ -110,6 +110,7 @@ The process.ENV object holds configuration provided to the node process from the
 @property {String} [SAML_PROVIDER_NAME] Display name for your service
 @property {String} [SLO_CALLBACK] URL for handling logout callbacks
 @property {Boolean} [LEGACY_ROLES] Enable legacy role checks
+@property {Boolean} [AUTHORIZATION_PROVIDER] Route workspace composition scope checks through the [authorization]{@link module:/workspace/authorization} provider registered by the composing host.
 */
 
 const defaults = {
--- a/apps/xyz/mod/workspace/authorization.js
+++ b/apps/xyz/mod/workspace/authorization.js
@@ -1,0 +1,83 @@
+/**
+## /workspace/authorization
+
+The authorization module exports the seam through which workspace composition
+routes template scope access decisions.
+
+A composing host may register an asynchronous authorization provider to answer
+scope decisions from an external policy store, eg. OpenFGA.
+
+The provider is only consulted when the AUTHORIZATION_PROVIDER flag is enabled
+in the xyzEnv. Without the flag the authorizeScope method returns the
+synchronous checkScope semantics for the user.roles array, matching the
+LEGACY_ROLES flag pattern.
+
+@requires /workspace/scopes
+
+@module /workspace/authorization
+*/
+
+import { checkScope } from './scopes.js';
+
+/**
+@global
+@typedef {Object} AuthorizationContext
+@property {Array<string>} scope The templateScope chain of the object being composed.
+@property {string} scopeKey The joined scope chain. The scopeKey matches the identifiers recorded in workspace.scopes.
+@property {Object} obj The object being composed.
+@property {array|boolean} [roles] The user roles from request params.
+*/
+
+/**
+@global
+@typedef {Object} AuthorizationProvider
+@property {function(AuthorizationContext):Promise<boolean>} checkScope Resolves whether the user has access to the scope.
+*/
+
+let provider;
+
+/**
+@function setAuthorizationProvider
+
+@description
+The method registers an authorization provider for workspace composition. The
+provider is only consulted with the AUTHORIZATION_PROVIDER flag enabled in the
+xyzEnv.
+
+Calling the method without a provider argument clears the registration.
+
+@param {AuthorizationProvider} [nextProvider] The provider to register.
+*/
+export function setAuthorizationProvider(nextProvider) {
+  provider = nextProvider;
+}
+
+/**
+@function authorizeScope
+@async
+
+@description
+The method resolves a scope access decision for workspace composition.
+
+The checkScope semantics for the user.roles array apply unless the
+AUTHORIZATION_PROVIDER flag is enabled in the xyzEnv.
+
+With the flag enabled the decision is routed through the registered provider.
+The provider must resolve true for the scope access to be granted. A missing
+provider or a provider error fails closed and access is denied.
+
+@param {AuthorizationContext} context The scope decision context.
+@returns {Promise<boolean>} Whether the user has access to the scope.
+*/
+export async function authorizeScope(context) {
+  if (!xyzEnv.AUTHORIZATION_PROVIDER) {
+    return checkScope(context.scope, context.roles);
+  }
+
+  try {
+    return (await provider?.checkScope(context)) === true;
+  } catch (error) {
+    console.error(error);
+    return false;
+  }
+}
--- a/apps/xyz/mod/workspace/authorization.js
+++ b/apps/xyz/mod/workspace/authorization.js
@@ -1,16 +1,11 @@
 /**
 ## /workspace/authorization
 
-The authorization module exports the seam through which workspace composition
-routes template scope access decisions.
+The authorization module exports the seam through which workspace composition routes template scope access decisions.
 
-A composing host may register an asynchronous authorization provider to answer
-scope decisions from an external policy store, eg. OpenFGA.
+A composing host may register an asynchronous authorization provider to answer scope decisions from an external policy store, eg. OpenFGA.
 
-The provider is only consulted when the AUTHORIZATION_PROVIDER flag is enabled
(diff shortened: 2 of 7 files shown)


dbauszus-glx (approved):
This works well. All tests passing.

I noticed the docs weren't building and couldn't be shown.

I fixed this in a commit to check the globals added in this PR.

AlexanderGeere (approved):

PR #2930, opened 2026-08-17
Replace AUTHORIZATION_PROVIDER environment variable with authorization_provider property of user object [ENG-89]

### LINKED ISSUE #2929: Authorization provider seam: pass the requesting user into the composition context
## Motivation

GEOLYTIX/xyz#2925 landed the first increment of the provider seam from GEOLYTIX/xyz#2922. The context carries `{scope, scopeKey, obj, roles}`. The `user` property proposed in GEOLYTIX/xyz#2922 was deferred.

Without the subject, the provider cannot delegate the decision. Policy stores answer subject questions — OpenFGA checks *"is* `user:rob@…` *an assignee of* `role:analyst`*"*. A provider given only a roles array can check which roles exist in the store, but must trust the token/ACL roles array for who holds them.

* The signed `user.roles` array stays the source of truth. Revoking a store assignment does not revoke composition access until the token or ACL record changes.
* Our OpenFGA provider can only filter `context.roles` against the store registry. It cannot read `role#assignee` tuples.
* The workaround — middleware enriching `user.roles` from the store — rebuilds the decision outside the seam. GEOLYTIX/xyz#2922 set out to remove exactly that duplication.

## Current behaviour

Both entry points hold the user but forward only its roles:

* `mod/workspace/getLocale.js` → `composeObj(locale, params.user?.roles)`
* `mod/workspace/getLayer.js` → `composeObj(layer, params.user?.roles)`

`roles` threads through `mergeTemplateIntoObj`, `parseTemplates`, `templatesArray`, `rolesTemplates`, `arrayProperty`, `queryTemplate`. Both `authorizeScope` call sites build the context from it.

## Proposal

Additive, as typed in GEOLYTIX/xyz#2922:

* `AuthorizationContext` gains optional `user` (read-only for providers).
* `composeObj(obj, roles, user)` — optional third parameter, threaded where `roles` travels. Both `authorizeScope` contexts include it.
* `getLocale`/`getLayer` pass `params.user`.

Invariants:

* Flag disabled: threading is inert, `checkScope(scope, roles)` unchanged.
* `roles === true` still bypasses the provider.
* Absent user stays absent, so providers can tell anonymous from not-threaded.
* Fail-closed rules from GEOLYTIX/xyz#2925 unchanged.

## Tests

1. Provider receives the user in the context.
2. Chained template decision (`retail.analyst`) receives the same user each call.
3. `user` absent when caller passes none.
4. Flag off: user argument changes nothing.
5. `getLocale` forwards `params.user`.
6. Same user reference across one composition — no clone per call.

## Out of scope

* `rolesTemplates` role-keyed merges and negated `!role` keys.
* Response-side roles filtering (`mod/utils/roles.js`).

dbauszus-glx:
The user object is now defined as a global in the authorization module.

Instead of passing the roles param to the composeObj method the user object with the roles property is passed on.

The user object has an optional authorization_provider string property which can be used to define an alternative authorization provider eg. openfga.

### DIFF
--- a/apps/xyz/mod/query.js
+++ b/apps/xyz/mod/query.js
@@ -68,7 +68,7 @@ export default async function query(req, res) {
   // Must be run after the layerQuery method since the query template could be defined within the layer [template].
   const template = await composeObj(
     { template: req.params.template },
-    req.params.user?.roles,
+    req.params.user,
   );
 
   if (template instanceof Error) {
--- a/apps/xyz/mod/user/cookie.js
+++ b/apps/xyz/mod/user/cookie.js
@@ -118,6 +118,11 @@ export default async function cookie(req, res) {
         user.session = payload.session;
       }
 
+      // The authorization_provider property is assigned by a composing host and is not stored in the ACL. The property must be carried over from the current token.
+      if (payload.authorization_provider) {
+        user.authorization_provider = payload.authorization_provider;
+      }
+
       //If we have a sessionIndex from a SAML IdP then we need to pass it from the
       //already created cookie as this is not stored in the acl.
       if (decodedCookie.sessionIndex) {
--- a/apps/xyz/mod/user/key.js
+++ b/apps/xyz/mod/user/key.js
@@ -69,6 +69,11 @@ export default async function apiKey(req, res) {
     roles: user.roles,
   };
 
+  // The authorization_provider property is assigned by a composing host and is not stored in the ACL. The property must be carried over from the requesting user.
+  if (req.params.user.authorization_provider) {
+    api_user.authorization_provider = req.params.user.authorization_provider;
+  }
+
   const key = jwt.sign(api_user, xyzEnv.SECRET, {
     algorithm: xyzEnv.SECRET_ALGORITHM,
   });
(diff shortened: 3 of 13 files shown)


RobAndrewHurst:
## Roles object gate for provider users

In `rolesTemplates` a user with an `authorization_provider` property is not gated by the `user.roles` array. Scope access for such a user is decided by the provider in `authorizeScope`.

Without this a provider user with no roles array would be denied any `roles: { <role>: true }` object with "No roles were provided". A provider deny could also be bypassed by a matching `user.roles` entry, since the roles object gate never consulted the provider.

- `mod/workspace/composeObj.js` — `rolesTemplates` returns `true` for a user with `authorization_provider` after the admin `roles === true` check. Role templates are still merged. Only the `user.roles` gate is skipped.

## authorization_provider carried across token refresh

The `authorization_provider` property is assigned by a composing host and is not stored in the ACL. The property must survive every path which re-signs the user token. Otherwise which authorization model applies would flip depending on which endpoint minted the current token.

- `mod/user/cookie.js` — copies `payload.authorization_provider` to the refreshed user before signing the new cookie. Same pattern as `session`.
- `mod/user/key.js` — copies `req.params.user.authorization_provider` to the `api_user` before signing the API key.

## Tests

- `tests/mod/workspace/authorization.test.mjs` — a provider user is not denied a `roles: { analyst: true }` object, with no roles array or with non matching roles.
- `tests/mod/user/cookie.test.mjs` — the refreshed cookie keeps `authorization_provider`.
- `tests/mod/user/key.test.mjs` — the API key keeps `authorization_provider`.

RobAndrewHurst (approved):

PR #2931, opened 2026-08-17
Improve icon rendering

### LINKED ISSUE #2921: Legend construction for layers not on display
A legend is currently created when the style panel is created from the layer view method.

Legends may reference hundreds of categories represented by svg icons which are filtered out prior to display.

Dozen of layers with hundreds of icons to represent different categories puts a strain on the initialisation of the application.

The legend should only be created once a layer is displayed.

It is to be decided if the style panel should be disabled if a layer has not been loaded yet.

wbecker:
# Fix SVG-as-image icon explosion in OpenLayers feature styling

## Context

A live heap snapshot of the map app showed **51% of the JS heap (117MB) held by `blink::IsolatedSVGDocumentHost` ×1,702**, with `Detached ` ×1,708 retained at distance 3 from a GC root. The retainer chain resolved to `Window.ol → style → IconImageCache → shared → cache_ → <data URI key> → image_` — OpenLayers' global `IconImageCache.shared` singleton, not application code. ~57s of System time in the same profile is style-engine and GC work over those 1,702 near-complete Blink rendering pipelines.

Mechanism, in three parts:

1. Chrome instantiates a full isolated `Document` (with `Page`, `LocalFrame`, `StyleEngine`) for **every unique SVG referenced as an image**, because SVG is a document format: rasterizing it needs XML parsing, CSS resolution, layout (`viewBox`/`preserveAspectRatio`), and paint. Blink attaches those to a `Document` rather than maintaining a second SVG-only render path. "Isolated" is a spec requirement — SVG-as-image renders in secure static mode with its own security context, so the embedder can't read its DOM. Inline `<svg>` in the DOM does none of this. The document also can't be discarded after one rasterization (SVG is resolution-independent, so it may re-layout at a new size/DPI), so it lives as long as the image resource.
2. OpenLayers keys `IconImageCache.shared` on `src + crossOrigin + color`. Our srcs are per-variant-generated `data:image/svg+xml,…` strings, so **each variant is its own resource** — nothing dedupes, and the platform gets 1,702 live documents where it expected a handful.
3. `IconImageCache.expire()` can't drain it: it only deletes entries where `!iconImage.hasListener()`, and touches roughly one in four entries per pass. Changing the cache size does not fix this.

Intended outcome: stop creating SVG documents per icon variant, and make each variant cost a small canvas instead.

## Diagnosis

Cache families in the live app, grouped by markup signature with numbers and colours blanked out:

```
idx  count  signature
0    427    data:image/svg+xml,<?xml version="#" … height="#pt" version="#" id="svg#"
1     11    data:image/svg+xml,<svg width="#" … <circle cx="#" cy="#" fill="C" r="#" …>
2      7    (same shape family as 1)
3      6    data:image/svg+xml,<?xml version="#" … xmlns:xlink=… viewBox="
4      3    (same prolog family as 0)
5-9  1-2    https://geolytix.github.io/MapIcons/… (remote urls — deduping correctly)
```

**430 of ~460 entries are one SVG template** — a 30pt map pin, fetched as a file (Inkscape output: XML prolog, `width="30pt"`, `id="svg133"`, `<defs id="defs137">`). Diffing two members shows they are **byte-identical except a single attribute**: `fill="#1a9641"` on `path2` versus `fill="#999999"`. Nothing else varies — not geometry, not ids, not the second `path2045`.

So the shape is: **one template, ~430 distinct fill colours, one full Blink document each.** The drawn-symbol generators are fine (11 and 7 entries), the remote URLs dedupe correctly, and `markerLetter` holds just 2 entries (`#2E6F9E`/`A` and `#EC602D`/`B` — the [listview.mjs:271-277](../../projects/mapp-next/apps/mapp/lib/ui/locations/listview.mjs#L271-L277) pin path, bounded by selected locations).

That kills the cheap fix as a standalone answer. Memoizing on `template + substitute` returns the same 430 distinct strings, because the variants genuinely differ. The variant count itself is the problem, so the fix has to make a variant cheap rather than make variants rarer.

Rough sizing for why that works: a 30pt pin rasterized at DPR 2 is 60×60 RGBA ≈ 14KB. 430 of those ≈ **6MB**, against 117MB of SVG documents — a ~20× reduction with no change to the visual output and no change to how themes are configured.

## Where this lives in the code

**Key file: [apps/mapp/lib/utils/olStyle.mjs](../../projects/mapp-next/apps/mapp/lib/utils/olStyle.mjs)** — `iconStyle()` (lines 121-149) is the single choke point every feature icon passes through:

```js
icon.url ??= icon.svg || mapp.utils.svgSymbols[icon.type || 'dot'](icon, feature);
...
image: new ol.style.Icon({ anchor, crossOrigin: 'anonymous', scale, src: icon.url })
```

- **[apps/mapp/lib/utils/svgSymbols.mjs:194-211](../../projects/mapp-next/apps/mapp/lib/utils/svgSymbols.mjs#L194-L211)** — `template()`, the offending generator. It `replaceAll`s each `icon.substitute` entry into the raw template string and `encodeURIComponent`s the whole multi-KB result. No memoization, so this also re-runs per feature per render (CPU, not just memory).
- **[apps/mapp/lib/utils/svgTemplates.mjs:30-34](../../projects/mapp-next/apps/mapp/lib/utils/svgTemplates.mjs#L30-L34)** — fetches template files and stores the raw text on `mapp.utils.svgSymbols.templates[key]`.
- **[apps/mapp/lib/utils/svgSymbols.mjs:11-32](../../projects/mapp-next/apps/mapp/lib/utils/svgSymbols.mjs#L11-L32)** — the only existing memoization (`dot`), and the pattern to generalize. Its key, `Object.values({fillColor}).reduce((a, b) => a + b)`, throws on an empty object and collides by concatenation.
- **[apps/mapp/lib/layer/featureStyle.mjs:44](../../projects/mapp-next/apps/mapp/lib/layer/featureStyle.mjs#L44)** — `structuredClone(layer.style.default)` per feature, which is why the `icon.url ??=` memo on `olStyle.mjs:132` never dedupes across features for default styles. Theme paths ([categorized.mjs:51](../../projects/mapp-next/apps/mapp/lib/layer/themes/categorized.mjs#L51), [graduated.mjs](../../projects/mapp-next/apps/mapp/lib/layer/themes/graduated.mjs), [distributed.mjs](../../projects/mapp-next/apps/mapp/lib/layer/themes/distributed.mjs), [basic.mjs](../../projects/mapp-next/apps/mapp/lib/layer/themes/basic.mjs)) all spread the **shared** `cat.style` reference, so those do memoize on the theme config.
- **[apps/mapp/lib/ui/elements/legendIcon.mjs:60-121](../../projects/mapp-next/apps/mapp/lib/ui/elements/legendIcon.mjs#L60-L121)** — already rasterizes icons to canvas via `ol.render.toContext`. This is the code to extract and reuse, not to reimplement. Note it also builds its own `ol.style.Icon` from data URIs (line 98), feeding the same global cache, and is very likely the source of the 2,749 `CanvasRenderingContext2D`s in the profile.

Since `substitute` values come only from layer style config (nothing in the library derives them from feature properties), ~430 variants means the config carries ~430 coloured categories — most likely a per-record colour field or a high-cardinality categorized theme. Worth confirming for context, but it does not change the fix.

Also: `crossOrigin: 'anonymous'` ([olStyle.mjs:142](../../projects/mapp-next/apps/mapp/lib/utils/olStyle.mjs#L142), [legendIcon.mjs:100](../../projects/mapp-next/apps/mapp/lib/ui/elements/legendIcon.mjs#L100)) is part of OL's cache key and is inert for data URIs.

## Stage 1 — Memoize

Necessary groundwork: Stage 2 must rasterize **per variant**, which needs a variant key and a cache to hang it on. On its own this fixes the CPU churn and the per-render allocation, not the 117MB.

1. **`svgSymbols.mjs` — memoize `template` first.** Key a `Map` on `icon.template + JSON.stringify(icon.substitute)`. Smallest possible change, removes a multi-KB `replaceAll` + `encodeURIComponent` per feature per render.
2. **Generalize to the remaining generators** (`target`, `triangle`, `square`, `diamond`, `semiCircle`, `markerLetter`, `markerColor`, `circle`), replacing the hand-rolled `dot` key with one shared helper. Build each key from only the fields that generator reads — `fillColor` + serialized `layers` for the shape family, `color` + `letter` for `markerLetter`, `colorMarker` + `colorDot` for `markerColor`, `strokeColor` + `strokeWidth` for `circle` — joined with a delimiter.
3. **`olStyle.mjs`** — share `ol.style.Icon` instances via a `Map` keyed on `src + anchor + scale`, and drop `crossOrigin` when the src is a data URI. `scale` must stay part of the Icon key and out of the src key: `clusterScale`/`zoomInScale`/`zoomOutScale`/`fieldScale` vary it per feature.
4. **Cap the maps.** ~430 variants is bounded per session, but a per-record colour field on a bigger dataset is not. A small LRU keeps our own module from becoming the new leak. Note our eviction can't force OL's cache to release the src — another reason Stage 2 is the real fix.
5. **No change needed in `featureStyle.mjs`** — once the generators memoize, the per-feature `structuredClone` is harmless.

## Stage 2 — Canvas-backed icons 

Goal: zero `IsolatedSVGDocumentHost`.

1. **Extract the rasterizer.** Pull the canvas path out of [legendIcon.mjs:60-121](../../projects/mapp-next/apps/mapp/lib/ui/elements/legendIcon.mjs#L60-L121) into a util (e.g. `apps/mapp/lib/utils/svgToCanvas.mjs`) that takes a data URI plus target size and DPR and returns a canvas (or `ImageBitmap`). Both `legendIcon.mjs` and `olStyle.mjs` consume it.
2. **Rasterize once per variant.** Cache canvases on the Stage 1 variant key. Pass the canvas to `ol.style.Icon`'s `img` option instead of `src` — a canvas-backed image never instantiates an SVG document. Rasterize at `devicePixelRatio` and account for the maximum `scale` a variant renders at, or icons will be soft when cluster/zoom scaling enlarges them.
3. **Handle the async gap.** Rasterization is load-driven (`img.addEventListener('load', …)`, `getImageState() === 2`). `iconStyle()` is synchronous and called from the render loop, so: return the cached canvas Icon when ready, otherwise fall back to today's data-URI Icon and trigger a re-render on completion. One SVG document per variant is created transiently during rasterization — it must be released after the draw, or Stage 2 buys nothing.
4. **Migrate `legendIcon.mjs`'s own `ol.style.Icon`** (line 98) onto the shared variants so legends stop adding cache entries.
5. **Optional follow-up, config-side:** if the ~430 colours turn out to be a continuous ramp, quantizing the palette to N steps multiplies the win and is visually near-identical. Out of scope for the library change, worth raising with whoever owns the layer config.

## Type of Change

## How have you tested this?

See unit tests

## Testing Checklist

- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist


apps/mapp/lib/layer/decorate.mjs
```diff
@@ -90,6 +90,8 @@ export default async function decorate(layer) {
   // Set layer opacity from style.
   layer.L.setOpacity(layer.style?.opacity || 1);
 
+  prewarmStyleIcons(layer);
```
RobAndrewHurst: **Prewarm runs before plugin hooks**

`prewarmStyleIcons` runs before the plugin / `mapp.layer` hooks. It also bakes `icon.url` onto the shared config icons.

Problems:
- Icons added by a plugin get no redraw listener. They show grey circles until something else redraws.
- A plugin symbol that reads `feature` throws here. `feature` is undefined. The layer fails to load.
- The baked url is copied into every clone. Later changes to `fillColor` etc. are never rendered.

Fix idea: run prewarm after the hooks. Do not write `url` onto the config icon.

apps/mapp/lib/layer/decorate.mjs
```diff
@@ -116,6 +118,76 @@ export default async function decorate(layer) {
...
+
+@param {layer} layer A decorated mapp layer.
+@property {layer-style} layer.style The mapp-layer style configuration.
+*/
+function prewarmStyleIcons(layer) {
+  const icons = styleIcons(layer.style);
+
+  if (!icons.length) return;
+
+  mapp.utils.svgBitmap.onBitmapReady(() => layer.L?.changed());
```
RobAndrewHurst: **Listener leak and extra redraws**

`onBitmapReady` adds a listener but never removes it. Every layer ever decorated stays in memory (layer → L → source → features).

`layer.update()`, `entries/layer.mjs:157` and `plugins/userLayer.mjs:93` decorate again and again. Each time adds one more listener. Every bitmap batch then redraws every layer, even removed ones.

Fix idea:
- return an unsubscribe from `onBitmapReady`
- push it to `layer.removeCallbacks`
- skip when `!layer.display`

apps/mapp/lib/layer/decorate.mjs
```diff
@@ -116,6 +118,76 @@ export default async function decorate(layer) {
...
+
+  const urls = icons.map((icon) => mapp.utils.iconUrl(icon)).filter(Boolean);
```
RobAndrewHurst: **Prewarm uses up the bitmap cap**

`prewarmStyleIcons` rasterizes every icon in the whole style tree. That is all themes, hover, highlight, cluster and selected. It does this for every layer, even hidden ones.

The bitmap cache has a hard cap of 256. It never evicts. Once full:
- `requestBitmap()` returns undefined
- new icons show grey circles for the rest of the session
- their styles are never cached, so `olStyle` re-runs per feature per frame

Fix idea: only prewarm visible layers and only default/theme icons. Or evict old bitmaps.

apps/mapp/lib/ui/elements/legendIcon.mjs
```diff
@@ -120,6 +144,39 @@ function createIconFromArray(style) {
...
+
+A `data:image` src is rendered from a rasterized bitmap where one is available. A bitmap backed Icon is loaded on creation, so the load state check of the drawIcons method resolves immediately.
+
+@param {string} src The icon url or data URL.
+@param {array} anchor The icon anchor.
+@param {number} scale The icon scale.
+
+@returns {Object} An Openlayers style Icon object.
+*/
+function legendImageStyle(src, anchor, scale) {
```
RobAndrewHurst: **Duplicated icon logic**

`legendImageStyle` copies the bitmap-vs-src Icon code from `olStyle.styleIcon`. That includes `scale / bitmap.pixelRatio`. `svgSymbolUrl` (line 266) copies `mapp.utils.iconUrl`.

They already differ:
- crossOrigin handling
- warn vs silent when the type is unknown

The url is also computed twice per icon (lines 71-73 and 113).

Fix idea: one shared `iconImageStyle(src, anchor, scale)` used by both. Use `mapp.utils.iconUrl` here.

apps/mapp/lib/ui/elements/legendIcon.mjs
```diff
@@ -157,6 +217,40 @@ function createIconFromInlineStyle(style) {
...
+  const width = style.width || legendIconSize;
+  const height = style.height || legendIconSize;
+
+  const canvas = mapp.utils.html.node`<canvas
+    style=${`width: ${width}px; height: ${height}px;`}>`;
+
+  mapp.utils.svgBitmap.prewarm([src]).then(() => {
+    const bitmap = mapp.utils.svgBitmap.iconBitmap(src);
+
+    if (!bitmap) return;
```
RobAndrewHurst: **Blank legend icon when there is no bitmap**

`createIconFromBitmap` returns early if there is no bitmap. The canvas stays empty. The old div always showed the image.

This happens when:
- the 256 bitmap cap is hit
- `image.decode()` failed once (failures are permanent)

Fix idea: fall back to the old background-image div when `bitmap` is undefined.

apps/mapp/lib/ui/elements/legendIcon.mjs
```diff
@@ -157,6 +217,40 @@ function createIconFromInlineStyle(style) {
...
+
+    canvas.width = bitmap.image.width;
```
RobAndrewHurst, line 291: **Non-square icons get squashed**

The canvas CSS box is forced to `style.width × style.height`. The bitmap is drawn at full size.

Example: a 24×36 pin in a 24×24 legend cell gets squashed. The old div used `background-size: contain`.

Fix idea: keep the aspect ratio. Draw the bitmap centred inside the box.

apps/mapp/lib/utils/olStyle.mjs
```diff
@@ -100,11 +107,176 @@ export default function olStyle(style, feature) {
...
+The number of ol.style.Icon objects to memoize.
+
+The scale is part of the key and may be a unique value per feature where a cluster or zoom scale is applied. The memo is bound to prevent unbounded growth. Entries are evicted in insertion order.
+*/
+const styleIconMemoLimit = 1024;
+
+/**
+Set by the styleIcon method where a fallback symbol was returned for the current olStyle call.
+*/
+let fallbackStyle = false;
```
RobAndrewHurst: **Global `fallbackStyle` flag is fragile**

The flag is:
- reset at the top of `olStyle`
- set deep inside `fallbackIcon`
- read back to decide caching

Any nested call to `mapp.utils.style()` in between resets it. Then a style with the grey circle gets cached and is never replaced.

Fix idea: return a boolean from `iconStyle`. Or check `Styles.some(s => fallbackSet.has(s.getImage()))`.

apps/mapp/lib/utils/olStyle.mjs
```diff
@@ -100,11 +107,176 @@ export default function olStyle(style, feature) {
...
+  if (bitmap) {
+    // The bitmap was rasterized at the device pixel ratio. The Icon must be scaled by the reciprocal to render at the size the SVG declares.
+    return memoizedIcon(`${icon.url}|${anchor}|${scale}`, {
+      anchor: anchor,
+      img: bitmap.image,
+      scale: scale / bitmap.pixelRatio,
+    });
+  }
+
+  if (icon.url.startsWith('data:')) return fallbackIcon(icon);
```
RobAndrewHurst: **Grey circle can get stuck**

`styleIcon` can return the grey fallback circle. Some callers keep that style forever. They have no style function and never redraw.

Example:
- `ui/locations/listview.mjs:271` builds `location.pinStyle` once
- `entries/pin.mjs:48` uses it as a static layer style
- The pin stays a small grey dot

Same for `entries/geometry.mjs:88` and `mapview.geometry({style:{icon:{type:'dot'}}})`.

Fix idea: when the bitmap is not ready, return an Icon with the data URL, not the grey circle.

apps/mapp/lib/utils/olStyle.mjs
```diff
@@ -100,11 +107,176 @@ export default function olStyle(style, feature) {
...
+@returns {string} The icon url or data URL.
+*/
+export function iconUrl(icon, feature) {
+  if (icon.url) return icon.url;
+
+  const type = icon.type || 'dot';
+
+  const svgSymbol = mapp.utils.svgSymbols[type];
+
+  if (typeof svgSymbol !== 'function') {
```
RobAndrewHurst, line 293: **`icon.svg` is ignored when `type` is set**

`iconUrl` checks `svgSymbols[icon.type]` first. The old code was `icon.svg || svgSymbols[type](...)`.

An icon like `{type:'svg', svg:'<svg>…'}` now:
- warns "svgSymbols type svg unavailable"
- returns nothing
- the feature has no icon

The repo fixture `apps/xyz/tests/assets/styles/styleParser.json:80` has this shape.

Fix: check `icon.svg` first, as before.

apps/xyz/tests/mapp/svgToBitmap.test.mjs
```diff
@@ -0,0 +1,165 @@
+import { beforeAll, describe, expect, it, vi } from 'vitest';
```
RobAndrewHurst, line 1: **These tests never run**

No vitest config includes `tests/mapp/`. `pnpm exec vitest run tests/mapp` prints "No test files found".

With an ad-hoc config the suite passes (30 passed, 4 skipped). The tests are fine. They are just not wired in.

Fix: add `./tests/mapp/**/*.test.mjs` to a vitest include. Or add a mapp test script to turbo.

RobAndrewHurst (changes requested):

apps/xyz/tests/mapp/svgToBitmap.test.mjs
wbecker: yeah - it works on mapp-next. purposefully left them in the same position while we figure out how we want to sort these out

RobAndrewHurst: But if it's a test that is added in with the context of changes in the PR shouldn't we have it running in the CI? if not then why have it at all?

tests/mapp/svgToBitmap.test.mjs
```diff
@@ -0,0 +1,165 @@
+import { beforeAll, describe, expect, it, vi } from 'vitest';
```
RobAndrewHurst: But if it's a test that is added in with the context of changes in the PR shouldn't we have it running in the CI? if not then why have it at all?

dbauszus-glx:
I'll remove the review comments and put this into draft. This PR and logic is still very much in development.

apps/mapp/lib/layer/decorate.mjs
dbauszus-glx: The logic has been moved into the styleParser.

apps/mapp/lib/ui/elements/legendIcon.mjs
dbauszus-glx: This has been resolved.

dbauszus-glx:
I now return the crisp icon once the limit is hit. I reduced the cap to 8 in order to test this with a limited variety of icons to load.
image

dbauszus-glx:
Please note that scaled icons will appear blurred since a bitmap is scaled. 
image

dbauszus-glx:
Non square items are no longer squashed but even more pixelated than square items due to scaling. 
image

dbauszus-glx:
svgToBitmap logic must be opt in since this can mess with the display of icons. Performance will not be affected if just a few icons are used as recommended but visual appeal is hugely degraded with the svg rasterized as bitmap. SVG which are not rectangular will even further degrade and have an altered scale to not appear squashed.

The `style.bitmap_icons: true` must be set to enable rasterization.

apps/mapp/lib/utils/olStyle.mjs
dbauszus-glx: Good catch — the ordering was wrong in two places, and they disagreed with each other.

iconUrl resolved the svgSymbols type before it consulted icon.svg, so the typeof svgSymbol !== 'function' guard returned early and the icon was dropped with a svgSymbols type … unavailable warning. That regressed the original icon.url ??= icon.svg || svgSymbols[icon.type || 'dot'](icon, feature), where a truthy icon.svg short-circuited and the type was never touched. It bites the one config the styleParser leaves un-migrated: catStyle returns on if (cat.style.icon.type) return; before iconUrlSVG runs, so the legacy { type: 'svg', svg } shape — which our own styleParser.json fixture uses — reaches olStyle with svg intact and no url. iconUrl now resolves the svg source ahead of the type.

The legend had the inverse precedence: createIconFromInlineStyle read style.icon?.svg || style.svg || style.icon?.url || style.url, svg first, while createIconFromArray and drawIcons were url first. An icon declaring both rendered one thing on the map and another in the legend, and the two picked different strings when deciding the data: bitmap route. All three legend sites now call mapp.utils.iconUrl, so there is a single set of rules — url, then svg, then symbol type — and the legend and the feature render share the resolved url string, which also means one bitmap cache entry per variant rather than two. The local svgSymbolUrl helper is gone.

Covered by two regression tests in tests/mapp/olStyle.test.mjs: the svg source surviving an unavailable type, and url winning when both are declared.

apps/mapp/lib/utils/olStyle.mjs
dbauszus-glx: This issue has been addressed.

apps/mapp/lib/utils/olStyle.mjs
dbauszus-glx: This has been resolved.

dbauszus-glx:
I put this into draft to be closed. The bitmap issue is quite complex and not a priority. Ondemand requests for icons for the legend creation should be resolved in a seperate PR. I will close this PR once I have a standalone ticket and PR for the legend issue.

dbauszus-glx:
This PR is superseeded by https://github.com/GEOLYTIX/xyz/pull/2961 and will be closed.

PR #2827, opened 2026-05-19
Monorepo Package Versioning

RobAndrewHurst:
### Summary
This PR updates the package/versioning setup so dependency versions are centralized at the workspace root and app packages no longer carry their own version fields.

**The Craic?**
- Upgraded the workspace package manager from `pnpm@10.33.0` to `pnpm@11.1.3`.
- Added a root-level pnpm `catalog` in `pnpm-workspace.yaml` for shared dependency versions.
- Converted app/package dependencies to use `catalog:` instead of repeating concrete versions in each `package.json`.
- Converted local monorepo references from `file:` to `workspace:*`.
- Removed `version` fields from app package manifests so the root package version is the single source of truth.
- Added `allowBuilds` entries for `bcrypt` and `esbuild` in `pnpm-workspace.yaml`.
- Removed the root `pnpm.neverBuiltDependencies` config from `package.json`.

**Intent**
This makes the repo behave more like a single-version monorepo:

```text
root package.json = authoritative version
apps/*/package.json = package identity and dependencies only
pnpm-workspace.yaml = centralized dependency version catalog
```

**Notable Dependency Changes**
Local workspace dependencies now use:

```json
"@geolytix/xyz-app": "workspace:*"
```

Shared external dependencies now use:

```json
"express": "catalog:"
```

with the actual version declared once in `pnpm-workspace.yaml`:

```yaml
catalog:
  express: 5.2.0
```

PR #2932, opened 2026-08-19
Add v5 release notes.

dbauszus-glx:
Just the release notes. Initially compiled by claude I removed some bullet points which were intermediate patches to the general changes in the v5 release.


RobAndrewHurst:
**Corrections**

- **`file:` provider entry rewritten.** The draft said paths resolve from `process.cwd()`. That was true for exactly one commit. `apps/xyz/mod/provider/file.js:46` joins against `xyzEnv.XYZ_CWD`, and `processEnv.js` always sets that value, either to the monorepo root or to the `XYZ_CWD` env variable. #2785 superseded #2775 here. This is the sentence deployers will act on when they rewrite every `file:` reference in their workspaces, so it has to be right.

- **`COOKIE_PROPS` default corrected and a hazard documented.** The real default is `Path=${DIR || '/'}` (`processEnv.js:160`), so the shipped default ends in `Path=/`. `COOKIE_PROPS` is also the one default still using `??=`. A schema-declared empty key therefore yields cookies with no `Secure`, `HttpOnly`, or `SameSite` at all. A wrong security default was the worst error in the draft.

- **"Express dependencies at the monorepo root" bullet deleted.** #2827 reverted #2781. At `HEAD` the root `package.json` contains no `express` and the apps declare `"express": "catalog:"`. The catalog bullet above it already describes the real state.

- **`.pem`/`.crt` gitignore claim removed.** Those rules were never removed. They are still in `.gitignore`, and `*.pem` got broader. The draft could lead a fork to believe certificates are now tracked.

- **Entry point named, contradiction resolved.** The draft said v5 has "a thin root server" and, one bullet later, "the root `server.js` is gone." There is no `.js` file at the repo root. The entry point is `apps/xyz/server.js`.

- **PR #2785 added.** It was missing despite introducing `XYZ_CWD`, the `DIR`/`AUTH_PATH` normalisation, and the SAML certificate relocation. All three matter for migration.

- **Router entry recast from a v4 baseline.** The removed `./api` and `./extensions` exports and `apps/auth/auth.js` never existed in v4.24.0, so describing their removal told a migrator nothing. The entry now states what the package actually exposes. It also says that provider request-signature verification is not wired into the default server. 
 
- `providerRequestSignature.js` has zero call sites, which we flagged separately as a possible code bug.

- **Varlock entry completed.** `import 'dotenv/config'` was deleted, so a plain `node apps/xyz/server.js` reads no `.env` and fails silently via `allowFail`. The "`??=` to `||=`" claim now names its one exception. The schema requirement now applies only under `varlock run`, which matches the "Varlock is optional" entry instead of contradicting it. The hashed Vite chunk in `public/js/lib` is documented for asset pipelines.

- **Small accuracy fixes.** Five packages reported three versions, not "four packages, four versions", and the auth app was missing from the list. "Requires pnpm 11" is gone because main already pins `pnpm@11.3.0`. `vercel.json` now lives at the repo root. The deleted MAPP and browser test suites (about 80 files) are acknowledged. Co-authors are restored per the commit trailers (@simon-leech, @RobAndrewHurst, @AlexanderGeere), matching earlier release notes.

**Structural addition**

- **A breaking-changes checklist now opens the notes.** All three reviewers flagged the same problem: the breaks sat mid-paragraph across seven entries under five headings, so a migrator had to build their own checklist. The new section lists each break once and points to the detailed entries and to `SETUP.md` and `DEPLOYMENT.md`.

RobAndrewHurst (approved):

wbecker (approved):

PR #2938, opened 2026-08-25
Mvt reload filters [ENG-85] [ENG-90]

### LINKED ISSUE #2927: Filtering: polygon layer when no data is returned leaves all features on the map
Use an mvt format layer with wkt_properties: true flag attached and viewport:true

Zoom to an area with some features on the map.

Apply a filter that you know does not yield any results. IE zoom into a City Centre and apply a filter of Town Centre

You can see in the Network Tab that the filter query returns no results. 

However, the data on the map is unchanged? I would expect the map to have no features on it. 

This issue is compounded if you apply a no results filter first, and then stack other filters on top as the results become meaningless.

dbauszus-glx:
Fixes for the layer filter panel and filter interfaces, and a change to how a wkt_properties MVT layer refreshes its tile sources on reload.

### MVT wkt_properties tile refresh

The MVT tiles of a `wkt_properties` layer hold feature geometries and ID only. The tile source is not affected by a change of the layer filter, style, or params, so it must not be refreshed for these changes — clearing it renders the layer blank until the feature properties for the current filter have returned.

`layer.reload()` now returns after calling `wktPropertiesLoad` for a `wkt_properties` layer. The tile sources are cleared and refreshed from the `wktPropertiesLoad` resolution instead, and only if the response holds a feature which has not been rendered before, eg. after a location geometry has been created. A feature which is no longer in the `featuresObject` is not rendered by the `featureStyle` method, so a removal requires no refresh. This removes the need for a `geometry` reload param on the `reload` method.

The `featuresObject` is no longer reset by the `reload` method. The current feature properties keep rendering until the response lands.

### Filter panel

- The filter type is derived from the infoj entry type. An entry type without a matching filter type is skipped silently — the `is not a valid filter type` warning is only logged for a filter type which has been configured.
- An entry may opt out of the filter list with `filter: false`.
- The skip, field, include, and exclude checks have been moved into an `excludeEntry` method.
- The filter panel drawer defaults to `disableOnHide`.
- The panel is updated from the `removeFilter` method rather than the dropdown callback. The panel is no longer updated for a filter card without a filter value.

### Drawer disabled with layer display

A drawer created with the `disableOnHide` flag is flagged with a `disable-on-hide` class. The layers view `changeEnd` method skips these elements when enabling nested elements, so the drawer remains disabled while the layer is not displayed.

### Filter interfaces

- The layer is no longer reloaded when a filter which is not current is deleted.
- A numeric filter only reloads the layer if `Min` or `Max` bounds are configured. A range which matches the min and max values of the data does not filter the layer.
- The `distinct_values` query of a `filter_in` interface falls back to the first configured layer table for a layer which is out of its zoom range. The query would otherwise be sent without a table param and fail.
- The `filter_in` interface handles an `Error` response from the `xhr` method and filters out `undefined` values.

### Query module

- The `locationCount` query sends the current layer geom, which is required for the viewport condition.
- The layer query responds `400` for a `viewport` param without `geom` and `srid` params. These are required to be substituted into the viewport condition.


dbauszus-glx:
My test layer looks like so:

```json
    "iuc": {
        "name": "Internet User Classification",
        "format": "mvt",
        "wkt_properties": true,
        "filter": {
            "viewport": true,
            "includeAll": true
        },
        "tables": {
            "9": null,
            "10": "geodata.uk_cdrc_open_internet_user_classification"
        },
        "geoms": {
            "11": "geom_3857_10m",
            "12": "geom_3857_5m"
        },
        "srid": "3857",
        "qID": "id",
        "infoj": [
            {
                "type": "key"
            },
            {
                "field": "geom_3857_5m",
                "fieldfx": "ST_asGeoJSON(geom_3857_5m)",
                "type": "geometry",
                "display": true,
                "class": "display-none"
            },
            {
                "field": "pin",
                "fieldfx": "ARRAY[ST_X(ST_PointOnSurface(geom_3857_5m)),ST_Y(ST_PointOnSurface(geom_3857_5m))]",
                "type": "pin"
            },
            {
                "title": "Group",
                "field": "iuc_name",
                "filter": {
                    "type": "in",
                    "distinct": true,
                    "dropdown_pills": true
                },
                "inline": true
            }
        ]
    }
```

simon-leech (changes requested):
Not sure if we address on here or separately but would prefer to do it here as its still a bug to me about the same thing! 

Image

Once you have applied a filter, the number can be too large for the space and it should be being formatted once the slider has been moved. 

The rest looks to be working fine for me, messed with a range of filter types.

cityremade:
@simon-leech is this in the multi filter dialog or the layer panel filter section ?

simon-leech:
@cityremade in the multi filter itself but I'll double check it this afternoon!

cityremade (approved):

simon-leech (approved):

PR #2940, opened 2026-08-27
Resolve workspace scopes from the cached workspace [ENG-241]

RobAndrewHurst:
## Summary

Requests to `/api/workspace/scopes` were slow because `getScopes()` called `workspaceCache(true)`. The forced rebuild cleared the workspace cache and the source map on every request, so every source was fetched again and every locale, nested locale, and layer was composed again. The rebuild also discarded the cache for concurrent requests to other endpoints.

## Changes

- `apps/xyz/mod/workspace/scopes.js`
  - `getScopes(force)` resolves the workspace from the cache. The cache is only rebuilt with the `force` param or when `WORKSPACE_AGE` is exceeded.
  - The composition promise is stored in a `WeakMap` keyed by the cached workspace object. Repeat and concurrent requests share one composition until the workspace cache is rebuilt. A failed composition is removed from the map so the next request retries.
- `apps/xyz/mod/workspace/_workspace.js`
  - The `scopes` handler passes `req.params.force` to `getScopes()`. `GET /api/workspace/scopes?force=true` restores the previous rebuild behaviour on demand.
- `apps/xyz/tests/mod/workspace/scopes.test.mjs`
  - Tests for repeat requests, concurrent requests, and the `force` param.

`composeWorkspace()` in `_workspace.js` still forces a rebuild and is unchanged in this PR.

## Testing

- `npx vitest run tests/mod/workspace` — 12 files, 119 tests pass.
- `npx biome check` on the changed files — clean.

Resolves ENG-241

### DIFF
--- a/apps/xyz/mod/workspace/_workspace.js
+++ b/apps/xyz/mod/workspace/_workspace.js
@@ -271,12 +271,16 @@ function roles(req, res) {
 @description
 The scopes method returns an array of scopes which are the templateScopes assigned to each template in the workspace.templates{} object.
 
+The scopes are resolved from the cached workspace. The workspace cache is rebuilt with the force request param.
+
 @param {req} req HTTP request.
 @param {res} res HTTP response.
 
 @property {Object} req.params HTTP request parameter.
 @property {Object} params.user User requesting the scopes.
 @property {boolean} params.user.admin Whether user has admin privileges (required).
+@property {boolean} [params.force] Rebuild the workspace cache before the scopes are resolved.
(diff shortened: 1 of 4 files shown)


simon-leech (approved):
This is working well for me - very slow on cold starts but subsequent calls its super speedy with the caching! 

I pushed a commit to change the logic on the admin view so that the scopes are injected once they are available, rather than making the admin view hang until they are created. This prevents a white screen for 30 seconds while they populate.

cityremade (approved):

PR #2941, opened 2026-08-27
bug(Roles): Two layers with different roles and same template leaking [ENG-276]

simon-leech:
This PR adds a failing test to expose an issue with roles. 
There are two layers inside locale_two_layers_roles.json. 

``` json
 "layers": {
    "layer": {
      "roles": {
        "Super": null,
        "Test": {
          "name": "Test A"
        }
      },
      "template": {
       "src": "file:./tests/assets/layers/template_test/layer.json"
      }
    },
    "layer_b": {
      "roles": {
        "Test": {
          "name": "Test B"
        }
      },
      "template": {
       "src": "file:./tests/assets/layers/template_test/layer.json"
      }
    }
  }
```

Role - Super should only see `layer`
Role - Test should see both layers. 

Currently, Super is seeing both incorrectly. 

When composing a layer or locale, its roles property is processed per-key:

roleKey: true / roleKey: null → pushed to both accessRoles and gateRoles ([composeObj.js:384-388](vscode-webview://1iriiuae07l8akuig5euccq1tftk99se99thodttv5tfsbqk1c05/apps/xyz/mod/workspace/composeObj.js#L384-L388)).
roleKey: {...} (object value, used to conditionally merge role-specific properties) → pushed only to accessRoles, never to gateRoles ([composeObj.js:390-397](vscode-webview://1iriiuae07l8akuig5euccq1tftk99se99thodttv5tfsbqk1c05/apps/xyz/mod/workspace/composeObj.js#L390-L397)).
The final access decision only checks gateRoles:

if (!gateRoles.length) return true;   // composeObj.js:412
So if a roles object contains only object-valued entries — no plain true/null key — gateRoles is empty and the object is granted to every user unconditionally, regardless of what roles are listed in accessRoles or what roles the user actually holds. The role keys are enumerated but never enforced.

### DIFF
--- a/apps/xyz/mod/workspace/composeObj.js
+++ b/apps/xyz/mod/workspace/composeObj.js
@@ -69,7 +69,8 @@ export default async function composeObj(obj, user) {
     );
   }
 
-  const rolesCheck = await parseTemplates(obj, user, templateScope);
+  // The obj is the root object of the composition. The root obj is gated by every role key in a roles object.
+  const rolesCheck = await parseTemplates(obj, user, templateScope, true);
 
   if (rolesCheck instanceof Error) {
     return rolesCheck;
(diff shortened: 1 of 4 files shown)


RobAndrewHurst:
**Gate root objects by every role key in composeObj**
A layer or locale with only object role values (eg. roles: { Test: {...} }) leaked to users without the role. Only true/null role values gated the object.

The root object of composeObj is now gated by every role key, matching the legacy roles check. Nested objects keep the existing rule so object role values (eg. skipEntry on an infoj entry) still merge without hiding the entry from other users.

AlexanderGeere (approved):

simon-leech:
I can't approve this as I made this PR but this is working well for me! 
I tested an example with two layers and correctly hidden now. 
I also  tested examples whereby infoj entries are using `skipEntry` under `roles` objects and that's still working as expected 👍

RobAndrewHurst (approved):

PR #2944, opened 2026-09-03
fix(getSrc/Router): Fix for getSrc and router to parse templates correctly [ENG-289]

AlexanderGeere:
## Description
There were two issues: 
1. The template for the message does not have a provider so in `getSrc` on line 68 an error was returned due to the missing provider. This call to getSrc is now made conditional on the presence of a provider e.g file:, https:, etc. 

A test was also added that would fail not on this PR

2. The verify route was missing the `{/:key}` part of the url, so when a verification url was clicked the path would not match as the link includes the verification token so you would fall through to the login screen. This was introduced when we changed from the `api.js` structure in 30a7d173

A test has also been added to check that verify flows through to the correct place as well. 

## Linear Issue
[ENG-289](https://linear.app/geolytix/issue/ENG-289/v501-new-user-registration-flow-incorrect)

## Type of Change
- ✅ Documentation
- ✅ Testing

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist


simon-leech (approved):
Tested the flow of adding a new user - and everything works as previously should👍

simon-leech (dismissed):
All working for me from a user registration perspective! 

However, as an administrator the email that came through to me did not have the url on it. 

'A new account has been verified on ' is all it said - it should have the url after this. 
The click through button on the email thus does not work as it redirect you to just 'api/user/admin?email=x' not the domain.

AlexanderGeere:
@simon-leech 
> All working for me from a user registration perspective!
> 
> However, as an administrator the email that came through to me did not have the url on it.
> 
> 'A new account has been verified on ' is all it said - it should have the url after this. The click through button on the email thus does not work as it redirect you to just 'api/user/admin?email=x' not the domain.

The host was never set in the verify endpoint, in [e6dd53b](https://github.com/GEOLYTIX/xyz/pull/2944/commits/e6dd53b2f72aba37a2d9e6beed52467f2df70bb2) the host is now set using reqHost

simon-leech (approved):
Working for me tested again and all looks good 👍

RobAndrewHurst (approved):

PR #2946, opened 2026-09-05
fix(Geometry/Modify): Attach the handler to the interaction so it can be cleared [ENG-88]

### LINKED ISSUE #2928: Modify interaction escape event listener
Something must have gone wrong with this commit. https://github.com/GEOLYTIX/xyz/commit/d0c46378b756d7d5e18386f8953cff79c7b54fd2

I can modify a point geometry and save the modified geometry. The interaction finishes.

However. I now press escape and the escape function is still triggered finishing now the select interaction.

Additional eventlistener are added everytime the modify interaction is called.

After modifying multiple geometries you end up with the eventlistener being trigger multiple times.

simon-leech:
## Description
Root cause of [#2928](https://github.com/GEOLYTIX/xyz/issues/2928): commit d0c46378b moved finish() out of modify()'s closure into a module-level function, but the escape keyup handler stayed nested inside modify(). 
removeEventListener silently does nothing when the reference doesn't match, so every modify interaction leaked its keyup listener permanently instead of being cleaned up on finish. 
Pressing Escape later then fires all those accumulated stale listeners, each calling mapview.interaction.finish() — but by then mapview.interaction points at whatever interaction is currently active (e.g. select), so it gets unexpectedly finished. 

## Type of Change

## How have you tested this?
Tested editing a geometry locally and all working as expected.

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Updated Existing Tests
- ✅ Ran locally on my machine

### DIFF
--- a/apps/mapp/lib/mapview/interactions/modify.mjs
+++ b/apps/mapp/lib/mapview/interactions/modify.mjs
@@ -95,8 +95,11 @@ export default function modify(params, mapview = this) {
     e.key === 'Escape' && mapview.interaction.finish();
   }
 
+  // Store the handler on the interaction so finish() can remove the exact listener that was added.
+  mapview.interaction.escape = escape;
+
   //End the modification on escape key.
-  document.addEventListener('keyup', escape);
+  document.addEventListener('keyup', mapview.interaction.escape);
 
   mapview.interaction.interaction = new ol.interaction.Modify(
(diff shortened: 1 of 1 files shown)


AlexanderGeere (approved):

cityremade (approved):
ok.

PR #2953, opened 2026-09-15
Fix base layer order and admin requests changing the workspace cache [ENG-326]

RobAndrewHurst:
## What this fixes

Fixes [ENG-326](https://linear.app/geolytix/issue/ENG-326).

### Base layers appear in the wrong order

When a locale uses a `template`, XYZ can put local layers before base layers in the left-hand panel. Base settings can also overwrite settings from the locale's template.

The fix applies settings in this order:

1. Base locale settings.
2. The locale's template.
3. Settings written directly on the locale.

Later settings override earlier ones, while base layers keep their position at the start of the list. A locale without its own template still inherits the base locale's template.

### Admin requests change the shared workspace cache

The admin workspace `test` and `compose` endpoints build the workspace with role checks turned off. They then overwrite cached locale definitions with that result, losing the original role rules. Later users can receive settings their roles should exclude.

The `compose` endpoint also removes `templates` and `scopes` from the cache, which can make later requests crash.

The fix builds admin output in a separate object with its own locale map. Other requests keep using the original locale definitions.

## Why this belongs in XYZ

XYZ controls the merge order and the shared workspace cache. Sorting the panel or refreshing the cache would leave the underlying bugs in place.

## Tests

Added nine test cases covering layer order, settings overrides, inherited templates, both admin endpoints, nested locales, and role changes while the workspace remains cached.

Before the fix, three tests fail on upstream `patch` at `5e5d138`:

- Layers appear as `Local, Base` instead of `Base, Local`.
- Editor-only settings remain after an admin `test` request.
- A request crashes after admin `compose` removes `workspace.scopes`.

After the fix:

- `pnpm --filter @geolytix/xyz-app test`: **341 tests passed across 50 files**.
- Biome checks and `git diff --check` passed.

We reproduced the cache bug through the admin endpoints. We could not reproduce an ordinary user's first request locking the cache to their roles. Any remaining delay after a role change needs a separate investigation.


wbecker (approved):
Seems like the fix is fine :)

As a broader question I wonder if getLocale needs to be refactored a bit - its just doing a lot of things one after another. I wonder if it would be easier to read if it were written like:

```
function getLocale() {
  doOneThing()
  doAnotherThing()
}
```

And you could individually test those smaller parts

wbecker:
Claude says:

The feedback is directionally right but understates the difficulty. This isn't a straightforward extract-method refactor — the control flow has structural complications that need to be addressed first, or the extraction just moves confusion around.

What makes naive decomposition tricky here:

The recursion on line 55–59 returns before the post-processing block (keys join, plugins envReplace, localeLayers). That means post-processing only runs on the final recursive call. If you extract the post-processing into a finalize(locale) step, you need to make sure the caller understands that finalize is skipped for parent locales. The current structure encodes that implicitly via early return.

params.locale is mutated by shift() on line 11 to drive the recursion. That's a side-effect coupling between "resolve the current key" and "set up the next recursive call." Any decomposition needs to either make that explicit or switch to an iterative approach.

composeObj is called twice with different arity (lines 43–44) — once without defaults, once with. That looks like a bug or an intentional two-pass merge, but either way it's not obvious why, and extracting it into a named step doesn't clarify it unless you also name why there are two passes.

What I'd actually do:

Replace the recursion with iteration. The function is processing a list of locale keys left-to-right, folding each into the previous. That's a reduce. Making it a loop eliminates the early return and makes the "post-processing runs once at the end" contract obvious.

```
export default async function getLocale(params) {
  const workspace = await workspaceCache();
  const keys = normalizeLocaleKeys(params.locale);

  let locale = null;

  for (const key of keys) {
    const resolved = await resolveLocale(key, workspace);
    if (resolved instanceof Error) return resolved;

    const merged = applyDefaults(resolved, workspace, locale);
    const composed = await applyUserOverrides(merged, params.user, workspace, locale);
    if (composed instanceof Error) return composed;

    composed.name ??= composed.key;
    locale = mergeParentLocale(composed, locale);
  }

  return finalize(locale, params);
}
```

Where finalize does the keys join, plugins envReplace, and localeLayers.

This makes each extracted function independently testable and makes the overall flow readable top-down, which is what the reviewer actually wanted.

Before doing any of it, clarify what the double composeObj call is supposed to do. If it's a bug, fix it first — refactoring around a bug locks the bug in.

AlexanderGeere (approved):
tested on `bugs_testing/layer_order/workspace.json` 

Order is preserved as it should be.

dbauszus-glx:
Were the review notes from @wbecker adressed? I don't see a committ after the comments but I also don't see a second call to the composeObj method from the getLocale module/method.

Is there a ticket linked to this bug for future reference?

PR #2954, opened 2026-09-18
Allow xyz to be used as a login server [ENG-107]

### LINKED ISSUE #2956: Support JWT claims
External authentication and authorization services may require information in regards to the JWT user token provided in cookies.

It should be possible to configure values for the JWT typ, iss, aud claims as process environment variables.

AlexanderGeere:
## Description
This PR will allow other apps on a the same domain as the xyz instance.

#### Existing
`SECRET_KEY` - Tells `xyz` to use `RS256` instead of `HS256`. 
`COOKIE_PROPS` - This will have to specified in the env as the domain needs to be set. 
#### Added
`TRUSTED_RETURN_HOSTS` - where the login is allowed to return to. 
`SESSION_TYP` -  A description of what type of token it is e.g. 'session'.
`SESSION_ISS` - A plain string describing wo or what issued the token.
`SESSION_AUD` - The intended recipient of the token.

### Functionality
The session claims are added to the token. These are not consumed here but by the service that requested to token. 
Existing login routes will not be affected and should work the same as before. 

`return_to` is set as a query param and this is where the redirect cookie is set to go once login is successful. 

### Utilities
`printLoginPublicKey.js` - This utility prints the derived public key from the secret key. This public key will be used in the calling app as an environment variable. 

## Linear Issue
[ENG-107](https://linear.app/geolytix/issue/ENG-107/login-page-redirect-successful-logins-from-admin-tool-back-to-admin)

## Type of Change
- ✅ Documentation
- ✅ Testing

## How have you tested this?
Tested Locally by checking login still works as expected

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ New Tests Added
- ✅ Ran locally on my machine

## Code Quality Checklist


utils/printLoginPublicKey.js
```diff
@@ -0,0 +1,23 @@
+// Derives a public key for LOGIN_PUBLIC_KEY from the SECRET_KEY private key
+// file, the one thing today's RS256 setup doesn't already produce.
+import { createPublicKey } from 'node:crypto';
```
wbecker, line 3: should think about how to package this

apps/xyz/mod/user/sessionClaims.js
```diff
@@ -0,0 +1,9 @@
+// Claims that let another service authenticate its users through xyz,
+// distinguishing a real session from a key.js API-key token.
+export default function sessionClaims() {
+  return {
+    typ: xyzEnv.SESSION_TYP,
```
wbecker, line 5: is there an .env-example or something to put these in?

utils/printLoginPublicKey.js
wbecker: like there should be a command in package.json to run it?

wbecker: on there is already :) Good job!

wbecker (changes requested):

dbauszus-glx:
Are typ, iss, aud common abbreviations? It's not entirely clear to me what these mean. Would it be possible to use type, issuer, audience as more descriptive names.

The session claim module and method need a bit more jsdocs. Claude is pretty good at generating these.

This will allow us to get a better idea of what a method is, does, and returns on hover in VSCode.

image

umbra is a private repository which I can see. Let me create a public issue in this repo in regards to the requirement for this PR.

apps/xyz/mod/user/sessionClaims.js
dbauszus-glx: The module and method will need to be documented in regards to the use of these variables.

dbauszus-glx:
@AlexanderGeere Are SESSION_TYP, SESSION_ISS, and SESSION_AUD only relevant to the cookie? Would it make more sense to user COOKIE_ as reference instead of SESSION_ as we already have other COOKIE_ variables. Either I believe this will become clearer once we see the sessionClaims module documentation.

AlexanderGeere:
@dbauszus-glx 
> @AlexanderGeere Are SESSION_TYP, SESSION_ISS, and SESSION_AUD only relevant to the cookie? Would it make more sense to user COOKIE_ as reference instead of SESSION_ as we already have other COOKIE_ variables. Either I believe this will become clearer once we see the sessionClaims module documentation.

Relevant to the JWT which ends up in the cookie, so I would say we could do JWT_TYPE, JWT_ISSUER, JWT_AUDIENCE?

AlexanderGeere:
@dbauszus-glx I renamed the env variables, but in the actual jwt those names `iss`, `typ` and `aud` are part of the jwt standard. 

Added function level documentation for `sessionClaims` and some further explanation in `processEnv.js`

dbauszus-glx (approved):
I think jwtClaims and JWT_ variables work best. I updated this with the descriptors for the JWT claims properties and created a linked ticket.

wbecker (approved):

RobAndrewHurst (approved):

Issue #2955, opened 2026-09-21
multiple legend methods are added to the layer.showCallbacks array


dbauszus-glx:
While reviewing #2945 I noticed that the theme method adds a legend method to the layer.showCallbacks[] array.

The method is never removed from this array. Switching between a couple of themes will add multiple legends method to the showCallbacks array.

<img width="545" height="545" alt="Image" src="https://github.com/user-attachments/assets/1089fd91-f1af-4edc-bbe3-0c508f024831" />

PR #2945, opened 2026-09-05
fix(LayerStyle): Legend not rendering after theme swap [ENG-295]

simon-leech:
## Description

Swapping a layer's theme dropdown or loading a layer that's displayed by default left the legend blank, because legend rendering had been moved onto layer.showCallbacks, which only fires from layer.show() — a path neither of those cases actually triggers. This fix renders the legend directly in both cases instead of relying solely on the deferred callback.

## Type of Change
- ✅ Bug fix (non-breaking change which fixes an issue)

## How have you tested this?
Off this PR - when you swap theme the legend no longer shows the thematic. 
Off this PR - when you have a layer on by default, the legend doesn't show the thematic. 

## Testing Checklist
- ✅ Existing Tests still pass
- ✅ Ran locally on my machine

PR #2959, opened 2026-09-22
Add review-pr skill

### LINKED ISSUE #2957: Claude PR review skill
It would be advantageous to have a Claude skill to review pull requests. Issues such as checking any method touched in it's entirety not just whether a specific issue has been resolved.

The skill should also make sure to check event listeners and whether an issue has been resolved without touching an unnecessary ammount of different modules.

Touched modules should have a full test coverage if touched in a PR.

Tests must be provided for new features. Documentation should conform with standards outlined in the documentation.md.

anush-933:
### Brief summary of the change made

Adds `.claude/skills/review-pr/SKILL.md`, a skill that reviews a pull request, a branch, or uncommitted working changes and writes a report to the terminal.

Built to the two decisions from #2957: named `review-pr` so the `review-*` namespace stays open for other review skills, and reporting to the terminal rather than posting a comment, so a contributor can check their own work before pushing and a maintainer can run it while reading someone else's change.

It covers the checks in the issue — reading each touched method in full rather than only the diff hunk, listener and callback registration, scope across modules, tests and coverage for touched modules, and JSDoc conformance with DOCUMENTATION.md. #2955 is written in as the worked example for the listener check, since a callback pushed to `layer.showCallbacks[]` and never removed is the exact shape that reads as ordinary assignment rather than as subscription.

fixes #2957

### Are there any other side effects of this change that we should be aware of?

None at runtime — one new file, nothing imported or built.

I ran the skill blind against two merged pull requests and rewrote it twice from what came back. Reviewing #2945 it independently reached #2955, and also flagged that the new `if (layer.display)` guard's justifying comment does not hold for the second caller of `panel()`: `apps/mapp/lib/ui/locations/entries/layer.mjs` builds the style panel before it calls `entry.display && entry.show()`. Reviewing #2944 it questioned whether widening `/api/user/verify{/:key}` moved the route onto a registration carrying a different middleware chain. Both look worth their own issues — happy to open them, though I would rather have a maintainer's read first since I am new to this codebase.

The testing also corrected the skill in ways worth recording. A first draft sent reviewers to a browser test suite whose source is not in the repository, and a later one would have accepted a cached `>>> FULL TURBO` result as a passing test run. Both are now handled explicitly.

Two unrelated documentation drifts I noticed and deliberately left out of this pull request, so it stays on one concern: DEVELOPING.md and SETUP.md both state pnpm 10 while `package.json` pins `packageManager: pnpm@11.3.0`, and SETUP.md describes three apps where there are now four.

### Pull Request checklist

- Test cases: not applicable — this adds a skill document and no code paths.
- Documentation: the skill is self-documenting, and no existing document describes the `.claude/skills` directory.
- Follow-up issues: noted above, happy to raise them if useful.


RobAndrewHurst:
This is already a built in skill in Claude.

dbauszus-glx (approved):
I am pretty happy with this. Does exactly what I asked it to do. I just run this on my own PR and got valuable insight.

@RobAndrewHurst This is different to code-review which I also ran and where I found different things.

The important thing is that this is a living skill which we can adjust the more we use on PR on our own judgement, documentation, and guide lines.

dbauszus-glx:
@anush-933 We really appreciate that you take the time to contribute to this open source project. 🙏🏻

